PR #2, opened 2022-05-14
Initial attempt at array theory

philzook58:
Attempt at encoding the theory of arrays.
Currently extraction is crashing.
Couldn't quite figure out how to add new primitives. Not sure if there is primitive disequality on i64? Tried encoding it as an i64,i64 -> i64 function where 0 is false and 1 is true.
Need primitive disequality to reflect into the `neq` table.


tests/array.egg
```diff
@@ -0,0 +1,70 @@
...
+
+(relation neq (Math Math))
+
+(relation blowup ())
+
+(rule ((neq x y))
+      ((neq y x)))
+
+(rule ((neq x x))
+      ((blowup)))
```
philzook58, line 26: panic?

PR #3, opened 2022-05-19
Add type checking example

remysucre:


tests/typecheck.egg
```diff
@@ -0,0 +1,97 @@
...
+)(
+    (= t t2)
+))
+
+; ctx |- x : t
+; ------------------ y != x 
+; ctx; y: t |- x : t
+(rule ( ; creating demand to check ctx |- x : t
+  (= (typeof (Cons y ty ctx) x) t)
+  ; (!= x y) TODO disequality not supported yet
```
yihozhang, line 60: `x` is an expression (var), not a string. The same applies to the next rule.

The following rules will run after #15 is merged :tada: :

```
(rule ( ; creating demand to check ctx |- x : t
  (= (typeof (Cons y ty ctx) vx) t)
  (= vx (Var x))
  (!= x y)
)(
  (= (typeof ctx vx) (typeof ctx vx))
))

(rule ( ; only check ctx; y: t |- x : t that has a demand
    (= (typeof ctx xv) t1)
    (= (typeof (Cons y ty ctx) xv) t2)
    (= xv (Var x))
    (!= x y)
)(
    (= t1 t2)
))
```

mwillsey:
@yihozhang Since we had to revert #15, we can't merge this yet. But I think we have what we need otherwise.

yihozhang:
Oh I thought you only reverted the last commit in #15. Everything except for the last commit in #15 is merge-ready (at the moment of the PR). Do you think we should attempt to merge #15 again, or since #15 will be obsolete regardless, should we wait until the new intermediate language is stabilized?

mwillsey:
@yihozhang yeah, can you PR those ready-to-go commits then?

PR #4, opened 2022-07-04
Initial Datalog Syntax

philzook58:
For human use, I vastly prefer datalog syntax to s-expressions. There is no good datalog convention for declaring schema, datatypes, or other commands like queries, so s-expressions is perfectly fine for those. My suggestion is that you can go into and exit datalog syntax by a command of the shape "(dl ...)" or "(datalog ...)". I'd prefer not to wrap this around every command, but instead enter a full block of datalog, but this requires a little more refactoring of the parser at the `Program` level.


mwillsey:
I'm not convinced it should be a unified syntax, but it's definitely good enough to work with for now.

PR #6, opened 2022-07-07
Initial Query Command

philzook58:
Minimal version of adding a query command. The current printing is subpar, it is unclear that we want to add sexp as a dependency. To be discussed.
Possibly when printing the query of an eclass id, we also want to extract a term for it rather than print the id. In this case, query may subsume the extract command.
The command `(query ((path x y)))` will print the entire table `path`. 
It is noted in a previous comment that `(query ((path 1 2)))` subsumes the capabilities of `(check (path 1 2))`
It is necessary but not done that we print what element of the sexp corresponds to which pattern variable

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ num-bigint = "0.4.3"
 num-integer = "0.1.45"
 num-rational = "0.4.0"
 num-traits = "0.2.15"
+sexp = "1.1.4"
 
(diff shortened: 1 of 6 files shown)


philzook58:
Still very rough, but it works and I think it is really useful to have this for debugging other branches. Try it out with `cargo run -- tests/datalog.egg`. It would be better to have a consistent s-exp format for all this data, since I think these query sexp might be consumable by other tools.

PR #15, opened 2022-07-21
Adding disequality

yihozhang:


mwillsey:
How does disequality work? Is it "local" like it is in eqsat? Or is it stratified or otherwise proven somehow.

yihozhang:
@mwillsey Right now disequality only works for primitives and no constraint is imposed. The user can do whatever they want, but only for primitives. We figured that disequality on sort and lattice values is essentially not sound (happy to talk more about this).

The purpose of this PR is mostly to support the type checking example. All the predicates like `!=` ignored during GJ and checked at the end. I would say the code right now is a little bit hacky and inefficient. For example, without modification, `(!= a b)` will be translated into `(!= a b c)` because `(!= a b)` has unit type, but there's no way to enumerate `c`. What I'm currently doing is to have special cases for predicates and loose the assertions made in gj.rs.

To be more efficient and scalable, we may need to rethink our design later. A good example we can look like is Souffle's architecture.

philzook58:
Just a thought:
```
pub struct Predicate {
    args: Vec<InputType>,
    f: fn(&[Value]) -> bool,
}
```
could this possibly be generalized to something like
```
pub struct Predicate {
    args: Vec<InputType>,
    outs: Vec<InputType>
    f: fn(&[Value]) -> Vec<&[Value]>,
}
```
In principle this capability would also subsume primitive functions. It makes sense to specialize for `-> bool` and functions
 though.

yihozhang:
@philzook58 Yes I wanted to do this, since after all a predicate is just a primitive that returns bool. But the current design is not yet ready for this for many reasons: for example, our generic join implementation does not support function operators; we don't know how to do type inference for overloaded primitives, etc., etc. I think ultimately we want something like [Rel](https://relational.ai) but with simplifications.

PR #18, opened 2022-08-12
Naive Linear Algebra

philzook58:
Very much related to the previous eqsolve example, but now using built in rationals (and guarding on rationals not being 0). A chaotic process of isolating variables and substituting them.

Something isn't working right now though. I'm a bit confused. Something seems to be breaking monotonicity. I remove facts and checks start passing.


philzook58:
Actually, it looks like checks pass 1 out of 4 times. Very odd.

PR #19, opened 2022-08-13
Simple bitwise operations

philzook58:
Simple bitwise operations as primitives. Useful for bitset operations and also the theory of bitvectors. See commented out section of test for bitset operations. Should supporting things like that be something in the surface language?


mwillsey:
I feel like those should be primitives, no? What is the advantage of them being functions instead?

PR #20, opened 2022-08-13
Bool Primitive

philzook58:
Simple bool primitive support. This may also be relevant to a theory of bitvectors depending on how it is designed.

Particularly intriguing is the `==` operator. Something very fun about the idea of having both `=` and `==`.

I'm a little concerned about the name clash between `!` used for boolean negation and for datalog negation, but I don't know what to do about it. We don't support datalog negation atm to my knowledge.

### DIFF
--- a/src/ast/expr.rs
+++ b/src/ast/expr.rs
@@ -9,6 +9,7 @@ pub enum Literal {
     Int(i64),
     Rational(BigRational),
     String(Symbol),
+    Bool(bool),
     Unit,
 }
 
@@ -33,6 +34,7 @@ macro_rules! impl_from {
 }
 
 impl_from!(Int(i64));
+impl_from!(Bool(bool));
 impl_from!(String(Symbol));
 
 impl Literal {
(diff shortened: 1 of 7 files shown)


mwillsey:
Why do we need `==`? What's the difference with `=`?

Also, make sure to run `cargo fmt` (or just make your editor do it :smile:).

mwillsey:
I would love to have a `not` for `Option<()>` expressions... there are actually pretty reasonable semantics for it I think. To me, this makes any use of `bool` unneeded. What do you think we should use `bool` for?

philzook58:
I don't think you could write `not : Option<()> -> Option<()>` as a primitive though? That's interesting. A negated predicate is in some sense very similar to your notion of "computed function" in that it can't ground it's arguments. It downcasts a regular function to a computed one or something.

`bool` is for actual data you may want to store the negation of, and for `ite` expressions. Consider `(function even (Math) bool)` vs `(relation even Math)`. They do kind of represent the same thing and actually you could also have `(relation odd Math) (relation even Math)` which is almost the same thing, except `(function even Math bool)` by definition can't represent something being even and odd. I feel `(function even Math bool)` is a bit more natural. Sometimes in souffle I have to make both `foo` and `not_foo`. This kind of does that for you.

philzook58:
I don't know what to do with this, but it is very reminiscent of the belnap booleans discussion in the Marshall paper https://dl.acm.org/doi/pdf/10.1145/3341703 page 10. They also construct booleans as a pair of incompatible semidecidable thingies

PR #22, opened 2022-08-15
Equational Theory of Sets

philzook58:
Some ideas on an equational theory of sets. WIP

### DIFF
--- a/tests/sets.egg
+++ b/tests/sets.egg
@@ -6,6 +6,12 @@
 
 ; I make no claim that the following axioms are complete.
 
+
+; For primitive sorts, built in native sets are vastly preferable
+; For eqable sorts, this representation has the benefit
+; that the sets can collapse and discover equalities
+
+
 (datatype Set)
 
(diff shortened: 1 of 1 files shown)


mwillsey:
We have primitive map now; should we close this?

philzook58:
Does primitive map supports sets of sorts or only sets of primitives? This can hold the former, which I think is still interesting. It also mimics the SMT theory of sets.

mwillsey:
Sure, why not. But it doesn't "respect" the fact that id's might change or merge keys.

philzook58:
Hmm. Is there a mechanism to normalize them via findroot? Would that be possible to add as a primitive? Primitive functions don't have access to the unionfind table so far as I can tell. Perhaps a very special one

mwillsey:
Yeah, idk how to do this. Ideally every sort would be able to canonicalize itself. Perhaps that's the interface... Makes me think that would be useful in general (e.g. polynomials or something).

PR #25, opened 2022-08-18
Replicating the lambda example

yihozhang:
Currently free(r) variable analysis is missing. To support it requires a StringSet lattice where merge is defined as set intersection. Interestingly, for such a StringSet, the monotonic query operation is `not_contains` (instead of `contains`)


yihozhang:
@mwillsey All the tests from `lambda.rs` (except those for explanations) are replicated. We can merge this now!

PR #28, opened 2022-08-22
Normalization by Evaluation

philzook58:
Maybe this is right.

In a sense it is barely using any egraph powers. It is not clear where other equational theories might go (into Value?).
I suspect the separation between Terms, Values and Neutral values is crucial. Terms are adts, pure syntax. Values are something more semantic and obey interesting equations.

Heavily based on the implementations here:
- https://github.com/AndrasKovacs/elaboration-zoo/blob/master/01-eval-closures-debruijn/Main.hs 
- https://proofassistants.stackexchange.com/questions/900/when-should-i-use-de-bruijn-levels-instead-of-indices
- https://www.andres-loeh.de/LambdaPi/LambdaPi.pdf

### DIFF
--- a/tests/nbe.egg
+++ b/tests/nbe.egg
@@ -1,0 +1,109 @@
+; Terms vs Values
+; Terms are trees. They have no equational theory
+; Values do have equations?
+; https://en.wikipedia.org/wiki/Normalisation_by_evaluation
+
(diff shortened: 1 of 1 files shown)


philzook58:
I think at least this is an interesting example of functional programming. I do not know at this moment where else to go with it

PR #30, opened 2022-08-23
specializing intersections in GJ

yihozhang:


mwillsey:
I think you can prefer to match on the slice itself rather than the length.

PR #32, opened 2022-08-25
proof-of-concept herbie example

dcao:
for actual integration/testing w herbie proper, we're prolly gonna have to do some racket hacking n all that but oh whale


tests/herbie.egg
```diff
@@ -0,0 +1,204 @@
+;; Implements the simplification layer of herbie in egg-smol🫡
+
+;; Bools!
+(datatype B (True) (False))
+(function or (B B) B)
+(rewrite (or (True) x) (True))
+(rewrite (or x (True)) (True))
+(rewrite (or (False) (False)) (False))
```
mwillsey, line 8: Could we replace these with primitive bools (once we add them)? Or are you using their eq-ability?

tests/herbie.egg
```diff
@@ -0,0 +1,204 @@
...
+  (Floor Math)
+  (Round Math)
+  (Log Math))
+
+;; This example has four analyses:
+;; an interval analysis consisting of a hi and lo component,
+;; a constant folding analysis, and a nonzero analysis
+(function hi (Math) Rational :merge (min old new))
+(function lo (Math) Rational :merge (max old new))
+(function evals-to (Math) Rational)
```
mwillsey, line 36: I'm not sure we need an evals-to function, since you should be able to just match on the `Num` constructor in the query, right?

tests/herbie.egg
```diff
@@ -0,0 +1,204 @@
...
+;; To check if something is zero, we just check if it evaluates to zero
+(rule ((= (rational 0 1) (evals-to e)))
+      ((set (is-zero e) (True))))
+
+;; The interval analyses are similar to the constant-folding analysis,
+;; except we have to take the lower/upper bound of the results we get
+(rule ((= e (Add a b))
+       (= la (lo a))
+       (= lb (lo b)))
+      ((set (lo e) (+ la lb))))
```
mwillsey, line 92: In these rules, you could just call `(lo a)` and `(lo b)` in the action. Do you think it's better to do in the query?

mwillsey: Hmm, nevermind, that might not work.

mwillsey (commented):
Nice! Looks like a good start!

tests/herbie.egg
dcao: yeah, if `a` and `b` don't already have a lower bound, this will just panic

dcao:
v0 of the herbie example is rdy! this implements most of the rewrites that herbie's simplification layer uses, along with an interval and non-zero analysis. for testing, i used the test cases that `src/core/simplify.rkt` from the herbie repo uses.

- i didn't implement interval analysis for pow, cbrt, sqrt, or log, since it's unclear how those would work
- i didn't implement some of the more exotic rewrites that herbie uses

tests/herbie.egg
```diff
@@ -0,0 +1,573 @@
...
+          (max (max (* la lb) (* la hb))
+               (max (* ha lb) (* ha hb))))))
+
+(rule ((= e (Div a b))
+       (= la (lo a))
+       (= ha (hi a))
+       (= lb (lo b))
+       (= hb (hi b))
+       (!= lb (rational 0 1))
+       (!= hb (rational 0 1)))
```
mwillsey, line 124: I think these guards aren't needed since division is partial (even in the action).

src/sort/rational.rs
```diff
@@ -40,7 +44,59 @@ impl Sort for RationalSort {
...
+                None
+            }
+        });
+        add_primitives!(eg, "try-cbrt" = |a: R| -> Option<R> {
+            if a.is_one() {
+                Some(R::one())
+            } else {
+                None
+            }
+        });
```
mwillsey, line 96: Why name everything `try-`? I think we can just be ok with the fact that functions are partial.

mwillsey (changes requested):
Couple more changes

src/sort/rational.rs
```diff
@@ -32,15 +34,75 @@ impl Sort for RationalSort {
...
+        add_primitives!(eg, "log" = |a: R| -> Option<R> {
+            if a.is_one() {
+                Some(R::zero())
+            } else {
+                None
+            }
+        });
```
mwillsey, line 81: I think we should delete this or replace the body (or even the `else`) with `todo!()` just to be clear it's a stub. Same with `cbrt` below.

src/sort/rational.rs
```diff
@@ -32,15 +34,75 @@ impl Sort for RationalSort {
...
         add_primitives!(eg, "-" = |a: R, b: R| -> R { a - b });
         add_primitives!(eg, "*" = |a: R, b: R| -> R { a * b });
-        add_primitives!(eg, "/" = |a: R, b: R| -> R { a / b }); // TODO option
+        add_primitives!(eg, "/" = |a: R, b: R| -> Option<R> {
+            if !b.is_zero() {
+                Some(a / b)
+            } else {
+                None
+            }
+        });
```
mwillsey, line 50: Prefer `checked_div` here.

mwillsey (changes requested):
Couple more nits

mwillsey:
Nice work!

PR #33, opened 2022-08-28
Precise/incremental rebuilding

ezrosent:
This is more or less an implementation of the algorithm described in [this gist](https://gist.github.com/ezrosent/dd6f1069d4ce9f08ef359c476a52760c). A few issues came up when I tried to port this over and get tests passing:

* The trick I described around `RawTable` by storing hashcodes in the index only doesn't handle hash collisions correctly. I think it could be extended to support collisions, but there are some subtleties there and I'd prefer to keep the implementation simple to start with. We instead store `Rc<[Value]>` in both the table and the index.
* Examples like `bdd` show how you can insert non-canonical ids into a function. By default, this breaks invariants for the index because we won't catch mentions of these old Ids (they have already been `update`d away). To get around this, we recanonicalize tuples on insertion to a function if they aren't already present.

Both of these changes make the implementation slower, and anecdotally more or less all of the tests take longer to finish with this change enabled. I'd still expect it to help with larger DBs though. If we want to push this further I think there's a long list of optimizations we could add to cut the overhead here.

### DIFF
--- a/src/extract.rs
+++ b/src/extract.rs
@@ -45,7 +45,8 @@ impl EGraph {
                     return vec![];
                 }
                 assert!(func.schema.output.is_eq_sort());
-                func.nodes
+                func.data
+                    .nodes
                     .iter()
                     .filter_map(move |(inputs, output)| {
                         (output == output_value).then(|| {
@@ -118,7 +119,7 @@ impl<'a> Extractor<'a> {
             for &sym in &self.ctors {
                 let func = &self.egraph.functions[&sym];
                 if func.schema.output.is_eq_sort() {
-                    for (inputs, output) in &func.nodes {
+                    for (inputs, output) in &func.data.nodes {
                         if let Some(new_cost) = self.node_total_cost(&func.schema.input, inputs) {
                             let make_new_pair = || (new_cost, Node { sym, inputs });
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -34,10 +34,12 @@ use util::*;
 
 use crate::typecheck::TypeError;
 
+type ArgPtrSet = HashSet<Rc<[Value]>>;
+
 #[derive(Default, Clone, Debug)]
 struct FunctionData {
     nodes: HashMap<Rc<[Value]>, Value>,
-    index: HashMap<Id, HashSet<Rc<[Value]>>>,
+    index: HashMap<Id, ArgPtrSet>,
 }
 
 impl FunctionData {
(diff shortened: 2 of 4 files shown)


ezrosent:
Hmm... not sure what's going on with the build, the eqsolve test does pass on my machine.

mwillsey:
It doesn't work on my machine either. Could the new algorithm is missing some canonicalizations perhaps?

ezrosent:
Glad it doesn't work on your machine! I figure it's an issue along those lines but debugging using only the actions is pretty painful as the iteration time is 4-5 minutes.

I'd like to reproduce this on my machine. I've tested it on an intel and an arm mac; no dice. I've checked that my version of `eqsolve.rs` is the same as the one currently on main. I've been running tests by doing `cargo test` and `cargo run tests/eqsolve.rs` with and without release.

Any sources of nondeterminism you can think of? This is with a recent version of rust (1.65)

mwillsey:
Weird... `cargo test` fail on eqsolve, but `cargo run tests/eqsolve.rs` works but is awfully slow, takes like 20 seconds.

ezrosent:
Is cargo test giving you a similar error to the github actions build? 

```
thread 'test_files' panicked at 'Top level error: Check failed: Value { tag: "Expr", bits: 584838 } != Value { tag: "Expr", bits: 861568 }', tests/files.rs:39:29
```

And yeah, I noticed some slowness in debug builds here as well. I think there are low hanging fruit for optimization but I wanted to get things passing first.

ezrosent:
Amazing!! that worked. I'll try and get a fix pushed now that I can reproduce; thanks for digging into this!

I'm surprised it worked though, does that mean we keep state around between tests, and that tests are order-dependent?

mwillsey:
I'm guessing the state is coming from the `symbol_table` library that I wrote, that globally caches symbols.

ezrosent:
Tests pass now. I reworked the algorithm a bit to reduce precision and only build the index during rebuilding. It's still incremental in the sense that new merges discovered during rebuilding don't force you to re-scan the whole database, but you always do a full scan after each run.

This makes it more self-contained and lowers the up-front cost of this setup because you don't have any index maintenance. I think it may lower the _overall_ cost as well if you amortize across insertions into the db ahead of rebuilding, but I haven't done the math there.

In any case, performance is now much closer to `main,` but none of the current benchmarks seem to benefit from incremental rebuilding (even when cranking up the 'run' numbers). I think the next two things to do here are:

* Run a test case stressing the incremental case (same as in the gist), to confirm there is some workload that this helps for in the actual codebase.
* Implement a couple optimizations to reduce the overhead of index maintenance somewhat.

ezrosent:
Confirmed we still see better scaling in the initial benchmark. With the more optimized code I think the overhead is still there, but it's much lower for smaller databases. Most workloads are still a bit slower, but doing `run 10` in `eqsolve` takes 30s at head and 26s with incremental. In that case rebuilding is 6-7 seconds faster, which makes up for applies being 2-3 seconds slower. Pretty mixed on the whole; doesn't look like we've found a workload "in the wild" that benefits much from rebuilding incrementally.

mwillsey:
Very interesting. I agree we should defer doing this until we have good evidence that it's worth it. In egg, we eventually got to where precise rebuilding was just faster than "simple" rebuilding, but I think the overhead might just be lower there.

PR #34, opened 2022-08-30
Porting the math test suite to egg-smol

yihozhang:
Right now the two rules below will blow things up. My speculation is that these two rules make AC rules to trigger too frequently.
```
;; (rule ((MathU a)) ((union a (Add a (Const (rational 0 1))))))
;; (rule ((MathU a)) ((union a (Mul a (Const (rational 1 1))))))
```


tests/math.egg
```diff
@@ -70,6 +70,10 @@
 (rewrite (Mul a b) (Mul b a))
 (rewrite (Add a (Add b c)) (Add (Add a b) c))
 (rewrite (Mul a (Mul b c)) (Mul (Mul a b) c))
+;; These two rules are not present in the original egg tests,
+;; but are necessary for the last check
+(rewrite (Add (Add a b) c) (Add a (Add b c)))
+(rewrite (Mul (Mul a b) c) (Mul a (Mul b c)))
```
mwillsey, line 76: Why are these needed? They seem redundant.

mwillsey: Ok, let's remove these before merging.

tests/math.egg
```diff
@@ -0,0 +1,178 @@
...
+(rewrite (Sub a b) (Add a (Mul (Const (rational -1 1)) b)))
+(rewrite (Div a b) (Mul a (Pow b (Const (rational -1 1)))) :when ((is-not-zero b)))
+
+(rewrite (Add a (Const (rational 0 1))) a)
+(rewrite (Mul a (Const (rational 0 1))) (Const (rational 0 1)))
+(rewrite (Mul a (Const (rational 1 1))) a)
+
+;; These two rules can make the first goal time out
+;; (rule ((MathU a)) ((union a (Add a (Const (rational 0 1))))))
+;; (rule ((MathU a)) ((union a (Mul a (Const (rational 1 1))))))
```
mwillsey, line 87: We discussed a fix for these in slack. These just need some additional guards to work.

tests/math.egg
```diff
@@ -0,0 +1,178 @@
...
+
+(define end-expr (Const (rational 1 1)))
+(check (= start-expr end-expr))
+
+(clear)
+
+;; math_simplify_factor
+(define start-expr (Mul (Add (Var "x") (Const (rational 3 1)))
+                        (Add (Var "x") (Const (rational 1 1)))))
+(run 7)
```
mwillsey, line 172: See above. Need to increase run to 8 here.

mwillsey (changes requested):
Looks good, ready to merge after these changes. Can you also delete `math-ast.egg` in this commit? It seems redundnant now.

PR #39, opened 2022-09-20
Unification-based pointer analysis

yihozhang:


mwillsey:
@yihozhang is this ready to merge? If so, can you make it not a draft?

yihozhang:
@mwillsey Yes this is a self-contained example, although a bit different from one in the paper

PR #44, opened 2022-10-17
Matrix Expressions

philzook58:
And interesting example of the utility of rules rather than rewrites is the guarding of certain rewrite rules upon correct typing conditions. Many of these guards can be removed under the assumption that only well typed expressions go into the egraph, but not all. The removal of these guards is a significant conceptual and performance advantage over encoding to other more systems like ATP or SMT. One such example is the interchange law, expressed here in matrix terminology. This algebraic manipulation is genuinely useful, turning for example a (NM)^3 operation into a N^3 + M^3 + (NM)^2 operation by pushing matrix multiplication in on smaller operations.
This particular example generalizes to other domains.

There is lots of other fun stuff in matrix algebra. It's a very rich thing.


tests/matexpr.egg
```diff
@@ -0,0 +1,123 @@
...
+ ;(= (type B) (Mat ra ca)) 
+ ;(= (type C) (Mat rc cc)) 
+ ;(= (type D) (Mat rd cd)) 
+ ;(= ca rc)
+ ;(= cb rd) ; implied
+ )
+  ((set 
+   (Kron (MMul A C) (MMul B D))
+   E
+  )))
```
mwillsey, line 82: Consider using the guarded rewrite syntax here: `(rewrite foo bar :when (cond1 cond2 ...))` expands to
```
(rule ((= x foo) cond1 cond2 ...) ((union x bar)))
```

PR #47, opened 2022-10-28
Improving rebuilding

yihozhang:
This is a low hanging fruit in improving rebuilding. Since right now the database is created afresh during each rebuilding, most time is spent on inserting tuples into a new database. However, we only need to process those that are changed.

Before:
```
Geomean egglog/souffle: 0.5349996183864182
Total egglog/souffle: 0.687601368899584
```
plot-before

After:
```
Geomean egglog/souffle: 0.4056677316834581
Total egglog/souffle: 0.4405201946577985
```

plot

~Also, there are still many opportunities for optimization: the rebuilding in this PR uses many small vectors. ~1/2 time of the rebuilding is spent on malloc/free.~ Not the case after using smallvecs with the latest main

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -98,21 +98,48 @@ impl Function {
 
         // FIXME this doesn't compute updates properly
         let n_unions = uf.n_unions();
-        let old_nodes = std::mem::take(&mut self.nodes);
-        for (mut args, out) in old_nodes {
+        let mut to_delete = vec![];
+        let mut to_add = vec![];
+        for (args, out) in &self.nodes {
             let mut new_timestamp = out.timestamp;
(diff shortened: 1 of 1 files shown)


yihozhang:
Oops some tests are failing

yihozhang:
Okay I fixed the test, but the speedup turns out to be insignificant :(
```
Geomean egglog/souffle: 0.533196396959818
Total egglog/souffle: 0.6305961777457302
```

mwillsey:
This could be a good fit for [`HashMap::retain`](https://doc.rust-lang.org/stable/std/collections/hash_map/struct.HashMap.html#method.retain).

PR #48, opened 2022-10-31
fix skolemization in type inference

yihozhang:


mwillsey:
@yihozhang sorry this got a little stale. Is this ready to merge?

yihozhang:
The example is ready to be merged, but I remembered you also added map Diff to main, so it may be more convenient to just copy the type inference file and make a commit to main

yihozhang:
See #74

PR #49, opened 2022-11-03
Removed print to stdout except from print commands

philzook58:
I don't know how IO pure run_command should be. From the perspective of building external tools that communicate to egg-smol via stdin stdout, it would probably be nice to make these errors s-expressions also.


mwillsey:
I think a goal should be to keep stdout clean and s-expr-y, and dump the rest to stderr.

PR #51, opened 2022-11-03
fix define in action

yihozhang:
The action abstract machine is not popping defined local variables right now, which causes panic


mwillsey:
What does this break? Seems like popping the stack shouldn't be necessary since you could just reset it when you need it?

yihozhang:
This [assert](https://github.com/mwillsey/egg-smol/blob/main/src/lib.rs#L735) was failing for me

PR #52, opened 2022-11-07
Choice Domain Example

philzook58:
A simple example demonstrating the possibility of emulating the Souffle choice-domain feature https://souffle-lang.github.io/choice
The soundness of this representation is still up for debate.

### DIFF
--- a/src/sort/string.rs
+++ b/src/sort/string.rs
@@ -24,6 +24,12 @@ impl Sort for StringSort {
         self
     }
 
+    #[rustfmt::skip]
+    fn register_primitives(self: Arc<Self>, eg: &mut EGraph) {
+
+        add_primitives!(eg, "const" = |a: Symbol, _b: Symbol| -> Symbol { a });
+    }
+
     fn make_expr(&self, value: Value) -> Expr {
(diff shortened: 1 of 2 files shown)


yihozhang:
Can you define the `const` function using egglog? Something like
```
(function const (Expr Expr) Expr)
(rewrite (const a b) a)
(define choice-function (Expr) Expr :merge (const old new))
```

philzook58:
Oh, interesting. I suppose I can. In my case I was using primitives though. I don't think I can do it over primitives. the `const` table would need to make set and  then discover the rewrite applies. It also materializes the const table. This seems to be a common tradeoff. primitives are kind of a pain in that you need to revert to datalog style programming instead of rewrite style programming, but you don't materialize as much junk.

mwillsey:
You can just say `:merge old`.

mwillsey:
This example seems basically the same as the path examples, so I'm going to close to avoid clutter for now.

philzook58:
Sure. This is demonstrating a different point to the previous path examples. This computes a spanning tree even in the presence of loops in the edge graph. This is unusual for a datalog.

PR #55, opened 2022-11-08
Egglog Isn't Prolog

philzook58:
Egglog can emulate prolog. There is no question there. Egglog has pure functional programming as a subsystem, and you can write a purely functional prolog interpreter.

A measure I think of how close egglog is to prolog is to attempt to build as shallow and simple interpreter of prolog as possible.

In order to pinpoint where I think Egglog fails to capture the essence of prolog, despite having horn clauses and union find, I wrote one kind of crippled prolog interpreter. I'll note that it is very, very difficult if not impossible to even shallowly encode Harrop clauses or lambda terms into a regular prolog without essentially writing an interpreter. It is a constant temptation to attempt to use one notion of variable for another notion and it rarely works out well.

The interpreter carries a goal stack. Egglog inherits from datalog that it is pattern matching based and not unification based like prolog. It is difficult to see how to shallowly embed unification to the heads of rules.

When you write a prolog interpreter in prolog https://www.metalevel.at/acomip/, you can borrow so many capabilities of the host language that it is trivial. You can make each of the capabilities in turn more or less explicit depending on how you want to modify the stock behavior.

The two main things (the essential components of prolog) to try to borrow from egglog are unification and nondeterminism. 

Basically, you can't borrow both in egglog to achieve prolog behavior. You can pick one, maybe.

Borrowing egglog nondeterminism, you can use non deterministic rewrite rules in the interpreter picking what search path to go down. However, now if you ever unify something it happens across all branches, which is incorrect. You can perhaps fix this by carrying around a first class union find in the interpreter.

If you instead make the nondeterminism explicit by carrying around a choice point stack in the interpreter, you can perhaps use skolemization and egglog union to do some aspects of unification, although the actual unification algorithm will still need to be encoded in further rules.
 
Note that that syntactic Prolog variables `X` are not just one semantic thing. 

Prolog consists of a query pass working backward from goals, searching for proofs. Upon finding the proof, there is an implicit forward pass returning an answer (something like a coroutining).

A unification variable in a goal/query is attempting to find a proof to an existential question.
A unification variable in the answer is the assertion of a universally quantified statement.
A unification variable in the head of a program clause is a universal quantified variable for that axiom.
A unification variable appearing only in the body of a program clause is a existentially quantified variable.

Datalog has no built in notion of query. It only has the forward pass of answers. Magic set can encode queries, but encoding is not shallow native support.

Egglog variables tend to feel mostly like existential answers
Egglog can kind of sort of assert universal facts by deriving them about fresh constants, but this is fishy, and then you cannot use those facts without a substitution operation, which is a fairly deep encoding at that point

### DIFF
--- a/tests/prolog.egg
+++ b/tests/prolog.egg
@@ -1,0 +1,84 @@
+
+; Prolog modulo equality.
+; This is prolog that supports only fully grounded queries.
+; In that sense it is not a prolog at all
+
+
+
+
+(datatype Term
+    (TSucc Term)
+    (TZero)
+)
+(datatype Goal
+    (GAdd Term Term Term)
+    (GLTE Term Term)
+    ; GAppend
+)
+(datatype GoalList (Cons Goal GoalList) (Nil))
+(datatype Proof (True))
+
+(function append (GoalList GoalList) GoalList)
+(rewrite (append Nil l) l)
+(rewrite (append (Cons g gs) l) (Cons g (append gs l)))
+
+(function resolve (GoalList) Proof)
+
+
+; The head-body relation could be inlined perhaps into it's use site
+(relation head-body (Goal GoalList))
+
+(set (resolve Nil) True)
+; We reuse nondetermistic rewriting to achieve the proplog search 
+(rewrite (resolve (Cons g gs)) (resolve (append body gs))
+    :when ((head-body g body))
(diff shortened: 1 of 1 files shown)


philzook58:
I'm going to keep this one (and some other closed requests and things I've never even mentioned) to be refined in my scratchpad branch https://github.com/philzook58/egg-smol/tree/scratchpad. There is certainly more to say on the relationship of egglog and prolog.

PR #56, opened 2022-11-08
Create public define function

saulshanabrook:
This PR makes extracts defining an expression into a public function.

I am working on Python bindings for this package and wanted to expose this in Python, for example, [to recreate the `eqsat_basic` example](https://egg-smol-python.readthedocs.io/en/latest/explanation/compared_to_rust.html#low-level-bindings-api).

Alternatively, `eval_expr` and  a "get function" methods could be exposed to recreate the same logic.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1033,7 +1033,12 @@ impl EGraph {
         })
     }
 
-    pub fn define(&mut self,  name: Symbol, expr: Expr,cost: Option<usize>) -> Result<ArcSort, Error> {
+    pub fn define(
+        &mut self,
+        name: Symbol,
+        expr: Expr,
+        cost: Option<usize>,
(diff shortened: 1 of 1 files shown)


mwillsey:
Looks like a formatting failure. You may want to configure your editor to automatically format on save.

saulshanabrook:
It looks like it might be useful to expose `eval_expr` anyways to support recreating the `extract` command which uses it to get a value from an expression. So it might be better to actually expose that along with an insert into function method?

mwillsey:
Yep, feel free to expose those. One thing to watch out for: exposing getting a function (or inserting into directly) isn't that useful, because many operations on functions (including insertion) require access to other stuff in the EGraph, like the unionfind.

PR #62, opened 2022-11-18
Make more of the API public

saulshanabrook:
This PR is a followup to #56, which makes more of the API public. There is now enough made public to expose all the same functionality as exposed in the text command interface through Python. The corresponding Python PR is https://github.com/metadsl/egg-smol-python/pull/4.

I opted to not expose the `Value`s to Python at all, relying on these higher-level function calls instead. I saw that there was a comment on the `Value` to make it private at some point.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1033,7 +1033,12 @@ impl EGraph {
         })
     }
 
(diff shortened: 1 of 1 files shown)


mwillsey:
That's correct. I'm not quite happy with Values just yet. Nice PR!

PR #63, opened 2022-11-20
Extract declaring a sort alias into a pub fn

saulshanabrook:
This PR is another follow-up to #56  and #62, which extract the behavior to name a sort, like `(sort StringSet (Map Var i64))`. The function `declare_sort` was already taken, to add a user-defined sort, so I wasn't sure what to call this. Happy to rename if you have other suggestions.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -332,7 +332,7 @@ pub struct NotFoundError(Expr);
 
 impl EGraph {
     pub fn add_sort<S: Sort + 'static>(&mut self, sort: S) {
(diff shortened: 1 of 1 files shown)


mwillsey:
Perhaps combine it with declare_sort? It could take an option Expr or something.

saulshanabrook:
@mwillsey Sounds good. I also added error handling for it not finding a presort and changed the `declare_sort` to re-use the  `add_arcsort` method which duplicates some of the logic there. I moved the error handling into `add_arcsort` to support this. 

Let me know how this looks!

PR #65, opened 2022-11-20
Z3 as a sort

philzook58:
This is an initial pass at making Z3 available from inside egglog. I thought it wise to seek comments / advice before continuing to fill it out because I had to perform violence in a couple places.
The high level z3 bindings expose a lifetime parameter that associates the datatype with the lifetime of the z3 context. I found it hard/impossible to get that to work while keeping an indexset of the z3 asts inside the Z3Sort.
There are a couple choices:
1. Trying to thread `Z3Sort<'ctx>`. This was not working. Since Z3Sort needs to hold a reference to Context, there is a circularity in the data structure.
2. Rebuild the high level z3 bindings from z3_sys to allow for hashmap storage of asts. This will largely be highly replicated from what the z3 crate already does
3. Use 'static to remove the lifetime parameter from z3::ast::Dynamic. mem::transmute will be needed at a few certain spots because the context is not actually static. I don't see how these asts could escape the lifetime of the context as used here, but this doesn't feel that good. There could also perhaps be a memory leak.
4. Actually have a static global Z3 context available something like Symbol.
5. Don't use bindings at all. Construct ast/s-expressions and call smt solver via a string interface. This has the advantage of safety, parallelism, and extending to solvers beyond z3. Replicatng an smtlib ast is tiresome though. This also may enable using  regular egglog datatypes.

Some other considerations:
1. Is it better to have a sort `Z3Sort`, `Z3Bool`, or `(Z3 bool)`. One does want to define user defined sorts in z3 eventually, which it is hard to see how that latter two could support.
2. Should one always call simplify before hashconsing the z3 expression. This may improve termination and deduplicate the database.
3. Parallelism and non blocking. check-sat calls are going to become very expensive. I note that an interesting possible capability of egglog is using eids as promises/futures from external systems.
4. I also added unsafe Sync and Send traits for the Z3Sort. I am uncertain what Z3's promises are in regards to thread safety nor how to protect it if it isn't thread safe. My understanding is that egg-smol is currently not multi-threaded.
5. z3 should be guarded under a feature at least since it will probably be too annoying to get the wasm build to work. Or perhaps some mechanism to install third party sorts without having to fork the code base.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ regex = "1"
 # binary dependencies
 clap = {version = "4", features = ["derive"]}
 env_logger = "0.9"
+z3 = "0.11.2"
 
 [build-dependencies]
 lalrpop = "0.19.7"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -321,6 +321,10 @@ impl Default for EGraph {
         egraph.add_sort(StringSort::new("String".into()));
         egraph.add_sort(I64Sort::new("i64".into()));
         egraph.add_sort(RationalSort::new("Rational".into()));
+        egraph.add_sort(Z3Sort::new(
+            "Z3Sort".into(),
+            StringSort::new("String".into()),
+        ));
         egraph.presorts.insert("Map".into(), MapSort::make_sort);
         egraph
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -318,13 +318,11 @@ impl Default for EGraph {
             seminaive: true,
         };
         egraph.add_sort(UnitSort::new("Unit".into()));
-        egraph.add_sort(StringSort::new("String".into()));
+        let string_sort = Arc::new(StringSort::new("String".into()));
+        egraph.add_arcsort(string_sort.clone());
         egraph.add_sort(I64Sort::new("i64".into()));
         egraph.add_sort(RationalSort::new("Rational".into()));
-        egraph.add_sort(Z3Sort::new(
-            "Z3Sort".into(),
-            StringSort::new("String".into()),
-        ));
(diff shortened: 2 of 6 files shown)


mwillsey:
I would prefer to keep this as a separate crate until we think it's applicable to package it with egglog under a feature. That said, egglog should definitely have featureful enough extensions to do this, so we should make enough things public that it's possible to do that. 

Regarding the options above, I think the best thing think is to hide the lifetime and just use 'static and some unsafe code. This is a pretty common pattern, and it's a little cleaner than mucking around with raw pointers. We'll have to do something to prevent these things from moving across threads though.

philzook58:
Makes sense. I'll close this.

A thing I don't like that much is if I make this an external crate that uses egg-smol as a library, that crate will need to create it's own copy of the executable already in egg-smol. Then for further capabilities, you'll need to write wrapper executables that add in all the capabilities you want.

The design space as I understand it is:
1. egg-smol accepts all kinds of weirdo extras. This creates a lot of maintenance burden and bloats dependencies, but also is uncomplicated and kitchen sink for users. Extras could be guarded under cargo.toml features.
2. Some mechanism (?) in egg-smol to dynamically load extra features (by finding some kind of .so on a path perhaps? importing some wasm file with some kind of interface?). Souffle user-defined functors are somewhat like this.
3. egg-smol as a library and external crate libraries and then if you want to combine features, you need to write a small executable wrapper project. This is actually fairly simple, but maybe to make it even easier some of the functionality in main.rs should be pushed into the library to avoid duplication. The more I think about it, this is probably the right approach. If someone wants to, they can maintain a kitchen sink distro.

mwillsey:
I like option 3. It’s the closest to what egg did as well.

PR #69, opened 2022-11-29
Report Code Coverage for Tests

saulshanabrook:
This PR changes the tests to record their code coverage, using the [Tarpaulin](https://github.com/xd009642/tarpaulin) tool and [source-based coverage](https://blog.rust-lang.org/inside-rust/2020/11/12/source-based-code-coverage.html) (https://github.com/xd009642/tarpaulin/issues/549).

## Local Usage

Now when running `make test` it will record the code coverage, and print that out:

```
Nov 29 13:02:10.910  INFO cargo_tarpaulin::report: Coverage Results:
|| Tested/Total Lines:
|| src/ast/expr.rs: 15/40 +0.00%
|| src/ast/mod.rs: 26/26 +0.00%
|| src/extract.rs: 58/76 +0.00%
|| src/gj.rs: 243/270 +0.00%
|| src/lib.rs: 482/616 +0.16%
|| src/main.rs: 0/30 +0.00%
|| src/sort/i64.rs: 31/31 +0.00%
|| src/sort/macros.rs: 18/20 +0.00%
|| src/sort/map.rs: 142/160 +0.00%
|| src/sort/mod.rs: 11/13 +0.00%
|| src/sort/rational.rs: 70/70 +0.00%
|| src/sort/string.rs: 13/15 +0.00%
|| src/sort/unit.rs: 15/19 +0.00%
|| src/typecheck.rs: 313/395 +0.25%
|| src/unionfind.rs: 83/131 +0.00%
|| src/util.rs: 8/12 +0.00%
|| src/value.rs: 12/12 +0.00%
|| 
79.55% coverage, 1540/1936 lines covered, +0.10% change in coverage
```

It will also create a `tarpaulin-report.html` file, which can be opened to see the coverage line-by-line.
Screenshot 2022-11-29 at 1 24 13 PM

I was also able to install the [Coverage Gutters](https://marketplace.visualstudio.com/items?itemName=ryanluker.vscode-coverage-gutters) VS Code extension, which picked up the coverage file and showed the information inline:

Screenshot 2022-11-29 at 1 04 22 PM

## Codecov

It uploads the results to [Codecov](https://about.codecov.io/) so that it is easy to see over time how coverage evolves, and also for each PR how it impacts the code coverage. This is useful when refactoring or implementing new features, to make sure all branches are covered.

I enabled it on my fork, to test it out, which you can browse [here](https://app.codecov.io/github/saulshanabrook/egg-smol/blob/coverage/src/lib.rs):

Screenshot 2022-11-29 at 1 26 35 PM

In order to upload to Codecov, I found I had to set the token, otherwise, I would get a 404 error. This is a known bug with Codecov (https://github.com/codecov/codecov-action/issues/557, https://github.com/codecov/codecov-action/issues/837). I made the token public, instead of setting it privately so that it would be used on any PRs from forks as well (https://github.com/codecov/codecov-action/issues/557#issuecomment-1306088202).

**Before merging this, if you choose to keep this approach, you would have to make an account on Codecov, add this project, and update the token in this PR with your token for this repo**. Making the token public does present some security risk (anyone would be able to submit coverage for this repo) which would have to be weighed against the advantage of allowing PRs from forks to have their coverage submitted (and compared).

## Alternative using grcov

I did want to note that I did try using the `RUSTFLAGS='-Cinstrument-coverage'` env variable directly and then using [grcov](https://github.com/mozilla/grcov/) to convert to lcov and HTML, instead of using Tarpaulin. This does have support for branches, which is nice, but I found the lcov generation to inaccurate (https://github.com/mozilla/grcov/issues/860).

It also requires installing the `llvm-tools-preview` component, since it uses that to do the coverage generation, whereas Tarpaulin reimplemented what they needed in rust.

## Performance

I also noticed that using this coverage increased the testing time from ~2 second to  ~13 second on my local machine, once things were cached. The alternative above using `instrument-coverage` directly takes ~5 seconds.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
       uses: Swatinem/rust-cache@v2
 
     - name: Build
-      run: make test coverage-lcov
+      run: make test
     - name: Upload to codecov.io
       uses: codecov/codecov-action@v3
       with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,6 +2,8 @@ name: Build
 
 on: [push, pull_request]
 
+env:
+  CODECOV_TOKEN: 9ce1d224-e986-4e7c-8ed1-5bdcce684f7c
 jobs:
   build:
     runs-on: ubuntu-latest
@@ -17,4 +19,4 @@ jobs:
       uses: codecov/codecov-action@v3
       with:
         files: ./target/debug/lcov.info
-        fail_ci_if_error: true
+        token: ${{ secrets.CODECOV_TOKEN }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,4 +12,8 @@ jobs:
       uses: Swatinem/rust-cache@v2
 
     - name: Build
-      run: make test
+      run: make test coverage-lcov
+    - name: Upload to codecov.io
+      uses: codecov/codecov-action@v3
+      with:
+        files: ./target/debug/lcov.info
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,5 +18,5 @@ jobs:
     - name: Upload to codecov.io
       uses: codecov/codecov-action@v3
       with:
-        files: ./target/debug/lcov.info
+        files: ./lcov.info
         token: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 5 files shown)


mwillsey:
Thanks for the PR! This a little more than I want to commit too at the moment. But it's great to have this PR standing by once we stabilize some more!

saulshanabrook:
@mwillsey Sounds good! I was surprised that it worked quiet easily, it just took a while to figure out the right way to do it.

PR #71, opened 2022-12-05
Columnar Indexes for incremental rebuilding and faster trie construction

ezrosent:
Refactor of core data-structures. The main changes are:

* Simpler union-find implementation; we weren't using the `UnionFindLike` trait or other union-find implementations, so this PR goes back to a structure closer to egg's.
* Refinement of `Function`'s `nodes` table. Entries are now sorted by timestamp. To handle in-place removals, entries are now tagged with the timestamp at which an entry becomes _stale_. There's a bit of machinery here to make invalid entries not perturb lookups into the map. Stale entries are GC'd if they constitute a sufficiently large portion of the overall table. The upshot of this structure is that we can estimate the size of a timestamp range in O(log(n)) time: this is a big deal for some workloads, even when considering that this implementation now overestimates size due to stale entries.
* The union-find implementation now keeps track of "dirty" ids that have recently ceased to be canonical. These ids are recorded on a per-sort basis.
* Columnar indexes: we keep an index on each column for every function. This, along with the extra bookkeeping in the union-find, lets us rebuild incrementally.
* LazyTrie now uses columnar indexes as its first level if the index contains a superset of the content in the `TrieAccess` and if the index does not contain "too many" entries outside of the given access range.

With these changes, a simple transitive closure benchmark on a line graph with 4096 nodes goes from ~144s to ~8s on my MacOS ARM desktop. The remaining time is mostly spent in allocation and deallocation for tries, so this PR also slots in `jemalloc`. With the new default allocator, the TC benchmark finishes in a bit under 5s.


ezrosent:
Looks like the new typeinfer test breaks doesn't pass. I'll try to get it passing; looks like a low level index invariant is getting broken here (other tests preserve it).

ezrosent:
One issue I've noticed is that this implementation of rebuild is noticeably slower than the last one if you are rebuilding the whole table (which seems to happen in some AC-style workloads). I think it'll help to add back an O(n) version along the lines of the current implementation on main that we can use if we notice that we're rebuilding a noticeable fraction of the table.

mwillsey:
How much slower? Is it fundamental? Less code is always better for sure.

ezrosent:
It's ~5x slower I think? I was able to get ~2-3x slowdown by running `eqsolve.egg` for 12 iterations. Basically all of that was rebuilding.

ezrosent:
rough mock-up is at [this branch](https://github.com/ezrosent/egg-smol/tree/hybrid_rebuild). That brings `eqsolve.egg` with 12 iterations down to 6.2s from 14s. Main does 4.8s right now. There's a flaking test on that branch, but the bug seems fixable if we want to take this route. I agree it's not that nice..

The high-level reason why incremental rebuilds are slower when you're rebuilding most of the table is that non-canonical entries are marked as stale, which is an extra remove+insert into the table on top of the insert that you do in the standard rebuild. A custom data-structure would allow us to cut the overhead to an extent, but I do think there's an inherent tradeoff here.

You'd hope that index maintenance could make up for this slowdown, but if you're rebuilding the entire table then you will also be rebuilding indexes from scratch a lot. Because we build _all_ the indexes with this change, that makes the rest of the code a bit slower; `main` just builds the indexes it needs on a per-query basis.

In other words, I think this particular query feels like something of a "worst case" for incremental processing. There's actually very little work to be done incrementally, so all the extra tracking is a just weighing performance down. I suspect most workloads won't look like this, but I think we'll want to keep this sort of issue in mind.

ezrosent:
Alright! Wrote up a new "table" implementation that seems a good bit faster than indexmap for our use-case; rebuilding is much more efficient. We still have the "same" rebuild implementation for everything, but we still use some heuristics to decide when to give up and just rebuild everything.

With those changes, this benchmark is a little slower (5.4s vs. 4.8s), but most of that change is explained by index maintenance.

ezrosent:
Couple of other things to note with the tables here:
* We still need a solution for "keeping values around" for proofs. I think having a table 'on the side' would work. I tried just not removing values at all and things were a lot slower.
* These tables aren't "flat": they just store vectors in the table. Unspooling these vectors is doable, but we rely on per-tuple metadata around staleness and timestamps. Supporting these is doable, but I think the code will be sufficiently subtle that it's worth putting in a separate PR.

ezrosent:
Took a stab at "flat tables" in [this branch](https://github.com/ezrosent/egg-smol/tree/flat_table)

PR #76, opened 2022-12-08
Calc command

philzook58:
[Dafny](http://dafny.org/dafny/DafnyRef/DafnyRef.html#sec-calc-statement) and [Lean](https://github.com/leanprover-community/leanprover-community.github.io/blob/newsite/templates/extras/calc.md) both have a notion of a calculational proof. Agda also has a style like this.
This is a proof command that allows one to break up a proof query with user given hints.
```
(calc ((a G) (b G))
    (* (* b (* (inv a) a)) (inv b))
    (* b (inv b))
    (I)
)
```
It works by 
1. pushing the egraph
2. inserting fresh constants (eids) as `a` and `b`. This is a way of asking a universally quantified question.
3. inserting pairs of subsequent entries of the calc command.
4. running until the two are found to be equal.
5. popping all of this out of the egraph

When I was trying to debug why my category theory proofs in metatheory.jl were not going through, I found an analogous command that I made there completely indispensable to manually bisect to the failure point.

It could also work in a unidirectional mode in the future by only inserting the first of the pair and seeking the second. This would also support calc in the `->` implication mode (inserting a fact and seeing when the consequent is derived).

This could possibly be a macro.

The way the universal quantification works here is a little goofy. I believe it is sound, but the way shadowing and scoping works is not typical. You may not make a previously defined name. For example, the following will error out. 
```
(calc ((I G) (b G)) ; error because I is previously defined in file
    (* (* b (* (inv a) a)) (inv b))
    (* b (inv b))
    (I)
)
```
Possibly in the future there needs to be a reserved internal namespace, or the ability to locally shadow tables.

A possible option that might be nice is an `:assert` flag that asserts the calc lemma as an axiom after it is proven.

Another proof command I'm excited for is a related lambda prolog-lite top down prover.

 The calc command is also a possible location for trivial parallelism

`declare_const` is an extremely limited form of raw make-set basically. It is relatively unproblematic compared to full make-set since the ids it makes are in correspondence with names and not currently available in rule actions. It is the analog of `declare-const` in smt and perhaps should be user exposed as such. It avoids the need to perform the following construction, which requires generating a name `temp_a` which needs to avoid clashing with user names.

```
(function temp_a () G)
(define a (temp_a))
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1110,25 +1110,23 @@ impl EGraph {
         }
         // Insert each expression pair and run until they match.
         for ab in exprs.windows(2) {
-            if let [a, b] = ab {
-                self.push();
-                *depth += 1;
-                self.eval_expr(a, None, true)?;
-                self.eval_expr(b, None, true)?;
-                let cond = Fact::Eq(vec![a.clone(), b.clone()]);
-                self.run_command(
-                    Command::Run(RunConfig {
-                        limit: 100000,
-                        until: Some(cond.clone()),
-                    }),
-                    true,
-                )?;
-                self.run_command(Command::Check(cond), true)?;
-                self.pop().unwrap();
-                *depth -= 1;
-            } else {
-                panic!();
-            }
+            let a = &ab[0];
+            let b = &ab[1];
+            self.push();
+            *depth += 1;
+            self.eval_expr(a, None, true)?;
+            self.eval_expr(b, None, true)?;
+            let cond = Fact::Eq(vec![a.clone(), b.clone()]);
+            self.run_command(
+                Command::Run(RunConfig {
+                    limit: 100000,
+                    until: Some(cond.clone()),
+                }),
+                true,
+            )?;
+            self.run_command(Command::Check(cond), true)?;
+            self.pop().unwrap();
+            *depth -= 1;
         }
         self.pop().unwrap();
         *depth -= 1;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -934,40 +934,12 @@ impl EGraph {
                 }
             }
             Command::Calc(idents, exprs) => {
-                let mut iter = exprs.into_iter();
-                // If it errors out, I'm in a bad pop state.
-                let mut depth = 0;
-                if let Some(a) = iter.next() {
-                    self.push();
-                    depth += 1;
-                    for TypeBind { ident, sort } in idents {
-                        let sort = self.sorts.get(&sort).unwrap().clone();
-                        self.declare_const(ident, &sort)?;
-                    }
-                    let mut a = a;
-                    for b in iter {
-                        self.push();
-                        depth += 1;
-                        self.eval_expr(&a, None, true)?;
-                        self.eval_expr(&b, None, true)?;
-                        let cond = Fact::Eq(vec![a, b.clone()]);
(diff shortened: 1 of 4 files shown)


tests/calc.egg
```diff
@@ -0,0 +1,32 @@
...
+(define A8 (* A4 A4))
+
+; if you remove :until, this will take a very long time
+(calc ()
+    (* A4 A4)
+    (* (* A2 A2) (* A2 A2))
+    (* A2 (* A2 (* A2 A2)))
+)
+; If you need multiple stop conditions, consider using a (relation relation stop (unit))
+; With rules filling it in with different stop conditions of interest.
```
mwillsey, line 25: Vestigal comments from a previous PR perhaps?

src/lib.rs
```diff
@@ -1087,6 +1095,62 @@ impl EGraph {
...
+                let cond = Fact::Eq(vec![a, b.clone()]);
+                self.run_command(
+                    Command::Run(RunConfig {
+                        limit: 100000,
+                        until: Some(cond.clone()),
+                    }),
+                    true,
+                )?;
+                self.run_command(Command::Check(cond), true)?;
+                self.pop()?;
```
mwillsey, line 1129: unwrap this pop, it shouldn't fail.

src/lib.rs
```diff
@@ -1087,6 +1095,62 @@ impl EGraph {
...
+                *depth -= 1;
+                a = b;
+            }
+            self.pop()?;
```
mwillsey, line 1133: unwrap this

src/lib.rs
```diff
@@ -1087,6 +1095,62 @@ impl EGraph {
...
     }
 
+    fn calc_helper(
+        &mut self,
+        idents: Vec<IdentSort>,
+        exprs: Vec<Expr>,
+        depth: &mut i64,
+    ) -> Result<(), Error> {
+        let mut iter = exprs.into_iter();
+        if let Some(a) = iter.next() {
```
mwillsey, line 1105: Something goofy is happening with this loop structure. Could you refactor into a loop over [`exprs.windows (2)`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.windows)? Also no need to `if let` this whole thing, I think just bailing early if `exprs.len() < 2` would suffice.

src/lib.rs
```diff
@@ -1087,6 +1095,62 @@ impl EGraph {
...
+    }
+
+    // Prove a sequence of equalities universally quantified over idents
+    pub fn calc(&mut self, idents: Vec<IdentSort>, exprs: Vec<Expr>) -> Result<(), Error> {
+        let mut depth = 0;
+        let res = self.calc_helper(idents, exprs, &mut depth);
+        if res.is_err() {
+            // pop egraph back to original state if error
+            for _ in 0..depth {
+                self.pop()?;
```
mwillsey, line 1146: unwrap this, it shouldn't fail.

mwillsey (changes requested):

philzook58:
Another musing:
The calc command is rather close to being a `sketch` command if it was unidirectional and allowed existentials. https://arxiv.org/pdf/2111.13040.pdf

mwillsey (approved):

PR #79, opened 2022-12-11
Logical Formula v0

philzook58:
This pull request is adding what I feel is the essence of datalog based theorem proving. 
The syntax of this is punning off of smtlib a little, despite the theorem prover being essentially constructive and not classical.
This is gated behind something being done to address the issue in https://github.com/mwillsey/egg-smol/pull/78
A blog post discussing the approach is [here](https://www.philipzucker.com/datalog-theorem-harrop/)
The thing that is most compelling is the ability to speak of proving quantified formula by introducing a fresh variable. This is our common practice, but it is nice to be able to write what one means.

There are a couple of additions that can be addressed in future pull requests:
- Query formula may also contain existentials and disjunctions. Disjunction is not hard and is done by compressing queries to DNF. Existentials will probably require fixing the query command.
- A better treatment of scoping and shadowing
- Possibly disjunction and conjunction in Goals, which correspond to database forking. Existentials in Goals is more interesting, requiring in some sense joint queries over multiple databases.
- Skolemization


philzook58:
Another thing to note is that a notion of goal is a natural place at which to apply magic set transformations.

mwillsey:
This PR seems busted at the moment. Can you merge `main` into this branch?

philzook58:
Maybe this will come back. It isn't clear this is on the main track of egg-smol. The branch exists should time come to revisit this

PR #82, opened 2022-12-23
Simplify command

philzook58:
`simplify` uses a simple greedy strategy. If the extraction lowers the cost, it clears the database and restarts it using this simplified term. If not, it continues expanding the database. This is somewhat reminiscent of beam or tabu search. Perhaps it would be better to have the more unopinionated version of simplify that just creates the term, runs the database, and then extracts the term.


mwillsey:
I'm not opposed to adding this, but it'd be nice if there was at least an example where it's "better" than just regular run and extract (1 time). Also, it doesn't really seem useful without the ability to set node or iter limits on the individual runs.

philzook58:
I have encountered a couple instances of the run-extract loop. It was mentioned in the new england systems conference during the coq talk and also in a private correspondence.

I do think I was basically overeager in adding this before the simpler version. Honestly, I just wanted to try it out.

It's actually kind of neat the way it works right now in that it prints the new term every time it discovers a better one, so it looks a bit like a rewriting proof, even though it isn't using typical provenance machinery.

I think I'll replace this with the simpler version and put the alternative strategy on the backlog of things to note. It might be nice to have a pile of different strategies available through flags.

I don't understand really see why node_limit would be that interesting with this technique. Is that available even in the regular run command? The total iter limit is specifiable and it's sort of taking an iterative deepening style thing in the inner loop to get out of valleys.

PR #83, opened 2023-01-01
Lightweight Provenance via Traces

philzook58:
The idea here is if we make `log::trace` basically run on "mutational" (union, makeset, and insert) events with machine readable sexp output that is effectively provenance. (Summarized) Traces of any program that is performing proof search are proofs themselves. This is true of the proofs objects for SAT, SMT, MIP, prolog and datalog. The order/timestamp of the trace statements is some of the most important things when it comes to datalog proofs, which is why just online trace printing makes more sense than it does in other systems. Also, datalog never makes a false move.

Using `log::trace` is also convenient in that "proof production" is off by default 
`RUST_LOG=trace cargo run tests/calc.egg` is sufficient to make it turn on.

Nevertheless, this would still require a lot of chewing by an external program to use and it's probable that the proof information is incomplete until someone actually tries to use it. Extracting a proof object for a particular fact involves traversing the trace structure backwards in time, but with enough hints this is much easier than doing it all in the first place.

In addition, this architecture will not allow deeper introspection of proofs from within egglog itself, either for proof minimization or other reasons. It is however about the simplest thing that could be called proof objects. This will largely be subsumed probably by future additions to the system (like non-destructive inserts, etc), but it could get some proof based things moving for now. In any case more trace statements helps debuggability in my opinion.

The trace event from union basically holds the same data that would go into a proof producing union find.

Some downsides of this is that even if proof recording isn't desired, there is still a bit of extra `Reason` information flowing around that before. An earlier commit had a still useful amount of logging with less reason information. Maybe this is the way to go.

Really, to be useful, the reason information needs to be enhanced. Two obvious things are including the subst `[Value]` dictionary in the reason for a rule and nicer printing besides just raw `value.bits`. But I wanted to see what people thought.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -187,7 +187,7 @@ impl Display for Rule {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(
             f,
-            "{} ==> {}",
+            "({} ==> {})",
             ListDisplay(&self.body, " "),
             ListDisplay(&self.head, " ")
         )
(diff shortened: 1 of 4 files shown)


philzook58:
Given that no one will consume this, it is best to not do it. Oliver's proof production is upcoming

PR #85, opened 2023-01-08
Add basic documentation

alstrup:
List talk and paper, as well as basic syntax.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,17 @@ You can extract the value of specific points of the function using `extract`:
 (extract (Fib 1))
 ```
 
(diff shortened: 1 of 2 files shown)


mwillsey:
Thanks for the PR! Yeah, we don't have much in the way of documentation yet. I will merge this for now, but in the future, I'd prefer to have "checkable" documentation, i.e. using `rustdoc` comments or perhaps an mdBook (@philzook58 was working on something like this). Checkable things are less likely to get stale over time.

But! Given that we literally have nothing anyway, I think this could definitely be useful until more integrated documentation is up and running.

PR #86, opened 2023-01-10
Add f64 sort for floating point numbers

alstrup:
Added lexing for floats and a new sort for these numbers. Implemented the basic primitives.
The goal is to be able to use egg-smol for an transformations on an existing language which has f64.


alstrup:
I'm a bit of a noob Github PRs, so don't know how to split things. I just added a VS Code extension for .egg files as well in this PR.
 
image

mwillsey:
The VSCode extension looks very cool! Thanks for the PR! It should, however, be in a separate PR. To do this, I would recommend creating another branch in your fork that has just the changes for the extension, and then make a PR from that. And then reverting this commit in this branch. Let me know if you need additional help!

alstrup:
What about this? Just merge this, and we can spend time on something that makes more sense. It is not illegal to have a PR with two things in it :-)

mwillsey:
Would you mind separating them? I think we can hold off an a VSCode extension of this kind for now. A lisp or SMT-LIB extension actually works better for egg's syntax (e.g. it has keywords), and you can use it via the `"files.associations"` setting.

alstrup:
Done. I added keywords to the VS code extension as well.

mwillsey:
Thanks for splitting them up!

PR #87, opened 2023-01-12
VS code extension

alstrup:
This time, also with keywords in a separate PR.

image


vscode/eggsmol-1.0.0/extension.js
```diff
@@ -0,0 +1,47 @@
...
+		const document = vscode.window.activeTextEditor.document;
+		document.save().then(() => {
+			var folder;
+			if (vscode.workspace.workspaceFolders != null && vscode.workspace.workspaceFolders.length > 0) {
+				folder = vscode.workspace.workspaceFolders[0].uri.fsPath;
+			} else {
+				folder = ".";
+			}
+			const relativeFile = document.uri.fsPath;
+			eggsmolChannel.appendLine("Running '" + `target/debug/eggsmol ${relativeFile}` + "' in " + folder);
```
mwillsey, line 23: Could you do `cargo run` instead? This might end up invoking a stale binary.

mwillsey (commented):
Do we need to commit the `.vsix` file? That seems like a binary artifact.

alstrup:
> Do we need to commit the `.vsix` file? That seems like a binary artifact.

This is the extension itself that is installed into VS Code, so yes, it is needed.

Issue #42, opened 2022-10-09
Merge during rebuild


ezrosent:
Today merge functions are only applied when we notice a merge while evaluating `Action::Set`, but we can also cause merges during a `rebuild`. Merge functions should also run in response to merges caused by rebuilds. Here's a small failing test case:
```
(datatype N (Node i64))
(function distance (N N) i64 :merge (min old new))

(define a (Node 0))
(define b (Node 1))
(define x (Node 2))
(define y (Node 3))
(set (distance x y) 1)
(set (distance a b) 2)

(union a x)
(union b y)

(run 1)
(check (= (distance x y) 1)) ; fails, the distance has gone up!
```

yihozhang:
@mwillsey @ezrosent what's the latest status on this issue? I'm actually surprised that this has not caused any real problems so far.

mwillsey:
This seems to still fail. Me too, I am surprised. I won't have time to fix for a while though :(

ezrosent:
I can take a stab at this this week.

ezrosent:
Sent a PR with a fix (this took less time than I thought!)

PR #89, opened 2023-01-14
Merge during rebuild

ezrosent:
This PR applies non-union merges discovered during a rebuild, fixing #42.

Couple notes:
* The reason why this fix isn't smaller is that rebuild borrows a `Function` mutably, but running a merge function requires mutable access to the whole `EGraph`, and the egraph owns the function. The usual strategy of moving the function out of the egraph and passing it mutably into `Function::rebuild` won't work for "recursive merges" that depend on the function they are merging.
* Instead, `Function::rebuild` hands back a list of "deferred merges" that get applied after the round of canonicalization finishes.
* This PR also returns an error when AssertEq merges are attempted on unequal values.

Question: is "returning an error" the right approach here, or should these panic immediately?


mwillsey:
I'll merge this, but I think this need to re-borrow is indicative of a bad pattern (we've seen this elsewhere as well). Perhaps we can ban recursive merge functions? I don't really see a use for them. That might enable a use of something like RefCell.

PR #90, opened 2023-01-17
Add knapsack problem and levenshtein distance to tests

rtjoa:
Some classic DP problems that could also be used as benchmarks if we give them big inputs


tests/knapsack.egg
```diff
@@ -0,0 +1,54 @@
...
+(datatype objects
+    (Cons i64 i64 objects)
+    (Nil))
+
+; Given a capacity and a list of objects, finds the maximum value of a
+; collection of objects whose total weight does not exceed the capacity.
+(function Knap (i64 objects) expr)
+
+(rule ((= f (Knap capacity (Cons weight val rest))) (<= weight capacity))
+      ((set (Knap capacity (Cons weight val rest))
```
yihozhang, line 18: maybe just use `f` here since you have bound `(Knap capacity (Cons weight val rest))` to `f`

rtjoa: Resolving from offline conversation - reusing f would work for `union` but not for `set`

PR #96, opened 2023-01-19
Quality of life improvements 

oflatt:
This PR includes
- A desugaring pass for programs so we don't have `rewrite` or `datatype` after desugaring
- Printing programs so we can view them after desugaring or any manipulation
- Rulesets, since I am using them heavily and Ryan wants to use them
- Merge actions thanks to @ezrosent 

We will want a proper module/lambdas/ruleset system but for now this works
This PR supports my current work on debugging egglog and adding proofs


Cargo.toml
```diff
@@ -27,6 +27,7 @@ num-integer = "0.1.45"
 num-rational = "0.4.1"
 num-traits = "0.2.15"
 smallvec = "1.10"
+symbolic_expressions = { git = "https://github.com/oflatt/symbolic-expressions", rev = "4c0ea5ca008f972450b2af72387e64d2c1c6a791" }
```
oflatt, line 30: The symbolic_expressions library has a weird property when printing strings, so I fixed that in my branch

saulshanabrook:
Hey @oflatt, I am starting to update my Python bindings with these changes and am trying to understand the purpose of the rulesets.

If I am reading [the code properly](https://github.com/mwillsey/egg-smol/blob/2df23f9e5207de148f10ad43d05d9c6d4cc0d123/src/lib.rs#L723-L738), adding a ruleset saves all of the current rules under a certain name, and then loading one adds all the rules saved under that name to your rules?

So saving a ruleset and then loading it directly after would be a no-op, is that right? Since you would just be duplicating all of your rules.

So when would you use them? If you have an example somewhere that uses them, feel free to point me to that too if that easier, I couldn't find one.

oflatt:
Hi @saulshanabrook, that's cool!
Disclaimer: Rulesets are probably going to change a lot in the quite near future.
I think you understand it properly. They are very useful when you need to compute something with one ruleset and then use the results with another ruleset (for efficiency or soundness). I think the datalog term for this is stratification.

Ex:
```
(rewrite (Add a b) (Add b a))
(add-ruleset addition-rules)
(clear-rules)
(rewrite (Mul a b) (Mull b c))
(add-ruleset mul-rules)
(clear-rules)
(load-ruleset addition-rules)
(run 1)
(clear-rules)
```

saulshanabrook:
@oflatt Ah cool, thank you for the example! 

One thing I am starting to realize when developing the bindings is that some of the naming/state management features of the `egg-smol` language, like rulesets or definitions, can also be achieved using the state of the host language...

Issue #97, opened 2023-01-20
Another parser bug to test


oflatt:
[Here](https://www.mwillsey.com/egg-smol/?program=XQAAgAD4AwAAAAAAAAAUGQgnfMUD9dO10JtyNyJ209ADDzVx_qN46Z7YaVUOKsVC2Rb3fM7Sb1YfWvEv5QYqp2E64KO4EAlvmPWLdyFGKWve5tp9NMRpXm6jYGP5PQ7-Bw3P_AuR2XS9jK9e1SkKGcv0SkBRhR7Wauarjrxji6U02fmGZfSfe2yQ2K-f8jF8Az2t3eRaFCslJSD5SDraEe4Qyy7GU660sAd5pYU90xjE-PJ3aoGFQ5yUN33S8qMFbWgKxLcdXsNHi51f27aG_0wPu6RNGWjr_3A3DRi6YFTeLgzyX1L3927fWsQYi-Ignv_V0eLJPwclv1DWThAiiQphRp6UzLn7rYcXcyEJb6LE9rloot4eakrAMECeJxXRDCsSlO_AodkRI6-uoy7E2ET8D1bA8rn5DWcXbYBARcwKhN5D2uJYIpHuKdxkjvhVeesOmRD01ocMm4aao38rPWhWvkegGW-Zq-f47QJyRYn0xszxsgLJjV-jLrYQKgP99XTivQ%253D%253D) is another good test case showing a bug in egglog.

I found it implementing a flattener for egglog rules

oflatt:
Fixed by #98

Issue #72, opened 2022-12-07
Panic on admittedly silly axiom


philzook58:
https://www.mwillsey.com/egg-smol/?program=XQAAgACeAAAAAAAAAAAUGQgnfMUD9dO10BC2jbJotgLClFTWhuxd3X1pNTX5W8B1pnvqFbP-Z9e9Nklrmkcj2koM9emIop4mcnfahUjjHhdqJLXm1KNuDIX1T2pDuh93M7OwPZsIlGSZ0J0jVlzUUL5wnBvhQ8MXkradYNkf_9rAEAA%253D

```
(datatype KAT
    (A)
    (par KAT KAT)
)

(rewrite (par p p) p)
(rule ((= r (par q r)) (= q (par q r))) ((union r q)))

; tests
(define q (par A A))
(run 10)
```

```
[INFO ] database size: 10
[INFO ] Made 1 updates
thread 'main' panicked at 'assertion failed: i < self.len()', src/unionfind.rs:65:9
stack backtrace:
```

oflatt:
Fixed by #98

yihozhang:
We can close this one @mwillsey

PR #98, opened 2023-01-23
Fix #94- query compiler doesn't bind variables that are non canonical

### LINKED ISSUE #94: Parser might mishandle differently-named values constrained to be equal
See the below program.
```
(datatype list
    (Empty)
    (Cons i64 list))

(relation eq (list list))

(eq Empty Empty)

; Oddly, this version works:
; (rule ((= x (Cons x1 rest1)) (= y (Cons x2 rest2)) (= 0 (- x1 x2)) (eq rest1 rest2))
(rule ((= x (Cons x1 rest1)) (= y (Cons x2 rest2)) (= x1 x2) (eq rest1 rest2))
      ((eq (Cons x1 rest1) (Cons x2 rest2))))

(define mylist (Cons 1 Empty))

(run 100)
```

Output:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `"i64"`,
 right: `"__bogus__"`', src/typecheck.rs:654:29
```

oflatt:
This PR fixes #94, #72, and #97, adding them to the tests
It does it by replacing the canonical things in the action. The query compiler is in need of refactor, but this works okay for now.
It also adds a racket script for minimizing egglog programs that throw an error


mwillsey:
Indeed we need to refactor. Great catch. Can you move minimize under a new scripts directory?

oflatt:
Done!

oflatt:
I'm confused about the failing build, it doesn't fail locally and the error doesn't make sense

mwillsey:
I bet it’s a rust version thing. At some point I think the rust-toolchain
file disappeared for some reason.

oflatt:
This one?
https://github.com/mwillsey/egg-smol/blob/564cfcfaec26b5bcd868da446623b837902d305d/rust-toolchain

mwillsey:
Yeah, I’m guessing that CI uses an older rust if you don’t tell it
otherwise.

oflatt:
Err, actually I still have that file. Should I bump it?
It's also fishy that I didn't change the rust code but it failed the second time. I'll try rerunning

mwillsey:
I just committed a new version of that file pinning us to 1.66.1, but that
didn't fix the issue. I fixed it just by changing the code.

oflatt:
Okie dokie
Should we merge this then?

PR #109, opened 2023-02-08
Add normalization pass when desugaring egglog programs

oflatt:
This PR is a large overhaul of our internal representation of egglog programs, desugaring many commands and creating a much simpler normal form for egglog.
Besides turning many things into SSA, we also turn queries into a special normalized form. Queries now consist of tuples binding unique names (which appear only once) and a set of equality or primitive constraints over those variables.

Since datatypes are now desugared into functions, and definitions are desugared into normal action `let` statements, the namespaces for functions and sorts are now the same.

For convenience, this PR adds slightly better ruleset functionality.


src/lib.rs
```diff
@@ -402,6 +426,7 @@ impl EGraph {
     }
 
     pub fn declare_function(&mut self, decl: &FunctionDecl) -> Result<(), Error> {
+        self.declare_sort(decl.name, None)?;
```
oflatt, line 429: The function and sort namespace are now the same

PR #114, opened 2023-02-22
Fix quoted strings

yihozhang:
A literal string like `"abc"` was parsed as a string with five characters, including the first and the last quote character. This PR tries to fix it.


src/desugar.rs
```diff
@@ -422,8 +422,8 @@ pub(crate) fn desugar_command(
             desugared.push(NormCommand::Fail(Box::new(last.unwrap())));
             desugared
         }
-        Command::Input { .. } => {
-            todo!("desugar input");
```
yihozhang, line 426: @oflatt Is there a potential pitfall here that prevents you from just building a `NormCommand::Input`?

yihozhang:
The tests are fixed

src/desugar.rs
oflatt: I think I just didn't understand what the input command did lol, thanks for the fix

PR #115, opened 2023-02-24
Add first-class proofs to egglog

oflatt:
This (large) PR adds:
- Proofs in egglog via only syntactic transformations on the input program, similar to the `proofs.egg` example
- A new typechecker to provide type information for the proofs pass
- Some tweaks to egglogs semantics to facilitate proofs (more on these below)


src/ast/parse.lalrpop
```diff
@@ -53,18 +58,17 @@ Command: Command = {
         <ruleset:(":ruleset" <Ident>)?>
     ")" => Command::BiRewrite(ruleset.unwrap_or("".into()), Rewrite { lhs, rhs, conditions: conditions.unwrap_or_default() }),
     "(" "define" <name:Ident> <expr:Expr> <cost:Cost> ")" => Command::Define { name, expr, cost },
+    "(" "let" <name:Ident> <expr:Expr> ")" => Command::Action(Action::Let(name, expr)),
```
oflatt, line 61: Define now desugars into let.

src/lib.rs
```diff
@@ -912,23 +859,24 @@ impl EGraph {
...
                     self.run_schedule(&sched);
                     format!("Ran schedule {}.", sched)
                 } else {
                     "Skipping schedule.".to_string()
                 }
             }
-            NormCommand::Calc(idents, exprs) => {
-                self.calc(idents.clone(), exprs.clone())?;
+            NCommand::Calc(idents, exprs) => {
+                todo!("Fix calc again");
```
oflatt, line 871: TODO fix before merge

src/ast/parse.lalrpop
```diff
@@ -53,18 +58,17 @@ Command: Command = {
...
+    "(" "simplify"  <limit:UNum> <expr:Expr> <until:(":until" <(Fact)*>)?> ")" 
         => Command::Simplify { expr, config : RunConfig { ruleset: "".into(), limit, until } },
     "(" "add-ruleset" <name:Ident> ")" => Command::AddRuleset(name),
-    "(" "run-schedule" <Schedule*> ")" => Command::RunSchedule(Schedule::Sequence(<>)),
     "(" "calc" "(" <idents:IdentSort*> ")" <exprs:Expr+> ")" => Command::Calc(idents, exprs),
     "(" "extract" <variants:(":variants" <UNum>)?> <e:Expr> ")" => Command::Extract { e, variants: variants.unwrap_or(0) },
-    "(" "check" <Fact> ")" => Command::Check(<>),
-    "(" "clear" ")" => Command::Clear,
-    "(" "query" <List<Fact>> ")" => Command::Query(<>),
+    "(" "check" <(Fact)*> ")" => Command::Check(<>),
```
oflatt, line 70: Deleted query, and check now performs a query just like `rule`

src/ast/parse.lalrpop
```diff
@@ -145,8 +149,9 @@ UNum: usize = {
     <Num> => <>.try_into().unwrap(),
 }
 F64: OrderedFloat<f64> = <s:r"(-)?[0-9]+\.[0-9]+(e(-)?[0-9]+)?"> => OrderedFloat::<f64>(s.parse().unwrap());
-Ident: Symbol = <s:r"([[:alpha:]][\w-]*)|([-+*/!=<>&|^/%]+)"> => s.parse().unwrap();
+Ident: Symbol = <s:r"(([[:alpha:]][\w-]*)|([-+*/!=<>&|^/%_]))+"> => s.parse().unwrap();
```
oflatt, line 152: Have as many underscores as you want

src/main.rs
```diff
@@ -26,13 +26,15 @@ fn main() {
...
-                Ok(_msgs) => {}
-                Err(err) => {
-                    log::error!("{}", err);
-                }
+        let mut program = String::new();
+        stdin
+            .lock()
+            .read_to_string(&mut program)
+            .unwrap_or_else(|_| panic!("Failed to read program from stdin"));
+        match egraph.parse_and_run_program(&program, true, false) {
```
oflatt, line 34: Big change- for convenience, my proofs code assumes it has the entire program. I think this can be changed fixed, but should it? The interactive mode has been replaced by the stdin accepting an entire program in parenthesis.

src/proofs.rs
```diff
@@ -0,0 +1,983 @@
+use crate::*;
+
+use crate::desugar::{desugar_commands, literal_name, Desugar};
+use crate::typechecking::FuncType;
+use symbolic_expressions::Sexp;
+
+pub const RULE_PROOF_KEYWORD: &str = "rule-proof";
```
oflatt, line 7: This keyword allows you to get the proof for a rule firing in the actions of a rule.

src/sort/map.rs
```diff
@@ -333,6 +388,14 @@ impl PrimitiveLike for Remove {
         map.remove(&values[1]);
         map.store(&self.map)
     }
+
+    fn get_type(&self) -> (Vec<ArcSort>, ArcSort) {
```
oflatt, line 392: TODO remove this again, I don't need it

src/desugar.rs
```diff
@@ -309,64 +274,112 @@ fn flatten_actions(actions: &Vec<Action>, get_fresh: &mut Fresh) -> Vec<NormActi
...
     desugar: &mut Desugar,
+    get_all_proofs: bool,
 ) -> Result<Vec<NormCommand>, Error> {
-    Ok(match command {
+    let res = match command {
         Command::Function(fdecl) => {
-            vec![NormCommand::Function(fdecl)]
+            vec![NCommand::Function(fdecl)]
         }
         Command::Datatype { name, variants } => desugar_datatype(name, variants),
```
oflatt, line 351: Another big change is the `declare` keyword, which declare an unbound variable.
By default, if you define a function with no arguments in `datatype`, it desugars into a `declare`.
This fixes the ambiguity between functions with no arguments and global declarations/definitions.

Previously, you could refer to functions with no arguments without parenthesis (which is evil)

src/lib.rs
oflatt: fixed

src/sort/map.rs
oflatt: removed

tests/until.egg
```diff
@@ -1,25 +1,25 @@
 ; A simple group
 (datatype G (I) (A) (B))
-(function * (G G) G)
+(function g* (G G) G)
```
oflatt, line 3: Functions and primitives can't conflict in name anymore for now

src/main.rs
oflatt: I fixed the REPL!

src/desugar.rs
```diff
@@ -1,206 +1,114 @@
-use crate::*;
+use crate::{proofs::RULE_PROOF_KEYWORD, *};
 
-pub(crate) type Fresh = dyn FnMut() -> Symbol;
+// TODO fix getting fresh names using modules
+const PROOF_UNDERSCORES: &str = "___";
```
oflatt, line 4: When we have modules, this will be fixed hopefully. Right now don't use three underscores in your egglog

src/proofheader.egg
```diff
@@ -0,0 +1,103 @@
...
+      ((set (ProofBetween__ t1 t2 t1) Null__)
+       (set (ProofCost__ t1 t1) 0))
+      :ruleset proof-extract__)
+;; demand for proof cost
+(rule ((= proof (ProofBetween__ t1 t2 tmid))
+       (= eproof (EqGraph__ tmid next)))
+      ((set (ProofCost__ t1 next)
+            100000000)))
+;; do one step of proof search to find a path
+(rule ((= proof (ProofBetween__ t1 t2 tmid))
```
oflatt, line 86: The proof extraction rules do seem to be a little slow for dense graphs at the moment (repro-unsound takes 50 seconds on my machine)

src/typechecking.rs
```diff
@@ -0,0 +1,577 @@
...
+
+#[derive(Default, Clone)]
+pub struct TypeInfo {
+    // get the sort from the sorts name()
+    pub presorts: HashMap<Symbol, PreSort>,
+    pub sorts: HashMap<Symbol, Arc<dyn Sort>>,
+    pub primitives: HashMap<Symbol, Vec<Primitive>>,
+    pub func_types: HashMap<Symbol, FuncType>,
+    pub global_types: HashMap<Symbol, ArcSort>,
+    pub local_types: HashMap<CommandId, HashMap<Symbol, ArcSort>>,
```
oflatt, line 23: Type info for specific rules- tells what the type of locally-bound variables are.
This new typechecker is redundant with the old one, we should remove the old one eventually

src/typechecking.rs
```diff
@@ -0,0 +1,577 @@
...
+                } else {
+                    return Err(TypeError::Unbound(*var1));
+                }
+            }
+        }
+        Ok(())
+    }
+
+    pub fn reserved_type(&self, sym: Symbol) -> Option<ArcSort> {
+        if sym == RULE_PROOF_KEYWORD.into() {
```
oflatt, line 432: The only reserved type right now besides primatives is the rule-proof keyword

saulshanabrook:
I am starting to upgrade egg-smol Python to include these recent commits.

I was curious if you were thinking that eventually this proofs flag will be removed, and it will either always be enabled or disabled? What is the best way to think about the differences in behavior between these flags?

Also now with command normalization, I am trying to decide if the Python API should expose the unnormalized or the normalized types... How do you think about the differences between them?

oflatt:
I think the plan is to keep the flag. With the flag on, your program is instrumented by a macro to add proofs to your code. So it makes some new relations available, as well as some rules for extracting proofs. There's an example in `before-proofs.egg`.

Also keep in mind proofs don't work super well yet, they are very much in flux

PR #117, opened 2023-03-01
Fix semantics of nullary functions and declare

oflatt:
This PR answers these questions:

1. When a nullary function is defined (as part of a datatype or something else), do we put something in the database?
2. What is the syntax for "calling" nullary functions in rules?

My opinions:
1. We should not put it in the database, it would behave weirdly compared to the other functions
2. Having optional parenthesis is very ambiguous, let’s not do that. If you want to call it without parens, make a global variable

For 2, I added syntactic sugar called `declare` which instantiates your new type and gives it a name.


src/ast/parse.lalrpop
```diff
@@ -43,10 +43,10 @@ Command: Command = {
         <merge:(":merge" <Expr>)?> <default:(":default" <Expr>)?> ")" => {
         Command::Function(FunctionDecl { name, schema, merge, merge_action: merge_action.unwrap_or_default(), default, cost })
     },
-    "(" "declare" <name:Ident> <var_type:Ident> <cost:Cost> ")" => Command::Declare(name, var_type, cost),
+    "(" "declare" <name:Ident> <sort:Ident> ")" => Command::Declare{name, sort},
     "(" "relation" <name:Ident> <types:List<Type>> ")" => Command::Function(FunctionDecl::relation(name, types)),
     "(" "ruleset" <name:Ident> ")" => Command::AddRuleset(name),
-    "(" "rule" <body:List<Fact>> <head:List<Action>> <ruleset:(":ruleset" <Ident>)?> <name:(":name" <Ident>)?> ")" => Command::Rule{ruleset: ruleset.unwrap_or("".into()), name: name.unwrap_or("".into()), rule: Rule { head, body }},
+    "(" "rule" <body:List<Fact>> <head:List<Action>> <ruleset:(":ruleset" <Ident>)?> <name:(":name" <String>)?> ")" => Command::Rule{ruleset: ruleset.unwrap_or("".into()), name: name.unwrap_or("".to_string()).into(), rule: Rule { head, body }},
```
oflatt, line 49: Also fix a parsing bug with rule names

oflatt:
I think this is good to go now, looked over the diff carefully

PR #122, opened 2023-03-10
Fix typecheck bug

oflatt:
Previously, unbound variables are caught by the SSA validity pass instead of the typechecker.
This is a simple fix


mwillsey:
I don’t quite understand expect_lookup, can you expand a bit?

oflatt:
So `expect_lookup` means that we expect the variables in the expression to already be bound. In actions, we expect variables to be already bound so we can "look up" their type. In facts, we have to infer the type based on the functions so we don't expect to be able to look them up.

Issue #124, opened 2023-03-26
Check passes for wrong assertion 


remysucre:
The following check should fail, but it doesn't: 

```
(function f (i64) i64 :merge (min old new))

(set (f 1) 4)

(check (= (f 1) 2))
```

yihozhang:
@clyben

yihozhang:
We can close this one! @mwillsey

PR #127, opened 2023-04-06
Add support for constaining by constants in GJ

ezrosent:
This is a draft supporting constant constraints in GJ. I think we'd want to clean this up a bit before submission, but it could serve as a useful proof of concept. With this change, the program in #124 fails (as expected).

The basic idea here is to add a degenerate variant of the `Intersect` instruction that just does a single lookup on a trie at a particular constant value. We generate one of these instructions per constant constraint in the query, and all of these constant instructions are run before any `Intersect`s are.

The alternative would be to munge these constraints into the existing primitive constraints, but this seems more straightforward and efficient.

### DIFF
--- a/src/gj.rs
+++ b/src/gj.rs
@@ -18,6 +18,11 @@ enum Instr<'a> {
         value_idx: usize,
         trie_accesses: Vec<(usize, TrieAccess<'a>)>,
     },
+    ConstrainConstant {
+        index: usize,
+        val: Value,
+        trie_access: TrieAccess<'a>,
+    },
     Call {
         prim: Primitive,
         args: Vec<AtomTerm>,
@@ -42,6 +47,13 @@ impl<'a> std::fmt::Display for Program<'a> {
                         write!(f, "  {}: {}", trie_idx, a)?;
                     }
                     writeln!(f)?
+                }
+                Instr::ConstrainConstant {
+                    index,
+                    val,
+                    trie_access,
+                } => {
+                    writeln!(f, " ConstrainConstant {index} {trie_access} = {val:?}")?;
                 }
                 Instr::Call { prim, args, check } => {
                     writeln!(f, " Call {:?} {:?} {:?}", prim, args, check)?;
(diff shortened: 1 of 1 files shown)


mwillsey:
I think it would also make sense to just special-case evaluate them up front. Even in cases where the result it "used", it kinda makes to eval the query as much as possible up front. This way we avoid having to ensure these run at the "top" of the eval order.

yihozhang:
This is my first dive into the query executor so I might be asking silly questions:

* Is it redundant that every `Intersect` will check the equality constraints specified by the `TrieAccess`? For example, if the atom is `R(1, x, y)`,` `R.0 = 1` will be checked when both intersect on `x` and on `y`, it this correct?
* In this PR, for atom like `R(1, 2)`, two `ConstrainConstant` will be generated. I think we only need to generate one so we only do one scan of the relation? Related, because `TrieAccess` is already doing the filtering, `ConstrainConstant` probably does not need to carry (and check) a `val` field?

Also re @mwillsey : I think it has value to have a self-contained IR for low-level execution. E.g., the user may want to specify their own execution plans when the existing ones are not sufficient (Souffle does this). Otherwise the current IR cannot even express the query `(= (f 1) 2)`.

ezrosent:
Max, feel free to weigh in if I have stuff wrong! I think we are "over-checking" constraints, but it's not obvious to me where we should be removing things.

> Is it redundant that every Intersect will check the equality constraints specified by the TrieAccess? For example, if the atom is R(1, x, y), R.0 = 1 will be checked when both intersect on x and on y, it this correct?

It is _often_ redundant.. I think it's there to handle cases like `R(x, x)`. We only make one trie per atom, and so we can't use `Intersect` to only yield elements of `R` with the first and second columns equal. For `R(1, x, y)` I think `ConstrainConstant` is redundant but even then it could speed the query up!

> In this PR, for atom like R(1, 2), two ConstrainConstant will be generated. I think we only need to generate one so we only do one scan of the relation?

That would work, but we'd need to modify ConstrainConstant to do the scan. The advantage of doing `ConstrainConstant` twice is that we don't do any scans at all: we only call `get` and never call `for_each` until the end. Doing two `ConstrainConstant`s lets us filter for the 1s and the 2s before doing the scan.

> Related, because TrieAccess is already doing the filtering, ConstrainConstant probably does not need to carry (and check) a val field?

Right... I think we could scan `constraints` in the access and look for a constant equal to `col` and use that in the call to `get`: this just avoids the scan of the constraints vector. It doesn't add information.

mwillsey:
@yihozhang yes I think it would make sense to have manual query plans at some point, but that's a bigger topic. Something to think about for the backend rewrite.

@ezrosent I'm going to defer to you for this PR, I unfortunately won't have time to actually review it this week. But if you say it's good to go then I'll merge. All badness will be washed away in the Great Backend Rewrite(tm) anyway :)

PR #129, opened 2023-04-13
Proper container support

yihozhang:
It is relatively simple to add proper support for containers like `set`, `map`, and `vec`. There are essentially two changes:

1. Let `canonicalize` be a sort-defined operation.
2. For each container of EqSort, instead of tracking changed containers, we track changed ids *in* the container.

Now, a sort can be one of the following three kinds
* plain sort,
* EqSort, and
* ContainerSort, of which a container of EqSort is a special case.

Current limitation: we do not allow nest EqSort containers within other containers. This will make tracking changed ids more difficult (essentially you need congruence closure). We can instead support `multi_map` if we want.


yihozhang:
As a side note, I think we can use functional data structures for containers...

Issue #123, opened 2023-03-26
Panic when extracting from a partial function 


remysucre:
The following program panics: 

```
(function f (i64) i64 :merge (min old new))

(set (f 1) 4)

(extract (f 2))
```

mwillsey:
What should the intended behavior here be? `(f 2)` is not defined.

remysucre:
Is panicking currently interpreted as error? If so then it's fine.

yihozhang:
@clyben It would be great if you can take a look!

mwillsey:
I think this is fixed now.

PR #132, opened 2023-04-18
Fix panic

clyben:
A simple fix to issue [#123](https://github.com/mwillsey/egg-smol/issues/123) by replacing the panic with just returning an Error. Also, adding a fail_check test as the one in fixed issue [#124](https://github.com/mwillsey/egg-smol/issues/124), also adding some fail_check in cyk.egg.


yihozhang:
This looks great! Could you run `cargo fmt` to fix the style issue? (You may need to install it first with `rustup component add rustfmt`)

Issue #93, opened 2023-01-18
Semi-naive should also inspect the deltas of the right-hand side


yihozhang:
The semi-naive will only evaluate based on the deltas of the left-hand side, but the deltas of the right-hand side may also impact the output. See the code below
```
(function f (i64) i64 :merge (max old new))

(set (f 0) 0)
(set (f 3) 0)


(rule ((= f0 (f 0))) ((set (f 1) f0)))
(rule ((= f1 (f 1))) ((set (f 2) f1)))

;; update f3 some iters later to make sure f(0) is inactive
(rule ((= f2 (f 2))) ((set (f 3) 3)))

;; This rule should fire and set f(0) to be 3, but because f0 is inactive, it does not fire (despite that f3 is active now)
(rule ((= f0 (f 0))) ((set (f 0) (f 3))))

(run 100)
(print f) ;; f0 is expected to have value 3, but has 0 in reality.
```

yihozhang:
We can close this one as well :)

PR #134, opened 2023-04-20
Semi naive fix

clyben:
Fixing the problem in issue[#93](https://github.com/mwillsey/egg-smol/issues/93), that in semi-naive eval, the change happens in the right-hand-side in set operation should affect the evaluation, but being ignored due to semi-naive. The fix was to add a new rule while meeting setting to a Call, add one more equal relation to the body part of the rule in the desugar_command phase, as well as let binding since they might bind variables in the Call.

### DIFF
--- a/src/ast/desugar.rs
+++ b/src/ast/desugar.rs
@@ -368,7 +368,12 @@ pub(crate) fn desugar_command(
         Command::Include(file) => {
             let s = std::fs::read_to_string(&file)
                 .unwrap_or_else(|_| panic!("Failed to read file {file}"));
-            return desugar_commands(desugar.parse_program(&s)?, desugar, get_all_proofs, seminaive);
+            return desugar_commands(
+                desugar.parse_program(&s)?,
+                desugar,
+                get_all_proofs,
+                seminaive,
+            );
         }
         Command::Rule {
             ruleset,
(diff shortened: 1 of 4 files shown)


src/ast/desugar.rs
```diff
@@ -375,11 +383,77 @@ pub(crate) fn desugar_command(
...
+                let mut new_rule = rule.clone();
+                // only add new rule when there is Call in body to avoid adding same rule.
+                let mut add_new_rule = false;
+
+                for head_slice in new_rule.head.iter_mut() {
+                    match head_slice {
+                        Action::Set(_, _, value) => {
+                            // if the right hand side is a function call,
+                            // move it to body so seminaive fires
+                            match value {
```
yihozhang, line 397: This can probably be simplified to `if matches!(value, Expr::Call(_, _)) {`. Also, I think `value` here does not necessarily have to be a `Call`? For example,

```
(rule (...)
    ((let x (f 1))
     (set y x)))
```
In the future, we can potentially do some reachability/type analysis to find out what needs to be moved to the body and what needn't. For now we can keep things simple (probably good to leave a comment here mentioning potential optimizations for future reference).

src/ast/desugar.rs
```diff
@@ -375,11 +383,77 @@ pub(crate) fn desugar_command(
...
+                            eq_vec.push(Expr::Var(symbol.clone()));
+                            eq_vec.push(expr.clone());
+                            new_rule.body.push(Fact::Eq(eq_vec));
+                        }
+                        _ => (),
+                    }
+                }
+
+                // remove all let action
+                new_rule.head.retain_mut(|action| match action {
```
yihozhang, line 423: `|action| matches!(action, Action::Let(_))` might be more concise. Also maybe considering put the retaining inside the `if add_new_rule { ... }` block?

src/ast/desugar.rs
```diff
@@ -375,11 +383,77 @@ pub(crate) fn desugar_command(
...
+                    Action::Let(_, _) => false,
+                    _ => true,
+                });
+
+                if add_new_rule {
+                    log::debug!("Added semi-naive desugared rules:\n{}", new_rule);
```
yihozhang, line 429: `semi-naive desugared rules` -> `a semi-naive desugared rule`

src/ast/desugar.rs
```diff
@@ -375,11 +383,77 @@ pub(crate) fn desugar_command(
...
+                    ]
+                } else {
+                    vec![NCommand::NormRule {
+                        ruleset,
+                        name,
+                        rule: flatten_rule(rule, desugar),
+                    }]
+                }
+            } else {
+                vec![NCommand::NormRule {
```
yihozhang, line 451: There are some redundancies here on the return value. I recommend we can change the program structure to be something like:
```
let mut result = vec![NCommand::NormRule { ... rule }] // original rule
if seminaive {
    ...
    if add_new_rule {
        result.push(NCommand::NormRule { ... new_rule });
    }
}
result // return here
```

yihozhang:
Thanks for the PR! This looks great. I left some comments. Also see the [suggestions](https://github.com/mwillsey/egg-smol/actions/runs/4759766423/jobs/8459367829?pr=134#step:4:283) from clippy.

src/ast/desugar.rs
```diff
@@ -375,11 +383,58 @@ pub(crate) fn desugar_command(
...
-            vec![NCommand::NormRule {
+            let mut result = vec![NCommand::NormRule {
                 ruleset,
                 name,
-                rule: flatten_rule(rule, desugar),
-            }]
+                rule: flatten_rule(rule.clone(), desugar),
+            }];
+
+            if seminaive {
```
oflatt, line 392: Please migrate to helper function

oflatt (approved):
Great work!

yihozhang:
It feels to me that #93 is a bug of semi-naive evaluation before anything semantical, since one would expect semi-naive and naive to produce the same result. So this PR is just to make sure semi-naive is sound with some program transformations.

(On lifting terms out of actions, I don't have much of an opinion, as long as we can keep the query executor pure (and leave the side effects to the eval_actions))

mwillsey:
To follow up from today's meeting: this PR definitely feels like a step in the right direction, but I think we need to understand a little better what's going on. As @oflatt noted, this transformation feels awfully similar to the mythical (and somewhat misnamed) "demand transformation" that we have previously discussed.

In particular, why do we not do this when the lhs of a `set` has a nested call? And what about unions? I don't think we _need_ it for unions, but I don't yet have a crisp understanding why.

src/ast/desugar.rs
clyben: I had a helper function

mwillsey:
@oflatt is this change going to break proofs? I recall we discussed something about that in the lab.

PR #135, opened 2023-04-25
Add an ordering on proof edges to facilitate checking/viewing

oflatt:
It's easier to visualize and check proofs if there is an ordering on edges, so this PR adds this ordering. It abuses a table as a counter and bumps it after every rule fires.


mwillsey:
From our convo: let's sidestep the semi-naive issue by just having a "magic variable" to get the current iteration

oflatt:
Okay this is ready for review. What it now does is 
1. Adds iterations to the proofs just using syntax transformations
2. Adds a `SetNoTrack` primitive to set without doing magic set transformation
3. Makes the tester more robust by testing that we can re-parse desugared egglog

oflatt:
You may notice `lambda_with_proofs` is suspiciously slow... that's just the proof extraction, it will get fixed soontm

PR #136, opened 2023-04-26
Expose results after running commands

saulshanabrook:
This PR adds an API for retrieving the results of the few commands that produce values: extracting expressions and triggering runs.

This works by saving the last values of these when they are created on the e-graph and allowing users to take them. While I assume this does add some overhead for existing egg-smol users, I don't imagine it should be very substantial, since I am only ever saying a maximum of one result at a time.

## Background

I am using this in the Python bindings to retrieve these results after a use executes a command. See  https://github.com/metadsl/egg-smol-python/pull/14 for where I am currently using this branch.

I made changes in the Python bindings to only interface at the level of command execution, instead of exposing individual e-graph methods, as I was doing previously. I did this in response to the addition of normalization and the recommendation to only expose the un-normalized commands to users https://github.com/mwillsey/egg-smol/pull/115#issuecomment-1488979996.

This actually works out a bit cleaner in the low level Python interface, since I don't have to decide how to translate between methods and commands.

However, I needed a way then to retrieve the values that were created in certain cases, like for extracting expressions. Instead of special casing these as methods calls, I store the most recent results and add a method to take them.

Originally, I did try to keep my old method of calling the actual methods on the e-graph. I found this to be messier though, since I needed to manage all the de-sugaring myself, which in turn required access to more private structures in the e-graph.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -56,11 +56,20 @@ pub trait PrimitiveLike {
 
 #[derive(Debug, Clone, Default)]
 pub struct RunReport {
-    updated: bool,
-    search_time: Duration,
-    apply_time: Duration,
-    rebuild_time: Duration,
+    pub updated: bool,
+    pub search_time: Duration,
+    pub apply_time: Duration,
+    pub rebuild_time: Duration,
 }
+
+
+#[derive(Debug, Clone)]
+pub struct ExtractReport {
+    pub cost: usize,
+    pub expr: Expr,
+    pub variants: Vec<Expr>,
+}
+
 
 impl RunReport {
     pub fn union(&self, other: &Self) -> Self {
@@ -73,7 +82,7 @@ impl RunReport {
     }
 }
 
-const HIGH_COST: usize = usize::MAX;
+pub const HIGH_COST: usize = usize::MAX;
 
 #[derive(Clone)]
 pub struct Primitive(Arc<dyn PrimitiveLike>);
@@ -156,6 +165,8 @@ pub struct EGraph {
     pub node_limit: usize,
     pub fact_directory: Option<PathBuf>,
     pub seminaive: bool,
+    extract_report: Option<ExtractReport>,
+    run_report: Option<RunReport>
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1128,6 +1148,16 @@ impl EGraph {
...
         self.proof_state.type_info.sorts.get(&value.tag)
     }
+
+    // Takes the last extract report and returns it, if the last command saved it.
+    pub fn take_extract_report(&mut self) -> Option<ExtractReport> {
+        self.extract_report.take()
+    }
+
+    // Takes the last run report and returns it, if the last command saved it.
+    pub fn takes_run_report(&mut self) -> Option<RunReport> {
```
mwillsey, line 1158: Typo in function name

src/lib.rs
```diff
@@ -1128,6 +1148,16 @@ impl EGraph {
     pub(crate) fn get_sort(&self, value: &Value) -> Option<&ArcSort> {
         self.proof_state.type_info.sorts.get(&value.tag)
     }
+
+    // Takes the last extract report and returns it, if the last command saved it.
+    pub fn take_extract_report(&mut self) -> Option<ExtractReport> {
```
mwillsey, line 1153: Why should these methods take the run report instead of returning a reference?

mwillsey (changes requested):
Seems good to me; a pretty limited change. See my comments.

src/lib.rs
saulshanabrook: It just seemed simpler and more efficient, but I can switch it to return a reference if you prefer.

src/lib.rs
saulshanabrook: Thanks, fixed!

src/lib.rs
mwillsey: Yes, I think a reference would be better. It avoids mutation.

saulshanabrook: Sounds good, fixed

mwillsey (approved):

saulshanabrook:
Thanks for merging this!

PR #140, opened 2023-05-02
Fix high cost overflowing

saulshanabrook:
This PR reduces the size of the HIGH_COST so that it won't overflow when added to other costs.

Instead of using all the bits, it will only use half of them, which should still give it a pretty high cost but also plenty of room to grow.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -80,7 +80,9 @@ impl RunReport {
     }
 }
 
-pub const HIGH_COST: usize = usize::MAX;
+// Set a high cost to avoid being extracted
+// Use half of our bits to avoid overflow when adding costs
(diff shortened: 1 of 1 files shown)


yihozhang:
Would it be nice to make cost have type `Option<usize>` to avoid magic number costs for unwanted constructors for extraction? I'm happy to make a PR.

saulshanabrook:
@yihozhang That makes sense to me. Would these unwanted constructors then never be extracted?

yihozhang:
Yes that's what I'm thinking

saulshanabrook:
@yihozhang That sounds better to me than this PR, so that these temporary variables are never returned.

mwillsey:
Ok, closing this for now in favor of that future PR.

PR #141, opened 2023-05-03
Clarifying several `define` related keywords.

yihozhang:
This PR has the following two changes related to costs:

* It is now impossible to extract variables introduced by `define`.
* It is possible to give `declare` a cost.

I believe this is the desired behavior in most cases. Moreover,

* Every nullary function can be referenced without parens. (We kinda already support it. This PR fixes some issues and cleans up the codebase)
* There is a new `:noextract` keyword to explicitly assert that we don't want to extract certain functions.

To summarize, there are currently several ways to introduce a function:
* `function`: this is the most general form.
* `datatype`: this desugars to `function`.
* `relation`: this desugars to `function` to `unit`
* `declare`: this desugars to nullary functions / uninterpreted constants.
* `let`/`define`: these two are equivalent and there are some code duplications surrounding them right now. Both introduce a new `noextract` nullary function that gets unioned with a given expression.

This PR currently breaks `proof`.


src/ast/desugar.rs
```diff
@@ -689,21 +685,17 @@ impl Desugar {
...
+            name,
+            schema: Schema {
+                input: vec![],
+                output: sort,
+            },
+            default: None,
+            merge: None,
+            merge_action: vec![],
+            cost,
+        })]
```
saulshanabrook, line 699: 👍 this seems cleaner than the previous version of creating a randomly named function.

saulshanabrook (approved):
This makes sense to me!

In Python, I was thinking I would expose this as the `cost` argument having a default value of `1`, and if the user passes in `None` translate to `:noextract`.

You could do something similar here, allowing some other special keyword for cost that isn't an integer, instead of adding another parameter (`:noextract`), so that it's clear that you can't supply both.

oflatt:
As far as I can tell this is resolved by https://github.com/egraphs-good/egglog/pull/158
Let bindings do not create function tables any more, and define is removed in favor of let

PR #145, opened 2023-05-19
Fix bug when extracting sets/maps with equality sort values

saulshanabrook:
This PR resolves a bug that previously triggered an error when trying to extract a set or map with a user-defined equality sort as its value. The issue stemmed from the fact that equality sorts were not properly handled in certain circumstances.

To fix this, I passed the egraph into the make_expr function on each sort. This allows sorts that may contain equality sorts to call the extract method and load those values when necessary.

I also add added a demonstration of this feature to provide a working example and test case to the `tests/fusion.egg` file.

During this process, I refactored some parts of the code to use the `extract` method. This ensures that all code determining whether a value is an equality sort uses the same logic.

These changes ensure that all code determining whether a value is an equality sort uses the same logic, which leads to more consistent behavior.

### DIFF
--- a/src/extract.rs
+++ b/src/extract.rs
@@ -2,7 +2,7 @@ use hashbrown::hash_map::Entry;
 
 use crate::ast::Symbol;
 use crate::util::HashMap;
-use crate::{EGraph, Expr, Function, Id, Value, ArcSort};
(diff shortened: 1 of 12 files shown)


saulshanabrook:
This is now ready for review. Let me know if you have any questions or comments about it.

Issue #143, opened 2023-05-08
panic: ConstrainEq on unbound variables


yihozhang:
```
(rule ((= x 1) (= y x) (= z y)) ())
(run 1)
```
panics, while
```
(rule ((= x 1) (= y x)) ())
(run 1)
```
does not.

Seems related to desugaring @oflatt?

This issue is blocking @clyben from fixing a follow-up issue of #93.

More details:

Currently, the program produces body
```
[AssignLit(x, 1), ConstrainEq(y, x), ConstrainEq(z, y)]
```
while the expected body is
```
[AssignLit(x, 1), AssignVar(y, x), AssignVar(z, y)]
```

(Note that we don't have AssignVar currently.)

`ConstraintEq`s are produced by `flatten_facts` and `flatten_equalities` on cases like `(= var1 var2)`.

We can also remove the checks since we are converting `NormFact::ConstrainEq` back to `Fact::Eq` anyway so it shouldn't matter, maybe?

yihozhang:
Also just curious, why do we need to distinguish between `NormFact::Assign` and `NormFact::ConstrainEq`? Is it for proof generation? After all, the query compiler does not distinguish between them.

oflatt:
Yeah I think the problem is that we disallow "unbound" variables in contrainEq. Variables need to be bound in an assignment
I'd rather keep this invariant- is there a way to work around it?

yihozhang:
Fixed by #176.

Issue #142, opened 2023-05-04
Follow-up issues on "Semi-naive should also inspect the deltas of the right-hand side"


yihozhang:
#93 is fixed by #134. However, there are several issues left with #134:

1. `let`-bound variables are conservatively handled. We will translate every let-bound right-hand side to the left-hand side even when they are not used in `set`. To handle this correctly, we may need a liveness analysis.

2. Namespaces of let-bound variables on the right-hand side are not correctly handled.
```
(rule ((= a (expr1)))
      ((let a (expr2))
       (set (...) a)) 
```

yihozhang:
The first issue becomes obsolete with #461 .

The second issue is also fixed, as the following program throws an error:

```

(rule ((= a 1))
      ((let a 2)
       (extract a)))
(run 1)
;; In 3:8-16 of web-demo.egg: (let a 2)
;; Variable a was already defined
```

PR #146, opened 2023-05-19
Draft: more fix for seminaive

clyben:
This PR fixes some remaining issues of #93 (which was described in #142) that the right-hand side of the set is not a function but bound to a function in the previous let binding. eg:  `(rule ((= f0 (f 0))) ((let a (f 3)) (set (f 0) a)))`
The previous fix didn't add semi-naive rule to match the variable bonded function call, and this fix conservatively moves all let binding to the body to ensure all functions are matched. 
This PR also removes `Action::SetNoTrack`, which should be unnecessary, and turn off the semi-naive transformation in final_desugar pass to avoid auto-generated let binding being moved again. 
A possible optimization is to do liveness analysis instead of moving all let binding to the body, but might be complex.
However, this fix is blocked by #143, as a test cannot pass without #143 being fixed.


mwillsey:
Subsumed by https://github.com/egraphs-good/egglog/pull/197

Issue #144, opened 2023-05-08
Visualization of EGraph State


saulshanabrook:
I am starting to look into adding some interactive visualizations to egglog-python and wanted to open an issue here, thinking I might start with adding some visualizations upstream.

I think could be useful for explaining e-graphs to new audiences and developing a mental model in particular for how egglog works.

To start with, I was hoping to generate graphs similar to those on the [`egg`](https://egraphs-good.github.io/) website:

![](https://egraphs-good.github.io/egraphs.svg)

I have started to look through the internal structure to pull out the pieces I would need to traverse:

* `EGraph` struct
  * `functions`: `HashMap<Symbol, Function>`
    * In the `Function` struct, can get the 
    *  In the `Function` struct, the `nodes.vals` (`Vec<(Input, TupleOutput)>`)  returns a number of input (list of `Value`s) output (`Value`) pairs which define the nodes of the function
         * `Value` has a `tag` which defines which type it is and a `bits` to store the value. For the primitive types, this is the value. For an "Eq Sort" i.e. user defined type, this is a lookup in the `unionfind`
 * `unionfind`: `UnionFind`
   * `parents`: Vector where each index is an ID and the values point to the parent item. 

For `egglog`, I was thinking I could generate something in a `.dot` syntax, which could be visualized using the `graphviz` CLI as well as [in the browser](https://github.com/hpcc-systems/hpcc-js-wasm/). I could either generate the strings directly or rely on a rust library like [`graphviz-rust`](https://github.com/besok/graphviz-rust) to create them.

To create the graph I was going to try:

1. Traversing the `parents` in the `unionfind` structure to create a cluster per union, with a node in each cluster item.
2. Traversing all the values of each all functions, setting the text of each node to the function name, and adding an edge between the output and each input. For non-eq/primitive nodes, I would create a new node in the graph for each. For output primitive nodes, I would have the text be the value as well as the function.

I am not sure how legible this will be, but wanted to see how it works out for some of the examples.

If this does work out, then I would try extending to some interactive formats. I have had luck in the past using [cytoscape.js](http://js.cytoscape.org/) for large graph visualization of programs (see https://github.com/vega/editor/pull/1023/).

mwillsey:
In my experience the e-graph quickly becomes too big to visualize, unfortunately. That was my inspiration for the "extract variants", as some kind of alternative to full state visualization.

saulshanabrook:
[...]

PR #147, opened 2023-05-22
Add EGraph Visualizations

saulshanabrook:
Adds the ability to visualize the state of an `EGraph` using Graphviz (addressing #144).

Goals:

* Help educate new users on what e-graphs are and how egglog implements them
* Help with debugging the state of an e-graph after some transitions
* Serve as a launching point for interactive visualizations down the road

Features:

* Colors each cluster based on it's sort
* Caps size of graph to reduce memory/time blowups with unreadable output
* Shows container values as nodes by adding `inner_values` method to `Sort`s
* Creates intermediate graph IR before encoding to graphviz. Could be exposed at a later time to allow other visualization frontends from say Python.
* Adds visualizer to web view, with dynamic transitions, and ability to pan and zoom.
* Adds CLI flags `--output-dot` and `--output-svg` to visualize any program
* Adds `make graphs` command to create SVG & dots of all examples
* Tests graph creation in CI

Possible next steps/follow-up issues:

* Graphviz does not allow edges to point to a cluster directly. [The best we can do is point to a node in the cluster and then set the `lhead` to clip at the cluster edge](https://graphviz.org/faq/#FaqClusterEdge). Currently we are doing a round-robin to point to nodes in each e-cluster. 
* Currently, e-classes that have nodes that point to themselves aren't rendered that well. The arrow won't stop at the cluster edge, but instead point directly to one of the nodes in the e-class. This is a [known issue with graphviz edges in clusters](https://stackoverflow.com/questions/46646181/connect-graphviz-cluster-to-itself). There are some workarounds in that linked issue we could explore.
* If we create a subgraph with `push` or `simplify` than that graph will not be included in the resulting graphviz. 

### `eqsat-basic` in web viewer

Screenshot 2023-06-07 at 1 43 16 PM

It will also transition between graphs smoothly:

https://github.com/egraphs-good/egglog/assets/1186124/f83c9bc6-7c33-42d5-ae0f-6cd2661cf3f5

### `eqsat-basic`
eqsat-basic

### `fibonacci`

fibonacci

### `fibonacci-demand`
fibonacci-demand

### `map`
map

### `rw-analysis`
rw-analysis

### `proofs`
proofs

<details><summary><h3>eqsat-basic in Python</h3></summary>
<p>

I've also been working on some Python examples, though they're not part of this PR.  I wanted to show them here to just give a sense of how that could work eventually:

*note that these screenshots were taken at a @recursecenter presentation previously and don't reflect the current graphviz styles*

In [this notebook](https://gist.github.com/saulshanabrook/b711f6f0f0d0db60aed64bc8da22774f#file-eqsat-basic-ipynb) we can see how the graph changes before and after running:
Screenshot 2023-05-15 at 4 05 02 PM

In [this other notebook](https://gist.github.com/saulshanabrook/b711f6f0f0d0db60aed64bc8da22774f#file-graphviz-interacitve-ipynb) we use [`d3-graphviz`](https://github.com/magjac/d3-graphviz) to animate the transitions between different graphs, doing one run in between each snapshot:

https://github.com/egraphs-good/egglog/assets/1186124/ad63c29f-59c2-4d95-8b4b-705491539aa9

</p>
</details> 

<details><summary><h2>TODO</h2></summary>
<p>

- [x] Resolve all clippy errors
- [x] Rename `Graph` to something more descriptive such as`EGraphState` or `ExportedEGraph`
- [x] Initially, keep `Graph` structure private to this crate. This could later be useful for the Python bindings to use with other visualization formats like [cytoscape.js](https://js.cytoscape.org/)
- [x] Add CLI commands to output `.dot` as well as `.svg`, instead of always emitting it
- [x] Test graph output on all examples (`make test-graphs`)
- [x] Investigate why some functions are added twice (realized it was based on ordering of arguments)
- [x] Make style closer to [e-graphs good website](https://egraphs-good.github.io/), by making e-class borders dotted, and decreasing size of arrowheads, making nodes square, and making borders round
- [x] Ensure all docstrings comments have `///`
- [x] Make node IDs stable
- [x] Fix self arrows (to make `lhead` work with self arrows, seem to [need a dummy point](https://stackoverflow.com/questions/46646181/connect-graphviz-cluster-to-itself) outside the cluster, which ends up looking odd, so leaving as is for now)
- [x] Switch functions that return built-in values to represent those as nodes instead of as strings, so that we can represent those that points to eqsats, like maps, sets, or vecs.
- [x] Position where arrows are coming from based on argument ordering.
- [x] Add graphviz to web
- [x] on web, set size to size of container
- [x] On web, animate transitions
- [x] On web, test on large graphs
- [x] Fix proof execution
- [x] Add CLI flag to switch between showing temp nodes, and not
- [x] Fix inconsistent node spacing
- [x] Switch so that unit isn't combined
- [x] add sort as cluster label
- [x] change outline of clusters which are e-classes vs not
- [x] try adding colors based on sort of clusters
- [x] Change font of nodes to helvetica

</p>
</details>

### DIFF
--- a/src/function/table.rs
+++ b/src/function/table.rs
@@ -302,7 +302,7 @@ impl Table {
     }
 }
 
-fn hash_values(vs: &[Value]) -> u64 {
+pub(crate) fn hash_values(vs: &[Value]) -> u64 {
     // Just hash the bits: all inputs to the same function should have matching
     // column types.
     let mut hasher = BH::default().build_hasher();
--- a/src/sort/mod.rs
+++ b/src/sort/mod.rs
@@ -42,10 +42,12 @@ pub trait Sort: Any + Send + Sync + Debug {
 
     // Only eq_container_sort need to implement this method,
     // which returns a list of ids to be tracked.
-    fn foreach_tracked_values<'a>(&'a self, value: &'a Value, f: Box<dyn FnMut(Value) + 'a>) {
-        let _ = value;
-        let _ = f;
-        unreachable!();
+    fn foreach_tracked_values<'a>(&'a self, value: &'a Value, mut f: Box<dyn FnMut(Value) + 'a>) {
+        for (sort, value) in self.inner_values(value) {
+            if sort.is_eq_sort() {
+                f(value)
+            }
+        }
     }
 
     // Sort-wise canonicalization. Return true if value is modified.
--- a/src/sort/mod.rs
+++ b/src/sort/mod.rs
@@ -62,7 +62,7 @@ pub trait Sort: Any + Send + Sync + Debug {
     /// Only eq_container_sort need to implement this method,
     fn inner_values(&self, value: &Value) -> Vec<(&ArcSort, Value)> {
         let _ = value;
-        unreachable!();
+        vec![]
     }
 
     fn register_primitives(self: Arc<Self>, info: &mut TypeInfo) {
--- a/src/sort/mod.rs
+++ b/src/sort/mod.rs
@@ -42,10 +42,12 @@ pub trait Sort: Any + Send + Sync + Debug {
 
     // Only eq_container_sort need to implement this method,
     // which returns a list of ids to be tracked.
-    fn foreach_tracked_values<'a>(&'a self, value: &'a Value, f: Box<dyn FnMut(Value) + 'a>) {
-        let _ = value;
-        let _ = f;
-        unreachable!();
+    fn foreach_tracked_values<'a>(&'a self, value: &'a Value, mut f: Box<dyn FnMut(Value) + 'a>) {
+        for (sort, value) in self.inner_values(value) {
+            if sort.is_eq_sort() {
+                f(value)
+            }
+        }
     }
 
     // Sort-wise canonicalization. Return true if value is modified.
@@ -54,6 +56,13 @@ pub trait Sort: Any + Send + Sync + Debug {
         debug_assert_eq!(self.name(), value.tag);
         let _ = unionfind;
         false
+    }
+
+    /// Return the inner values and sorts.
+    /// Only eq_container_sort need to implement this method,
+    fn inner_values(&self, value: &Value) -> Vec<(&ArcSort, Value)> {
+        let _ = value;
+        unreachable!();
     }
 
     fn register_primitives(self: Arc<Self>, info: &mut TypeInfo) {
--- a/src/sort/mod.rs
+++ b/src/sort/mod.rs
@@ -42,12 +42,10 @@ pub trait Sort: Any + Send + Sync + Debug {
 
     // Only eq_container_sort need to implement this method,
     // which returns a list of ids to be tracked.
-    fn foreach_tracked_values<'a>(&'a self, value: &'a Value, mut f: Box<dyn FnMut(Value) + 'a>) {
-        for (sort, value) in self.inner_values(value) {
-            if sort.is_eq_sort() {
-                f(value)
-            }
-        }
+    fn foreach_tracked_values<'a>(&'a self, value: &'a Value, f: Box<dyn FnMut(Value) + 'a>) {
+        let _ = value;
+        let _ = f;
+        unreachable!();
     }
 
     // Sort-wise canonicalization. Return true if value is modified.
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -17,6 +17,12 @@ jobs:
     - name: Build
       run: make web
 
+    - name: Upload web artifact
+      uses: actions/upload-artifact@v2
+      with:
+        name: www
+        path: target/www
+
     - name: Deploy
       uses: peaceiris/actions-gh-pages@v3
       # only actually deploy if pushed to main branch
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -18,7 +18,7 @@ jobs:
       run: make web
 
     - name: Upload web artifact
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@v3
       with:
         name: www
         path: target/www
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -243,6 +243,7 @@ version = "0.1.0"
 dependencies = [
  "clap",
  "env_logger",
+ "getrandom",
  "glob",
  "graphviz-rust",
  "hashbrown 0.13.1",
@@ -356,8 +357,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c05aeb6a22b8f62540c194aac980f2115af067bfe15a0734d7277a768d396b31"
 dependencies = [
  "cfg-if 1.0.0",
+ "js-sys",
  "libc",
  "wasi",
+ "wasm-bindgen",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,9 @@ clap = {version = "4", features = ["derive"], optional = true}
 env_logger = {version = "0.10", optional = true}
 
 ordered-float = {version = "3.4"}
+
 graphviz-rust = "0.6.2"
+# Need to add "js" feature for "graphviz-rust" to work in wasm
 getrandom = {version="*", features = ["js"]}
 
 [build-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ env_logger = {version = "0.10", optional = true}
 
 ordered-float = {version = "3.4"}
 graphviz-rust = "0.6.2"
+getrandom = {version="*", features = ["js"]}
 
 [build-dependencies]
 glob = "0.3.1"
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ ${DIST_WASM}: ${RUST_SRC}
 	wasm-pack build web-demo --target no-modules --no-typescript --out-dir ${WWW}
 	rm -f ${WWW}/{.gitignore,package.json}
 
-graphs: $(patsubst %.egg,%.svg,$(filter-out tests//fail-typecheck/%, ${TESTS}))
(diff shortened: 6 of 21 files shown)


saulshanabrook:
Ok, this is ready for review! Sorry about the large PR size. Once I started working on it, I couldn't help but tweak it a bunch to try making it look OK.

src/function/table.rs
```diff
@@ -313,7 +313,7 @@ fn hash_values(vs: &[Value]) -> u64 {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-struct Input {
+pub(crate) struct Input {
```
saulshanabrook, line 316: I made this public in the crate so that I could access every call of every function efficiently.

src/sort/mod.rs
```diff
@@ -56,6 +56,13 @@ pub trait Sort: Any + Send + Sync + Debug {
         false
     }
 
+    /// Return the inner values and sorts.
```
saulshanabrook, line 59: At first I had re-written `foreach_tracked_values` to use this new function, but then reverted to not degrade that performance at all.

oflatt:
Woah! Amazing!
I was just about to implement graphviz stuff

oflatt:
Quick question: what does it do on the demo when the egraph is huge? We might want to now show it above a certain size

oflatt:
I'm also having trouble running this locally on a mac
image
No error message besides this

oflatt:
I fetched, unfortunately, that wasn't it.
Console:
image

oflatt:
No luck still
image

src/sort/mod.rs
saulshanabrook: In a small test on my computer, rewriting `foreach_tracked_values` to use this function didn't slow things down, so I changed it to do that again, to reduce code duplication.

oflatt:
Very strange- I did `cargo clean` and removed `target` altogether but I'm still getting it.
Could it be possible that my other egglog repository clones on my machine are causing problems?
Thanks for the help

saulshanabrook:
> Indeed, downloading the github web artifact and serving it works for me (:

Great!

> Very strange- I did cargo clean and removed target altogether but I'm still getting it.
> Could it be possible that my other egglog repository clones on my machine are causing problems?

I am not sure, that is very strange.

oflatt:
Small feedback:
IMO we shouldn't extract variables- I think it's confusing that they are in the egraph (it's just our internal representation). Functions have a `is_variable` field

oflatt:
That's true too- maybe we should have a small label that shows the bindings?

oflatt:
I made a small PR to your branch- one problem I'm having is that when I generate a PNG file there's a ton of white space around the visualization. Not sure why

saulshanabrook:
> I made a small PR to your branch- one problem I'm having is that when I generate a PNG file there's a ton of white space around the visualization. Not sure why

Yeah, I have noticed that too, with the `svg`s which are generated as well... I am not sure why.

mwillsey:
Just a heads up; I don't think I'll have time to properly review this before PLDI, so if you need this code you'll have to use a branch for now!

saulshanabrook:
> Just a heads up; I don't think I'll have time to properly review this before PLDI, so if you need this code you'll have to use a branch for now!

Thx for the warning, sounds good!

saulshanabrook:
@oflatt asked:

> For function outputs that are primitive, we seem to be putting the primitive in a yellow box with the function
Maybe it should be an arrow to the primitive instead, like for other things?

Yeah, currently functions with output that are primitives have those primitives appear in the output box.

The arrow currently represents function inputs, not outputs. Outputs all appear in the same box as the function. With e-classes, this means just other function calls, whereas with primitives this is a value.  The border of the box for primitives is dotted rather than dashed borders for e-classes.

The color is based on the sort, which is not consistent between different graphs.

Originally, I had the output values that are primitives as just text in the function which produced them. However, this doesn't work for collection outputs, which could themselves point to other e-classes.

oflatt:
I'm having a lot of trouble fitting the d3 svg to the outer graph container
image

I wonder if it's my display scaling or something

oflatt:
More info:
image
image


Just scaling by `window.devicePixelRatio` didn't solve it

saulshanabrook:
@oflatt Yeah, I also had some trouble with this. It often scales a bit smaller than it should. 

There is [a StackOverflow issue](https://stackoverflow.com/questions/67626414/scale-and-center-d3-graphviz-graph) and a github issue (https://github.com/magjac/d3-graphviz/issues/195) about it... I tried to resolve it but had trouble figuring it out preserving the transitions.

Since you can zoom and pan manually, I thought it might be workable enough for now, but would be happy for any fixes to it.

saulshanabrook:
@oflatt I believe if we remove this it won't reset the zoom/pan:

https://github.com/egraphs-good/egglog/blob/1d92c82e28a72caac85b7dfa409a73fc5559e012/web-demo/static/index.html#L109-L111

oflatt:
Thanks! Worked for me

oflatt:
What's the status on this? Does @mwillsey have time to review?
Saul and I discussed it some- I think the conclusion was to somewhat simplify the intermediate representation Saul made. Eventually we want only one format that extraction, visualization, and dump to text uses

saulshanabrook:
> Saul and I discussed it some- I think the conclusion was to somewhat simplify the intermediate representation Saul made. Eventually we want only one format that extraction, visualization, and dump to text uses

@oflatt I saw that you opened your PR that also has its own IR for the graph. I would be happy to commit to merging our two IRs once both PRs are merged, if that's easier as well. In that merge, it might also be good to think about having a format we could expose in Python to allow other forms of extraction easily, like the work @philzook58 [was experimenting with](https://egraphs.zulipchat.com/#narrow/stream/328972-general/topic/ASP.20Extraction/near/371525784).

oflatt:
Yep, I'm in favor of merging this and fixing it later then.
We should also coordinate later with @mwillsey's new format for the extraction gym:
https://github.com/egraphs-good/extraction-gym

mwillsey:
After conversation with @oflatt, I really think we can use a common format for both this and the extraction gym. So I'm going to close this PR for now, as I do not intend to merge it in it's current state. Details about the format are coming soon!

saulshanabrook:
@mwillsey So would you recommend implementing this on top of the new common format and then re-opening?

FWIW the Python bindings are already published with this fork and I will continue including this code in my fork and keeping it up to date, because I have been finding the graphviz helpful for education.

mwillsey:
Yes! I think the idea is that you’d use the egraph serialize library as a dependency. Egglog (and egg) will hopefully soon have support for exporting to the in-memory representation of that library, thereby supporting not only serialization but also hopefully visualization. So the Python bindings could use mainline and still have visualization.

saulshanabrook:
My plan for following up with this work is as follows:

- [x] Add serialization support to egglog #171 
- [ ] Add support for converting serialized format to graphviz in the https://github.com/egraphs-good/egraph-serialize/ repo (under a feature flag, so graphviz requirements are optional), with methods to produce the string fo the graphviz, and save as a `.dot` and `.svg` file
- [ ] Add support for splitting out all nodes of certain sorts into their own e-classes as a method on `Egraph` in `egraph-serialize` (so that fns which return or take primitives can not share them in the viz)
- [ ] Update this branch to use the graphviz support from serialize to save `svg`s to disk and in the web UI.

mwillsey:
Yes, that sounds good! I would even go further and say that any "singleton" e-class (an e-class with just a single node, like all primitives) could be inlined directly into the parent for easier visualization.

saulshanabrook:
Yeah, I think that might work too... let me take a look at some examples from this branch to see what would make sense.... 

Here are a couple of examples from three consecutive commits in the history of this branch:

```shell
git checkout <hash>
cargo run tests/<name>.egg --save-svg
# Screenshot to convert to cropped PNG due to bug at these commits with size
```

|  | `77e80d4` |  `9da623c` | `7ddabb8` |
| -- | -- | -- | -- |
|  | All equal primitives in shared node. <br /> Current behavior of **exporter**. |  Unit primitives in their own nodes. | All primitives in their own nodes. <br /> Current behavior of **this branch**. |
| `fibonacci` | Screenshot 2023-07-21 at 12 16 00 PM | Same as ← |  Screenshot 2023-07-21 at 12 15 34 PM |
| `fibonacci-demand` |  Screenshot 2023-07-21 at 12 13 48 PM  | Same as ← |  Screenshot 2023-07-21 at 12 13 48 PM |
| `path` |  Screenshot 2023-07-21 at 12 28 09 PM | Screenshot 2023-07-21 at 12 55 20 PM | Screenshot 2023-07-21 at 12 56 42 PM  |

Does anyone have thoughts on which are preferable? Happy to add other examples too.

mwillsey:
For things of type unit, what makes the most sense to me is to group by function name. So basically approach 1, but split up by function name; so all the path tuples in one box, edge tuples in another, and so on. I think we can just elide the actual `()` node.

For other primitives (i64, etc), I'd like to see what the inlining approach looks like for when the primitives are used as inputs to functions. A quick example:

```
Node
|  |
v  v
1  foo
```

Could become:

```
Node(1, ·)
|
v
foo
```

in the situation where `1` is in a singleton e-class (or maybe is a primitive or something) but `foo` is not.

saulshanabrook:
> For other primitives (i64, etc), I'd like to see what the inlining approach looks like for when the primitives are used as inputs to functions. A quick example:


That's a cool idea! It would definitely cut down on the number of nodes...

We would still put collection primitives as nodes b/c they can point to e-classes...?

For functions that return primitives, we could do `Node(1, ·) -> 2` too?

mwillsey:
Sure! We'll have to try it to see

saulshanabrook:
I have added these examples, along with the existing examples, to the e-graph serialize PR, so that we can see how inlining compares! https://github.com/saulshanabrook/egraph-serialize/tree/viz/tests-viz

saulshanabrook:
@mwillsey I have updated this PR to use the e-graph serialize graphviz implementation (https://github.com/egraphs-good/egraph-serialize/pull/4 and #171).

The changes seem out of date in the GitHub diff, even though I pushed them to the branch. Maybe if you re-open it they will be refreshed?

PR #149, opened 2023-06-01
Fix subtleties in rebuilding

### LINKED ISSUE #121: `math_with_proofs` fails in debug builds
The `math_with_proofs` test triggers an assertion failure:
```
thread 'math_with_proofs' panicked at 'assertion failed: `(left == right)`
  left: `Value { tag: "Proof__", bits: 59896 }`,
 right: `Value { tag: "Proof__", bits: 10121 }`: [19586] EqGraph__([Value { tag: "Ast__", bits: 99019 }, Value { tag: "Ast__", bits: 59914 }]) = TupleOutput { value: Value { tag: "Proof__", bits: 59896 }, timestamp: 14 }
ResolvedSchema { input: [EqSort { name: "Ast__" }, EqSort { name: "Ast__" }], output: EqSort { name: "Proof__" } }', src/lib.rs:277:14
```

The relevant assertion that's failing is:
```
                assert_eq!(
                    output.value,
                    self.bad_find_value(output.value),
                    "[{i}] {name}({inputs:?}) = {output:?}\n{:?}",
                    function.schema,
                )
```

So rebuilding missed this node! This is probably an edge-case in rebuilding somewhere

ezrosent:
Fixes a couple small bugs:

The method `rebuild_at` was slightly wrong in a couple ways:
1. Merge functions should run on canonicalized values, meaning we should take the previous value in a given function row and canonicalize it before merging it.
2. We should not unconditionally remove the old row in a table: the keys to a row may not have changed. Instead, run the merge function first and only remove the old row if needed. 

On top of that there was (I think) an unintentional variable shadowing in the `add` method for "composite" indexes.

Fixes #121

### DIFF
--- a/src/function/index.rs
+++ b/src/function/index.rs
@@ -79,8 +79,8 @@ impl CompositeColumnIndex {
     }
 
     pub(crate) fn add(&mut self, v: Value, i: usize) {
-        if let Some(i) = self.0.iter().position(|index| index.sort() == v.tag) {
-            (self.0)[i].add(v, i);
+        if let Some(index) = self.0.iter().position(|index| index.sort() == v.tag) {
+            (self.0)[index].add(v, i);
         } else {
             let mut index = ColumnIndex::new(v.tag);
             index.add(v, i);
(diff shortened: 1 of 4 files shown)


oflatt:
Thanks for the fix! You are a hero!

oflatt:
We should merge this

PR #153, opened 2023-06-15
Vec fix, tests, and improvements

### LINKED ISSUE #152: Vec include getting items and length?
I was looking at the Vec sort and as far as I can tell it doesn't have a way to retrieve an item at a certain index or update an item at a certain index? It seems to only contain a way to check if an item exists.

It seems similar to a set then, but maybe allowing duplicates and ordered.

I was curious if you would be open to a PR which adds the ability to get an item from a particular index as well as one which returns the length.

saulshanabrook:
* Fixes a typo so that the `Vec` sort will register a Vec instead of a Set
* Fixes the function name for extracted vecs
* Adds test for existing Vec methods
* Add vec get and length (closes #152)


yihozhang:
Great catch!

Issue #113, opened 2023-02-22
Created map type is always first defined map type


saulshanabrook:
I have noticed a slight oddity in the current type system.  The order in which you defined map sorts influences the type of maps you create.

## Example

For example, this works:

```
(sort MyMap (Map i64 String))
(sort MyMap1 (Map i64 i64))

(define my_map1 (insert (empty) 1 "one"))
```

whereas this fails:

```
(sort MyMap1 (Map i64 i64))
(sort MyMap (Map i64 String))

(define my_map1 (insert (empty) 1 "one"))
```

This means you cannot create two different types of maps in one level. 

## Why does it happen

When you register a map sort, like `(sort MyMap (Map i64 String))`, this will register the primitives, including `create`. If you do this twice, then two primitives will be registered under the name `create`. However, when you call `create`, the first one registered will be chosen. 

## Workaround
You can get around this by pushing a level, creating a map, then going up a level:

```
(push)
(sort MyMap1 (Map i64 i64))
(define my_map1 (insert (empty) 1 1))
(pop)

(sort MyMap (Map i64 String))

(define my_map2 (insert (empty) 1 "one"))
```

## Solution

I am not sure if there an elegant way to address this oddity at the moment. I am not sure if it would be classified as a bug. 

My understanding of the underlying issue is that type system of egg-smol is mainly [the simply typed lambda calculus](https://en.wikipedia.org/wiki/Lambda_cube#(%CE%BB%E2%86%92)_Simply_typed_lambda_calculus). Users can create custom type (sorts) and define function which map between different types. Users cannot create functions that are generic or create types that depend on other types.

Builtin maps are supported as the only parametric types, but they are added in such a way as to make them appear as simple types.

This leads to this issue, since there is no way to pass a type into a function. To solve it, I believe we would need to be able to pass a type into the `create` function to parameterize it with something like `(create (Map i64 i64))` or `(create i64 i64)`.

This would mean changing the type system of egg-smol from a simply typed lambda calculus to the [System Fω](https://en.wikipedia.org/wiki/Lambda_cube#(F%CF%89)_System_F%CF%89), since it requires having types that depend on types (Map) and terms that depend on types (create). This would also allow users to define other custom types, such as lists or simply typed functions (like a lambda from int to str).

I am not suggesting this one oddity would be enough motivation for such a large change, since it is possible to work around it currently. But I wanted to open this issue to write out my thoughts here, since I had been thinking about the possibility of adding user-defined parametrics for a while :)

[...]

PR #155, opened 2023-06-19
Builtin Lambda Calculus Experiment

saulshanabrook:
*[zulip thread](https://egraphs.zulipchat.com/#narrow/stream/375765-egglog/topic/embedded.20lambda.20functions/near/368338074) discussing this PR*

This is a test branch to experiment with a way of allowing users to define lambda terms in the e-graph.

Current (known) limitations:
* Only works when functions input as an e-class not a primitive
* Function with duplicate variable names will break. (Could have a fix for this with desugaring pass)
* Can only use functions with one sort currently, i.e. if you want i64 -> i64, cannot also use i64 -> String in one program. I believe this is partially due to  https://github.com/egraphs-good/egglog/issues/113. This means we can't do higher order functions sadly!

Changes:
* Add new `Lambda` builtin sort, with functions `var`,`lambda`, and `apply`.
* Pass the `EGraph` into `apply()` method on sorts so that when applying a lambda function can muck about in the EGraph
* Add a `register_egraph` method to sorts, so that the Lambda sort can register two regular functions `var` and `lambda`. I couldn't figure out a way to do this otherwise, registering non-primitive functions in a sort based on the types based in.
* Adds a `vec-map` function for vectors map a function across them.

Examples:

```lisp
(define double (lambda (var "x") (Mul (var "x") (Num 2))))
(define two (apply double (Num 1)))
(run 1)
(check (= two (Num 2)))
```
lambda-builtin

```lisp
(sort MathVec (Vec Math))
(define vec (vec-of (Num 1) (Num 2) (Var "3")))
(define doubled-vec (vec-map vec double))
(run 1)
(check (= doubled-vec (vec-of (Num 2) (Num 4) (Mul (Var "3") (Num 2)))))
```
lambda-builtin

### DIFF
--- a/src/function/table.rs
+++ b/src/function/table.rs
@@ -51,7 +51,7 @@ pub(crate) struct Table {
     max_ts: u32,
     n_stale: usize,
     table: RawTable<TableOffset>,
-    vals: Vec<(Input, TupleOutput)>,
+    pub(crate) vals: Vec<(Input, TupleOutput)>,
 }
 
 /// Used for the RawTable probe sequence.
@@ -304,7 +304,7 @@ fn hash_values(vs: &[Value]) -> u64 {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-struct Input {
+pub(crate) struct Input {
     data: ValueVec,
     /// The timestamp at which the given input became "stale"
     stale_at: u32,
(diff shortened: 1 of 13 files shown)


saulshanabrook:
Closing this because it's rather outdated and https://github.com/egraphs-good/egglog/pull/348 was merged which is another way to approach this problem.

Issue #139, opened 2023-04-28
extract:variants panic in base type


clyben:
```
(function f (i64) i64 :merge (max old new))

(set (f 1) 1)

(extract:variants 1 (f 1))
(extract:variants 2 (f 1))
```

The first extract is ok but the second extract panicked. When the function map to user-defined types, it won't panic.
Should the second extract behave the same as the first one?

mwillsey:
Sure, just returning fewer variants always seems ok.

yihozhang:
Fixed by #156

PR #156, opened 2023-06-21
Fix table len and extract:variants panic

### LINKED ISSUE #120: `Table::len()` is not computing the right size because of stale entries
This causes issues with `print-size` and the printed summary. I made an ad hoc fix locally for `print-size` which the artifact relies on but we need a more careful fix.

clyben:
This PR fixed #120 and #139 
Fixing #120 by adding a new function num_offsets() in table.rs, referring to the original len() function, and changing all original len() call to num_offsets() calls.
Fixing #139 in lib.rs extract_expr's default case, only call self.extract_variants() when it's a sort, excluding the case when it's a base type.


src/function/table.rs
```diff
@@ -182,13 +182,18 @@ impl Table {
     }
 
     /// One more than the maximum (potentially) valid offset into the table.
-    pub(crate) fn len(&self) -> usize {
+    pub(crate) fn num_offsets(&self) -> usize {
         self.vals.len()
     }
 
+    /// One more than the actual valid offset (not including stale) into the table.
```
oflatt, line 189: Is this comment right?

clyben: I think so? The table len not including stale should be the actual valid offset?

oflatt:
Looks good to me

PR #158, opened 2023-06-28
New Term Encoding

oflatt:


oflatt:
This PR makes egglog not much slower!???
```  Summary [   5.136s] 72 tests run: 72 passed, 0 skipped```
VS
```Summary [   5.039s] 67 tests run: 67 passed, 0 skipped```

File: src/termdag.rs [hunk 1]
```diff
@@ -0,0 +1,148 @@
+use crate::{
```
oflatt, line 1: TODO migrate to new EGraph package instead of termdag

oflatt:
Update on this PR: A couple benchmarks are 100x slower due to the new encoding,
Max and I are working on making the query compiler smarter / changing the encoding to get back the performance.

oflatt:
Closing for now- this will probably be the right way to do things, but we need a better backend.

PR #160, opened 2023-07-04
Add support for third party sorts

saulshanabrook:
This PR adds support for third party packages to add their own sorts.

I am using it in https://github.com/metadsl/egglog-python/pull/31 to add a custom `PyObject` sort, with primitive operations for executing code and converting.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -21,7 +21,7 @@ use proofs::ProofState;
 use symbolic_expressions::Sexp;
 
 use ast::*;
-use typechecking::{TypeInfo, UNIT_SYM};
+pub use typechecking::{TypeInfo, UNIT_SYM};
 
 use std::fmt::{Formatter, Write};
 use std::fs::File;
@@ -35,7 +35,7 @@ use std::rc::Rc;
 use std::{fmt::Debug, sync::Arc};
 use typecheck::Program;
 
-type ArcSort = Arc<dyn Sort>;
+pub type ArcSort = Arc<dyn Sort>;
 
 pub use value::*;
 
@@ -1230,6 +1230,10 @@ impl EGraph {
 
     pub(crate) fn get_sort(&self, value: &Value) -> Option<&ArcSort> {
         self.proof_state.type_info.sorts.get(&value.tag)
+    }
+
+    pub fn add_sort<S: Sort + 'static>(&mut self, sort: S) {
+        self.proof_state.type_info.add_sort(sort);
     }
 
     // Gets the last extract report and returns it, if the last command saved it.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1236,7 +1236,6 @@ impl EGraph {
         self.proof_state.type_info.add_arcsort(arcsort)
     }
 
-
     // Gets the last extract report and returns it, if the last command saved it.
     pub fn get_extract_report(&self) -> &Option<ExtractReport> {
         &self.extract_report
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -719,7 +719,7 @@ impl EGraph {
         Ok(())
     }
 
-    fn eval_expr(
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -21,7 +21,7 @@ use proofs::ProofState;
 use symbolic_expressions::Sexp;
 
 use ast::*;
-use typechecking::{TypeInfo, UNIT_SYM};
+pub use typechecking::{TypeInfo, UNIT_SYM};
```
saulshanabrook, line 24: Must make public so that can access `TypeInfo` which is an argument to a method on sorts

src/lib.rs
```diff
@@ -35,7 +35,7 @@ use std::rc::Rc;
 use std::{fmt::Debug, sync::Arc};
 use typecheck::Program;
 
-type ArcSort = Arc<dyn Sort>;
+pub type ArcSort = Arc<dyn Sort>;
```
saulshanabrook, line 38: Not strictly necessary but a useful alias

src/lib.rs
```diff
@@ -719,7 +719,7 @@ impl EGraph {
         Ok(())
     }
 
-    fn eval_expr(
+    pub fn eval_expr(
```
saulshanabrook, line 722: I started using this to eval the Python expr (which looks like `(py-object-hash 123123 123123)`) to the value, so then I can retrieve the Python object from it.

src/lib.rs
```diff
@@ -1232,6 +1232,11 @@ impl EGraph {
         self.proof_state.type_info.sorts.get(&value.tag)
     }
 
+    pub fn add_arcsort(&mut self, arcsort: ArcSort) -> Result<(), TypeError> {
+        self.proof_state.type_info.add_arcsort(arcsort)
+    }
+
+
```
saulshanabrook, line 1239: Originally I just exposed the `add_sort`, but I needed the more general `add_acrsort` so that I can save the sort in Python.

yihozhang:
Interesting! I did some hacking a while ago trying to support user-defined appliers for [szalinski](https://github.com/yihozhang/szalinski-egglog), which has some custom appliers for list manipulation. Appliers are basically primitives with a handle to the e-graph so that it can have side effects. I kinda abandoned that project though because AFAIR to type-check these user-defined primitives/appliers requires access to handles to types like `List`, and this requires some refactoring I remember.

Do you have a similar use case where you need to define your own primitives for types (that you don't own)? It seems all you need is just to make type_info public?

saulshanabrook:
> Do you have a similar use case where you need to define your own primitives for types (that you don't own)? It seems all you need is just to make type_info public?

I just added some primitives for converting to/from strings with some test cases: https://github.com/metadsl/egglog-python/pull/31/commits/943acc08f00c6c03b0e6e462396650d0b8121cf0 Is this what you mean?

> Appliers are basically primitives with a handle to the e-graph so that it can have side effects. I kinda abandoned that project though because AFAIR to type-check these user-defined primitives/appliers requires access to handles to types like List, and this requires some refactoring I remember.

This sounds a bit similar to what I was experimenting with in the lambda calculus PR https://github.com/egraphs-good/egglog/pull/155, where I added the egraph to the `apply` args so that I could mutate in the primitive applications:

https://github.com/egraphs-good/egglog/blob/d5c85e38b6c9818c3945cde395d214a7a71b667d/src/sort/lambda.rs#L148-L177

yihozhang:
Ah I see. Thanks, very interesting! I also saw that you use `type_info.get_sort()` to get the sort you need for type checking. So it now seems that the idea of porting Szalinski to egglog is feasible as you've done something similar in your lambda calculus and egglog-python.

saulshanabrook:
Ah cool, yeah feel free to ping me if you have any other questions about the approach.

PR #161, opened 2023-07-07
Update README Python binding doc link

bskinn:
README links to https://egglog-python.readthedocs.io/en/latest/, but that page 404s.

It appears these docs are at https://egg-smol-python.readthedocs.io/en/latest/?

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ This is repo for the `egglog` tool accompanying the paper
 If you use this work, please use [this citation](./CITATION.bib). 
 
 See also the Python binding, which provides a bit more documentation:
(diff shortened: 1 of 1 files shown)


mwillsey:
Good catch! I think I got a little overzealous with find-and-replace when I renamed this repo from `egg-smol` to `egglog`. @saulshanabrook, should the `egg-smol-python` package be renamed?

saulshanabrook:
@mwillsey Yeah I had an issue for that, they could not rename it, said I will have to delete and remove the project which I can do https://github.com/metadsl/egglog-python/issues/18

PR #162, opened 2023-07-10
add log2 for i64 to help qualify shift rewrites

ryan-berger:


mwillsey:
Uh oh, looks like this is an unstable library function?

mwillsey:
I merged from main, where we updated our rust verison

ryan-berger:
@mwillsey Sorry, just saw this. ilog2 only recently was accepted as a proposal. Glad y'all were able to just update Rust to fix the issue, wasn't sure what you'd decide to do.

Thanks again!

PR #171, opened 2023-07-18
Add serialization support

saulshanabrook:
This PR adds serialization support to the [`egraph-serialize` format](https://github.com/egraphs-good/egraph-serialize).

It pulls some ideas and code from the viz pr (#147) around requirements for the serialization format.

I also started using the [insta](https://github.com/mitsuhiko/insta) library for snapshot testing, to verify that certain examples are serialized properly. So far, I just tried to manually look over them to make sure they seem sensible.

Once the viz work is redone on top of this, it will also be easier to quickly verify the results.

The choice of how to determine the node IDs is not the most straightforward one, but conforms to certain properties that were useful for the graphviz work, which would use these directly (since in graphviz you can't actually point to a cluster, you have to point to nodes).

[Link](https://egraphs.zulipchat.com/#narrow/stream/328972-general/topic/e-graph.20serialization.20format/near/374049206) to some relevant convo on zulip around how to serialize.


src/sort/mod.rs
```diff
@@ -56,6 +58,13 @@ pub trait Sort: Any + Send + Sync + Debug {
         false
     }
 
+    /// Return the inner values and sorts.
+    /// Only eq_container_sort need to implement this method,
+    fn inner_values(&self, value: &Value) -> Vec<(&ArcSort, Value)> {
+        let _ = value;
+        vec![]
+    }
+
```
saulshanabrook, line 67: I added this so that we can properly handle e-classes inside of collections. If we just use `extract` (what I tried previously) then it will only give us a particular e-node from a collection, instead of referencing the full e-class.

mwillsey:
While there are definitely some benefits to snapshot testing, I'm not ready to commit to it with this PR. Can you please back it out for now?

Otherwise the PR looks great!

saulshanabrook:
Thanks! Sure, I just removed the tests.

PR #174, opened 2023-07-26
Fix serialize checking for temp name

saulshanabrook:
Switches the check in the serialize to use the `number_underscores` variable instead of hardcoding them so it is more accurate.

Discovered when adding test files to egraph-serialize with `--save-json` CLI command https://github.com/egraphs-good/egraph-serialize/pull/4


oflatt:
Good idea!

PR #175, opened 2023-07-28
Add conversions between ints and floats

saulshanabrook:
This PR adds conversions between ints and floats.

I wasn't sure if I should include the argument type name in the function name (i.e. `f64-to-i64` instead of `to-i64`), but I opted not to since there is already a `to-f64` in the `Rational` sort. However, I did notice that the `Map` and `Set` functions are prefixed with the argument names, so I am unclear on if there is a best practice here.

I need this in https://github.com/metadsl/egglog-python/pull/31 in an example replicating the [Array API](https://data-apis.org/array-api/2022.12/index.html#)


mwillsey:
We don't have a strong convention at the moment, so I think this is fine!

PR #176, opened 2023-07-31
Cleanup from terms PR

oflatt:
This PR:
- Adds more desugaring to clean up the front end more
   - Got rid of define in favor of let
   - All run statements are converted to schedules
- Handles globals as an environment instead of declaring tables
- Fixes several bugs in saturation
   - One bug where we were inserting new timestamps too eagerly
   - One bug where saturation relies on seminaive
- Gets rid of proofs code because they don't work
- Adds a resugaring so that we can look at the IR more easily
- New command line arguments for looking at the IR
- Various other small changes

Sorry all these changes are together, it was tough to get them from my terms PR.


src/sort/map.rs
```diff
@@ -45,6 +45,20 @@ impl MapSort {
     }
 }
 
+impl MapSort {
+    pub fn presort_names() -> Vec<Symbol> {
```
oflatt, line 49: Presort names is so bad bad but I'm not sure how else to get presort names

src/ast/desugar.rs
```diff
@@ -449,69 +587,63 @@ pub(crate) fn desugar_command(
...
+                )?
+                .into_iter()
+                .map(|cmd| cmd.command)
                 .collect()
         }
         Command::Check(facts) => {
-            let mut res = vec![NCommand::Check(flatten_facts(&facts, desugar))];
+            let res = vec![NCommand::Check(flatten_facts(&facts, desugar))];
 
             if get_all_proofs {
```
ezrosent, line 633: There's lots of stuff commented out in this block. Should it be removed? If you want to leave it in, I think it'd help to add another comment as context.

src/extract.rs
```diff
@@ -61,71 +68,114 @@ impl EGraph {
...
                 .costs
                 .get(&id)
-                .unwrap_or_else(|| panic!("No cost for {:?}", value));
-            (*cost, self.expr_from_node(node))
+                .unwrap_or_else(|| {
+                    eprintln!("No cost for {:?}", value);
+                    for func in self.egraph.functions.values() {
+                        for (inputs, output) in func.nodes.iter() {
+                            if output.value == value {
+                                eprintln!("Found unextractable function: {:?}", func.decl.name);
```
ezrosent, line 113: I think `eprintln!`s would be better off being replaced as log statements. These can probably be `log::error!`s. The nice thing about logs is that callers can opt out, which is useful when using egglog as a library.

src/gj.rs
```diff
@@ -30,44 +35,91 @@ enum Instr<'a> {
...
+struct VarInfo2 {
+    occurences: Vec<usize>,
+    intersected_on: usize,
+    size_guess: usize,
+    is_nonparent_input: bool,
+}
+
+struct InputSizes<'a> {
+    cur_stage: usize,
+    // Each stage we're intersecting {set_k}_k
```
ezrosent, line 49: Having wrote this comment, I'm not sure that it really makes sense out of context. Instead, how about "a map from from stage to vector of costs for each stage, where 'cost' is the largest relation being intersected".

src/lib.rs
```diff
@@ -841,23 +936,32 @@ impl EGraph {
                     "Skipping schedule.".to_string()
                 }
             }
-            NCommand::Extract { var, variants } => {
+            /*NCommand::Extract { var, variants } => {
```
ezrosent, line 939: Same question here: I think it'd help if large commented-out blocks like this should either be removed or documented a bit more.

src/sort/vec.rs
```diff
@@ -69,21 +83,8 @@ impl Sort for VecSort {
         result
     }
 
-    fn canonicalize(&self, value: &mut Value, unionfind: &UnionFind) -> bool {
```
ezrosent, line 72: Why is this getting backed out? Don't we still want to canonicalize vectors?

tests/files.rs
```diff
@@ -7,45 +7,51 @@ use libtest_mimic::Trial;
...
+        };
+
+        if !self.resugar {
+            self.test_program(&program, "Top level error");
+        } else if self.resugar {
             let mut egraph = EGraph::default();
-            egraph.set_underscores_for_desugaring(4);
+            egraph.set_underscores_for_desugaring(3);
             let parsed = egraph.parse_program(&program).unwrap();
+            // TODO can we test after term encoding instead?
```
ezrosent, line 31: Not sure what this comment refers to: what gets tested post-term encoding?

ezrosent (commented):
Took a quick pass here.

tests/files.rs
oflatt: This test tests that desugaring and resugaring round-trip still works. I was wondering if we could add a round-trip with term encoding (but it's no longer relevant so I'll remove it)

src/sort/vec.rs
oflatt: Oof, good catch, it shouldn't be backed out.
We really need better testing for containers

src/extract.rs
oflatt: good idea

PR #182, opened 2023-08-10
Change remaining README uses of `define` to `let`

wilcoxjay:
I think this was an oversight when removing `define` in #176.


wilcoxjay:
cc: @oflatt

PR #183, opened 2023-08-10
Print extracted terms to stdout

oflatt:


src/typecheck.rs
```diff
@@ -886,17 +886,23 @@ impl EGraph {
...
+                        log::info!("extracted with cost {cost}: {}", extracted);
+                        println!("{}", extracted);
                     } else {
                         if variants < 0 {
                             panic!("Cannot extract negative number of variants");
                         }
                         let extracted =
                             self.extract_variants(values[0], variants as usize, &mut termdag);
                         log::info!("extracted variants:");
+                        println!("(");
```
oflatt, line 899: This wraps the variants in parenthesis for easy parsing as a unit.

PR #186, opened 2023-08-11
Exposes visualizations in CLI and in the web

saulshanabrook:
This PR adds the ability to export an egglog graph to SVG via the command line and also renders a live graph in the web viewer. 

It uses the serialization support added in #171 and the graph support in the serialization repo added in https://github.com/egraphs-good/egraph-serialize/pull/4 to do the heavy lifting.

It is the same branch as #147, but I have opened a new PR so that the diff is clearer (the old diff seemed to stop being updated once it was closed).

Also if we want to refer to the changes in that old PR for posterity, we can continue looking at it in its frozen state.


.github/workflows/deploy.yml
```diff
@@ -17,11 +17,17 @@ jobs:
     - name: Build
       run: make web
 
+    - name: Upload web artifact
+      uses: actions/upload-artifact@v3
+      with:
+        name: www
+        path: target/www
+
```
mwillsey, line 25: What's going on here? What is the artifact used for?

src/serialize.rs
```diff
@@ -51,7 +51,7 @@ impl EGraph {
     /// - Functions which return primitive values will be added to the e-class of that value.
     /// - Nodes will have consistant IDs throughout execution of e-graph (used for animating changes in the visualization)
     /// - Edges in the visualization will be well distributed (used for animating changes in the visualization)
-    ///   (Note that this will be changed in https://github.com/egraphs-good/egglog/pull/158 so that edges point to exact nodes instead of looking up the e-class)
+    ///   (Note that this will be changed in `<https://github.copom/egraphs-good/egglog/pull/158>` so that edges point to exact nodes instead of looking up the e-class)
```
mwillsey, line 54: typo here

.github/workflows/deploy.yml
saulshanabrook: I was using it to manually verify the website contents in this PR (to make sure it was built correctly). Otherwise, there isn't a way in a PR to verify what the website will look like (since it isn't deployed from other branches).

I am happy to either:

* Remove this
* Add a comment to say you can manually download the artifact to look at the website

What would you prefer?

src/serialize.rs
saulshanabrook, line 54:
```suggestion
    ///   (Note that this will be changed in `<https://github.com/egraphs-good/egglog/pull/158>` so that edges point to exact nodes instead of looking up the e-class)
```

.github/workflows/deploy.yml
mwillsey: That’s clever. Sure just comment it. I assume artifacts are ephemeral?

saulshanabrook: Yep, they are stored for 90 days by default it [looks like](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts#about-workflow-artifacts)?

> By default, GitHub stores build logs and artifacts for 90 days, and this retention period can be customized. For more information, see "[Usage limits, billing, and administration](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#artifact-and-log-retention-policy)." The retention period for a pull request restarts each time someone pushes a new commit to the pull request.

saulshanabrook: Added a comment which also links to this upstream issue for PR github pages deploys: https://github.com/orgs/community/discussions/7730

PR #187, opened 2023-08-12
Fixes and adds extra tests to BDD & Arrays

TrevorHansen:
For BDD, this: 
(1) fixes some operations, 
(2) reduces the number of rewrites required by adding commutative operations,  (3) adds more tests,
(4) moves the tests into a separate frame so that the changes made during testing are cleaned up when testing finishes.

For arrays, this adds a necessary addition rewrite.


tests/bdd.egg
```diff
@@ -16,81 +16,90 @@
...
-)
-;(check (= b123 (ITE 3 ()))
-
 (function or (BDD BDD) BDD)
+(rewrite (or x y) (or y x))
 (rewrite (or True n) True)
-(rewrite (or n True) True)
 (rewrite (or False x) x)
-(rewrite (or x False) x)
 (rewrite (or (ITE n a1 a2) (ITE m b1 b2))
```
TrevorHansen, line 37: Note that this uses "b1" & "b2" which have been defined above in the test code. So these aren't free variables.

PR #194, opened 2023-08-17
Minor termdag changes

### LINKED ISSUE #193: TermDag Confusion
I am taking a closer look at the [`TermDag`](https://github.com/egraphs-good/egglog/blob/main/src/termdag.rs) introduced by @oflatt in https://github.com/egraphs-good/egglog/pull/176 and am a bit confused on how the different indices function in it.

My understanding from reading the code is that:

* Each `App` term points to its arguments with an integer, that represents an index in the `nodes` vec ([cite](https://github.com/egraphs-good/egglog/blob/3f7c3fed7dde072dfc8c97b6c2dfba257a4deb7c/src/termdag.rs#L101))
* When a node is added, the hashcons is updated with the term pointing to the index of the term in the vec.

So am I understanding things correctly that the `hashcons` provides no additional information over the `nodes` vector itself? And it serves just as a way to do faster lookups from each term to its index?

Originally when I saw the `hashcons` structure, I thought it meant something about e-classes, but looking at the implementation that does not seems to be the case.

If this is so, I am happy to add a PR with some comments to make things clearer.

saulshanabrook:
This PR adds makes a few minor changes to the termdag. It adds a few comments (which will close #193) and it also makes the a few methods not mutate the termdag itself.


oflatt:
Looks like you have a compile error, thanks for comments though

saulshanabrook:
Closing due to #200 covering the documentation needs

PR #195, opened 2023-08-19
Clean up logging

yihozhang:
This PR attempts to make the logging and printing behavior of egglog consistent. In particular, it follows the following rules:

* Printing to stderr should be done using the log crate.
* Printing to stdout should be as simple as possible and intended to be consumed by another program.
* All logging (e.g., `declared XXX`) should be printed to stderr to keep stdout clean.
* No `println` is allowed besides in `main.rs`. In particular `parse_and_run_program` should not print anything to stdout. No `eprintln` is allowed anywhere (use the log crate instead).

Right now, the only commands that print to stdout are `print-size`, `print-table`, and `extract`.


mwillsey:
If you really wanted to disallow println, you could use this: https://rust-lang.github.io/rust-clippy/master/#/disallowed_macros

yihozhang:
This is a good idea. I have disallowed `(e)print(ln)` in clippy.toml

Issue #64, opened 2022-11-20
Print Line Number on Parse Error


philzook58:
Conversion of character number to line col number would be a much more pleasant programming experience. I get around it by installing an extension in my vscode editor that prints character number. Bonus mode if one can print the line with little squiggles under the problem. I attempted to do this a few months back based on these comments https://github.com/lalrpop/lalrpop/issues/323#issuecomment-366681594 but got lost somewhere in  custom error implementation egg-smol has.

mwillsey:
I think we should target using something like [miette](https://docs.rs/miette/latest/miette/index.html).

yihozhang:
Closing given #199 and #398

PR #199, opened 2023-08-21
Show parse error's line number and column number

hatoo:
This PR makes a parse error message more readable

Before 
```
[ERROR] Unrecognized token `)` found at 169:170
    Expected one of r#"(([[:alpha:]][\\w-]*)|([-+*/?!=<>&|^/%_]))+"# or r#"(-)?[0-9]+"#
```

After
```
[ERROR] Unrecognized token `)` found at ../tmp/path.egg:11:5:../tmp/path.egg:11:6
    Expected one of r#"(([[:alpha:]][\\w-]*)|([-+*/?!=<>&|^/%_]))+"# or r#"(-)?[0-9]+"#
```

This will be very helpful for new users like me.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,5 +1,5 @@
 use clap::Parser;
-use egglog::{CompilerPassStop, EGraph, SerializeConfig};
+use egglog::{CompilerPassStop, EGraph, Error, SerializeConfig};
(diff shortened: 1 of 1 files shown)


yihozhang:
Thanks for the PR! Yeah this would be very useful. Also related to #64

PR #200, opened 2023-08-21
Document src/termdag.rs

wilcoxjay:
Wrote some docs.

I also deleted this old comment:

``` rust
    // users can't construct a termnode, so just
    // look it up
```

which I believe to be wrong. `Term` is exported publicly via `src/lib.rs`:

``` rust
pub use termdag::{Term, TermDag};
```

So users *can* construct `Term`s by hand.

Because of this, I included in the documentation for various methods what the
behavior of the current implementation is when given malformed `Term`s.

### DIFF
--- a/src/termdag.rs
+++ b/src/termdag.rs
@@ -4,6 +4,9 @@ use crate::{
     Symbol,
 };
 
+/// Like [`Expr`]s but with sharing and deduplication.
+///
+/// Terms refer to their children indirectly as indexes into an ambient [`TermDag`].
 #[derive(Clone, PartialEq, Eq, Hash, Debug)]
 pub enum Term {
     Lit(Literal),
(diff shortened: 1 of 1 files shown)


wilcoxjay:
@oflatt please review my docs for accuracy before merge.

src/termdag.rs
```diff
@@ -71,6 +90,11 @@ impl TermDag {
         }
     }
 
+    /// Recursively converts the given expression to a term.
+    ///
+    /// This involves inserting every subexpression into this DAG. Because
+    /// TermDags are hashconsed, the resulting term is guarantee to maximally
```
oflatt, line 96: guaranteed

src/termdag.rs
```diff
@@ -4,15 +4,25 @@ use crate::{
     Symbol,
 };
 
+/// Like [`Expr`]s but with sharing and deduplication.
+///
+/// Terms refer to their children indirectly as indexes into an ambient [`TermDag`].
```
oflatt, line 9: Would be best if we could prevent users from constructing these but still allow pattern-matching on them. Not sure how to

oflatt (approved):
Thanks for the documentation!

PR #202, opened 2023-08-22
Make TermDag indices opaque, write tests for TermDag

wilcoxjay:
@oflatt mentioned he would prefer callers not to be able to construct Terms directly, but still be allowed to pattern match on them. I don't think exactly that is possible in Rust, but here is an alternative. We make the indices into the DAG opaque, but we leave the Term enum fully public. That way the API kinda pushes you toward only constructing one layer of a term at a time, which is exactly how it should be used. Oliver, what do you think?

This PR also adds a bunch of tests about terms (and a bit about exprs). If we decide not to make the term index change, I can split this PR to just merge the tests.

There is also a test that demonstrates how a caller can still shoot themselves in the foot by taking TermId from one TermDag and passing it to a different TermDag. This could cause wrong results or panic. I think this is not a dealbreaker and is a relatively easy failure mode to explain to callers.

### DIFF
--- a/src/ast/expr.rs
+++ b/src/ast/expr.rs
@@ -193,3 +193,24 @@ impl Display for Expr {
         write!(f, "{}", self.to_sexp())
     }
 }
+
+// currently only used for testing, but no reason it couldn't be used elsewhere later
+#[cfg(test)]
+pub(crate) fn parse_expr(s: &str) -> Result<Expr, lalrpop_util::ParseError<usize, String, String>> {
+    let parser = ast::parse::ExprParser::new();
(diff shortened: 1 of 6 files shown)


oflatt:
this is a definite improvement!
Another design would be to improve the `match_term_app` macro and force people to use that. Then we could not even ever hand back a term id, we would always expose a `Term`. I think that would be better but more complicated

Another tweak I thought about was for all the terms to reference the "ambient" term dag so we don't have to pass it around but this is also complicated

Anyways this is good we should merge

PR #205, opened 2023-08-23
Fix ExtractReport generation and extracting in sub-graph

saulshanabrook:
In the switch to using an extract action over an extract command in #176, the saving of the extracted expressions in a report was lost.

This PR restores that behavior and also refactors the `ExtractReport` to be an enum. In #176, the extraction logic was also changed, so that if passed in a non-zero number of variants, it would no longer perform a normal cost-based extraction as well. So now the report captures these two cases as distinct.

This PR also fixes printing the extracted expression in the `simplify` command, which was broken in https://github.com/egraphs-good/egglog/pull/195.


src/lib.rs
```diff
@@ -1109,31 +1116,6 @@ impl EGraph {
         }
     }
 
-    // Extract an expression from the current state, returning the cost, the extracted expression and some number
-    // of other variants, if variants is not zero.
-    pub fn extract_expr(&mut self, e: Expr, num_variants: usize) -> Result<ExtractReport, Error> {
```
saulshanabrook, line 1114: This function was previously used in a number of places before #176 and was now only being used in the un-tested Output command.

mwillsey:
Do we want the report to be an enum? I feel that the single and multiple extraction case could be handled uniformly, no?

saulshanabrook:
> Do we want the report to be an enum? I feel that the single and multiple extraction case could be handled uniformly, no?

The logic now for the extraction now seems disjoint between the variants in the best. If there are varients, then there is no "best" expression extracted, and vice versa:

https://github.com/egraphs-good/egglog/blob/9e530381961a59524f2bbacd89973575b4e036d8/src/typecheck.rs#L879-L909

Whereas previously, if you extracted some variants, a best was also extracted.

So I thought now an enum is appropriate?

saulshanabrook:
On another note, I noticed that `simplify` is not working with this new scheme, due to it running the extraction in a sub-egraph.

As a solution, I am considering changing from a saved "report" to a callback on the EGraph which is called during extraction.

Any thoughts on that?

mwillsey:
What simplify are we talking about? I’m not sure what the callback would do.

saulshanabrook:
> What simplify are we talking about?


The `simplify` command, i.e. like for this example:

```
(datatype Expr
  (Num i64))

(let x (Num 1))
(simplify (run) x)
```

> I’m not sure what the callback would do.

Instead of adding a report, it would call a callback on every extraction, so that from Python, we could still see what was extracted.

I opted not to do this, due to the added complexity, and instead copy reports from children to parents when popping.

oflatt (approved):
looks like a good change to me!

src/lib.rs
```diff
@@ -273,7 +279,18 @@ impl EGraph {
     pub fn pop(&mut self) -> Result<(), Error> {
         match self.egraphs.pop() {
             Some(e) => {
+                // Copy the reports and messages from the popped egraph
+                let extract_report = self.extract_report.clone();
+                let run_report = self.run_report.clone();
+                let messages = self.msgs.clone();
                 *self = e;
+                if let Some(report) = extract_report {
+                    self.extract_report = Some(report);
```
oflatt, line 288: Use `Option.or` instead of this if statement

oflatt (changes requested):
This looks like a good change, and fixes a bug in preserving messages across push/pop.
Had a small nit before we merge.

oflatt (approved):

PR #206, opened 2023-08-23
Fix to_sexp for extract

saulshanabrook:
In #176 the `ExtractCommand` is now created with `query-extract` instead of `extract`.


oflatt:
Good catch- could you rename `Extract` to `QueryExtract`?
I'm also realizing there's a big TODO where I'm just dropping `variants` altogether right now 💀

oflatt:
I'll fix the todo

saulshanabrook:
> Good catch- could you rename Extract to QueryExtract?


Sure, maybe in a follow up so as not to conflict with #208?

oflatt:
That would be great

oflatt (approved):

PR #207, opened 2023-08-23
Expose TermID

saulshanabrook:
Follow up from https://github.com/egraphs-good/egglog/pull/202 to expose the TermID so that we can create/access them in the Python bindings.

See #192 for the original PR which made these public.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,7 +18,7 @@ use index::ColumnIndex;
 use instant::{Duration, Instant};
 pub use serialize::SerializeConfig;
 use sort::*;
-pub use termdag::{Term, TermDag};
+pub use termdag::{Term, TermDag, TermId};
 use thiserror::Error;
 
 use proofs::ProofState;
--- a/src/termdag.rs
+++ b/src/termdag.rs
@@ -5,7 +5,7 @@ use crate::{
 };
 
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
-pub struct TermId(usize);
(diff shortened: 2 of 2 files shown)


wilcoxjay:
Whoops, good catch.

wilcoxjay:
If we need to expose the innards of TermId for python, then there's no point in having that type at all, and we should just go back to usizes. Sorry, I didn't really understand the constraints of they python integration when I refactored terms like this.

saulshanabrook:
> If we need to expose the innards of TermId for python, then there's no point in having that type at all, and we should just go back to usizes. Sorry, I didn't really understand the constraints of they python integration when I refactored terms like this.

The current Python integration is basically data in, data out. Every struct/enum which is passed into or returned from any of the needed egraph methods are exposed as Python classes (around new Rust structs). 

The only object which shares a reference between Python and Rust is the EGraph itself, everything else is copied.

So if we want to expose any struct methods to Python, we would have to convert the struct and the arguments first, then call it, and then convert the return value.

Of course, we could change any of that, but as a first pass in getting it up to date with the new changes, I was trying to do the minimal needed.

wilcoxjay:
Ok makes sense. In that case I think we should change TermIds to not have a wrapper struct, since the only reason I put the wrapper struct there was to make the field not public. So it should be 

```rust
type TermId = usize;
```

oflatt:
Sorry we broke this @saulshanabrook!

wilcoxjay:
LGTM!

oflatt (approved):

PR #213, opened 2023-08-24
Rename ExtractReport::{expr -> term}

wilcoxjay:
The old name was confusing because `Expr` refers to a different data structure
without sharing.

cc @oflatt


oflatt (approved):
nice catch

PR #214, opened 2023-08-24
Make match_term_app! support arbitrary patterns

wilcoxjay:
Now you can write arbitrary patterns in `match_term_app!`. I will use this in
eggcc as I refactor CFG conversion to use terms instead of expressions.

### DIFF
--- a/src/termdag.rs
+++ b/src/termdag.rs
@@ -34,19 +34,13 @@ pub struct TermDag {
 #[macro_export]
 macro_rules! match_term_app {
     ($e:expr; { $(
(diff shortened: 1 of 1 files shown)


wilcoxjay:
cc @oflatt

oflatt (approved):
Yay much better. Would be cool to be able to match literals but this is a good step

oflatt:
@wilcoxjay is on vacation so I'm planning on fixing the failing clippy on monday

wilcoxjay:
fixed, happy sunday from the airport!

oflatt:
Thanks! But you forgot to `cargo fmt` lol

PR #215, opened 2023-08-24
Fix desugaring foreign primitives

### LINKED ISSUE #212: Fix type checking custom primitives
It looks like in #176 there was a slight regression that breaks rules that use custom primitives that are not present in [the default `TypeInfo`](https://github.com/saulshanabrook/egg-smol/blob/47dba2bba6a426d0a554ce956eab238b2ff8d7bc/src/typechecking.rs#L35-L61).

This issue is blocking upgrading the Python bindings to recent commits (https://github.com/metadsl/egglog-python/pull/42) because we use a custom `PyObjectSort`.

The underlying problem introduced is that the desugaring pass checks (in `normalize_expr`) if a function is primitive by [creating a default `TypeInfo`](https://github.com/saulshanabrook/egg-smol/blob/47dba2bba6a426d0a554ce956eab238b2ff8d7bc/src/ast/desugar.rs#L128) and using that, instead of using the `TypeInfo` associated with the current graph:

I am not sure how to resolve this.

saulshanabrook:
Fixes the bug introduced by #176 which made foreign primitives desugar improperly. Closes #212.

It does this by moving the `TypeInfo` into the `Desugar` struct, as suggested by @oflatt.

I removed the `ProofState` struct, since it was just holding both the `TypeInfo` and `Desugar` together.


src/lib.rs
```diff
@@ -985,7 +983,7 @@ impl EGraph {
                         }
                         NormAction::LetLit(var, lit) => {
                             let value = self.eval_lit(lit);
-                            let etype = self.proof_state.type_info.infer_literal(lit);
+                            let etype = self.desugar.type_info.infer_literal(lit);
```
saulshanabrook, line 986: I thought about making a `type_info` method on the `EGraph`, to not have to go through `desugar` every time, but opted to leave it as is for a minimal change.

Happy to refactor to that though, if that is preferred.

oflatt: I like the type_info method idea!

oflatt (approved):
Good fix!

src/lib.rs
saulshanabrook: I gave that a go, refactoring all non-mutable usages of `egraph.desugar.type_info`.

Issue #196, opened 2023-08-20
Query atoms are order-sensitive


yihozhang:
The following rule fails to compile

```
(relation R (i64))

(rule 
  ((= x y)
   (= y 1)) 
  ((R x)))
(run 1)
(print-table Num)
```
with error message `'Unbound variable y in primitive computation'`. Note the same program compiles in Souffle.

This is a lower-priority bug.

In some sense, since we have already taken an "imperative" view of actions, it does not hurt to also take view triggers order-sensitively, despite potentially being unsatisfactory to Datalog-minded folks.

yihozhang:
Fixed by #320

Issue #133, opened 2023-04-19
Duplicate literals in query cause egglog to panic


yihozhang:
The following program panics.

```
(relation f (i64))

(rule ((= x 1)
       (= x 2))
      ((f x)))
      
(run 1)
(print f)
```

mwillsey:
What’s the desired behavior here? A warning? Type checking error?

yihozhang:
I think semantically this should just be a valid although unfirable rule. It should have the same semantics as the following rule (which also panics currently)
```
(relation f (i64))

(rule ((= 2 1))
      ((f 1)))
      
(run 1)
(print f)
```

A warning would be nice but not necessarily essential.

yihozhang:
Fixed by #219

Issue #80, opened 2022-12-14
Type Inference Failure


philzook58:
```
;ThCategory
(datatype TYPE)
(datatype TERM)
(function type (TERM) TYPE)
(function Ob () TYPE)
(function Hom (TERM TERM) TYPE)

(function id (TERM) TERM)
(rule ((= goofball (type (id A))))  ((let A_type (type A))))
(rewrite (type (id A)) (Hom A A) :when ((= (type A) Ob)))

(function compose (TERM TERM) TERM)
(rule ((= goofball (type (compose f g))))  ((let f_type (type f)) (let g_type (type g))))
(rewrite (type (compose f g)) (Hom A C) :when ((= (type f) (Hom A B)) (= (type g) (Hom B C))))


(birewrite (compose (compose f g) h) (compose f (compose g h)) :when ((= (type A) Ob) (= (type B) Ob) (= (type C) Ob) (= (type D) Ob) (= (type f) (Hom A B)) (= (type g) (Hom B C)) (= (type h) (Hom C D))))
(birewrite (compose f (id B)) f :when ((= (type A) Ob) (= (type B) Ob) (= (type f) (Hom A B))))
(birewrite (compose (id A) f) f :when ((= (type A) Ob) (= (type B) Ob) (= (type f) (Hom A B))))
```
https://www.mwillsey.com/egg-smol/?program=XQAAgAB9AwAAAAAAAAAdlQkEMyquloSsDef8Ne_CNtIKYi_rMgTflb9nB-DF-gk3Yjhk1l0lAFzyvSLtDQMTUWeekzTMgv3D9Ss2FhdWDmn3YGsm5y8dRgAwvMCVB0KoS2MxJfv7fkWXV5ZVvtbgJmePIO7P_q-JLbIg3vP3iv_lpP9EmmcBDPCHHJBFg_NH5CnfTYQKPnW33_xU20s6iKclWEzR6pcGY153-dZy7A9FRBPWSglI_kfJ5juBrwYIrb0G1_zySgYm8trnH2wL_ed1VGmlGPQi3Sz9dccLj95aiagKa6Lqu8eiBHDOH4l7wmZuMeZLVaqwSvs3QCpDPwmrDie94AM6kqOEilsJzHgqxv5Z_aY%253D

This fails with 
```
[ERROR] Errors:
    Failed to infer a type for: __rewrite_var
    Failed to infer a type for: f
```
on the last two rewrite rules. I think it ought to be able to infer the type of f from the :when clause.

PR #219, opened 2023-08-29
Constraint based type inference

yihozhang:
This PR proposes a constraint-based global type inference algorithm for egglog. More importantly, it proposes a core IR framework for writing rule analysis and transformation passes, which facilitates future development.

In terms of type inference:
* The type inference algorithm is inspired by Souffle and other Datalog compilers' design: it goes through each query and action atom, generates respective type constraints, and uses a simple fixpoint-based solver to solve type constraints. This design is declarative (one can inspect generated constraints) and scalable (one can add new type constraints to handle new features).
* This PR fixes a number of issues with the current type checker, e.g., #196, #133, #113, and potentially #80.
* The type inference algorithm supports _both_ polymorphism and ad hoc overloading: for example, it supports both polymorphic functions like `vec-add : vec<a> -> vec<a> -> vec<a>` and overloaded functions like `add`, which has several definitions such as `f64 -> f64 -> f64` and `String -> String -> String`.

In terms of the core IR:
* The old `typecheck.rs` is refactored into several orthogonal passes: 
  * A lowering pass that transforms a `Rule` into `CoreRule`,
  * A congruence pass that gets rid of redundant atoms, 
  * A canonicalize pass that removes equality check and "canonicalizes" variables,
  * A type checking pass.
  * Another lowering pass to `Program` (not currently implemented).
* A `CoreRule` has two parts: `Query` and `Actions`. `Query` uses a conjunctive query--based IR, while `Actions` uses an imperative SSA-based `NormAction`, which is currently used in desugaring. The goal is to gradually phase out `NormExpr` once `desugar.rs` is refactored into several compiler passes over core IR, so that we have a unified internal IR for query analysis and transformation.


yihozhang:
I had a discussion with @oflatt about the differences between the proposed conjunctive query-based IR and the existing `NormFact` IR. The goal of the proposed IR is to provide a flexible representation where we can develop a series of nano-passes and analyses that progressively transform the query into an efficient form.
However, this is lacking in `NormFact`.
To summarize my ideas, here are some advantages of the proposed IR in my opinion:

**It is declarative and unordered.**

The existing `NormFact` IR enforces an order between atoms, as is shown in #196. For example, consider the rule
```
(rule 
  ((= x y)
   (= y 1)) 
  ((R x)))
```
The query corresponds to `NormFact`
```
AssignVar x y
AssignLit y 1
```
which is not executable and causes egglog to panic.

In the proposed IR, it will first be lowered to
```
(rule 
  ((value-eq x y)
   (value-eq y 1)) 
  ((R x)))
```
After that, a canonicalization pass will substitute x with y in the rule, and then substitute y for 1:
```
(rule 
  () 
  ((R 1))
```

One way around this is to eliminate the ordered-ness of the NormFact and use ConstrainEqLit, etc. to model an equality check similar to `value-eq`. But even with this, NormFact is still not a good target for query optimizations. See the second point.

**The proposed IR is freer in the sense that it gives you a bigger space of expressible programs.**

`value-eq`, which is a primitive for equal checks, subsumes all of `AssignVar`, `AssignLit`, and `ConstrainEq` and is more general. Consider a simple rule

```
(rule 
  ((= x (Num 0))
   (= z (Mul x y)))
  ())
```

In `NormFact` it looks like this

```
(AssignLit __var 0)
(Assign x (Num __var))
(Assign z (Mul x y))
```

This indirection introduced by variable `__var` complicates query evaluation, since now without looking at the context we don't know if (Num __var) is a look-up (which can be further optimized) or scan. The point is that because of the lack of flexibility of NormAction, there is no way of inlining 0 in place.

Consider another example:

```
(rule 
  ((= (Num y) (Num x)))
  ())
```

In `NormFact` it looks like this

```
(Assign z1 (Num y))
(Assign z2 (Num x))
(ConstrainEq z1 z2)
```

Again, there is no way of expressing the same query efficiently in `NormFact`, because the IR disallows "inlining" `z1` and `z2`.
In contrast, in the proposed IR, the above NormFact looks like this:

```
(Num y z1)
(Num x z2)
(value-eq z1 z2)
```

A canonicalizing pass will substitute z1 with z2 and give you

```
(Num y z1)
(Num x z1)
```

which directly corresponds to an efficient join. 

Oliver had a good point that for proof generation, you may need to handle literals specially, so with `NormAction` one can assume all the arguments to functions in `Assign` is a variable, which can be handled uniformly. However, because the proposed IR is more expressive, it captures programs expressible in NormAction, so one can have a "de-canonicalization" pass that replaces every argument with a variable and generates a bunch of `(value-eq var val)` atoms. Then, after proof generation is done, another canonicalization pass can be used to eliminate these unnecessary `value-eq`s

**A uniform view of atoms is useful for query compilation.**

(This point is less relevant to NormFact IR but I think it's very cool.)

In the proposed IR, a query is just a vec of atoms, which abstracts over both functions and primitives. This uniform view is useful for further optimizations in query compilation. For example, consider

```
(rule 
  ((R x y z)
   (= (+ x y) z))
  ())
```

In the proposed IR, it corresponds to two atoms

```
(R x y z unit)
(+ x y z)
```

The only difference between `+` and `R` is that `+` enforces a functional dependency. One can therefore generate a query plan as
```
for x in R.x
	for y in R[x].y
		for z in Isect(R[x,y].z, +[x, y].z):
			output [x, y, z]
```
The current query compiler, however, will go through each z in R[x, y] and do the check afterward.

yihozhang:
Update: All the tests are passing! Some big TODOs left (maybe for future PR):

- [] #196, #133, #80: Since the errors are thrown by NormFact, we have to refactor or remove `NormFact` to fix this issue.
- [] Refactor type inference for actions: Currently this is done by reusing the existing action checker and is not integrated into the proposed framework.
- [] Refactor query compilation: The proposed IR does not split up atoms into primitives and functional atoms, while the query compilation and evaluation pass does. To be compatible with query compilation I introduced `func()` and `filters()` methods that simulate what was the `atoms` and `filters` fields. This is however inefficient. The principled way to do this is to also refactor query compilation to reflect this unified view.

oflatt:
As far as I can tell, this PR doesn't replace `typechecking.rs`- which means that there will still be two typecheckers if we merge this. Is that right?

yihozhang:
Is the type information from `typechecking.rs` used in other places? If this is true then you are right.

Are you okay with merging it while two typecheckers are present? If not, I'm happy to put more work into this PR, but this might make code review more complex than splitting into several PRs of smaller scopes.

oflatt:
Typechecking information is used by my terms and proofs encoding unfortunately. 
I think ideally, there would be just one type checker and it would happen just after desugaring but before the program reaches the back-end. In my mind sending a type-annotated AST to the back end would be best.

oflatt:
Merging with two typecheckers would raise the question- which one are we going to keep in the future?

oflatt:
Yeah, I understand it's quite difficult to avoid conflicts with my branch. 
However, I still believe that will some small tweaks we can merge it much sooner than after the eggcc project, which will take a lot of time.

Regardless, as you said this typechecker seems better and it makes the backend cleaner. I think your suggestion of keeping both typecheckers for now is a good compromise. I would just ask that you don't remove the old typechecker and that you do not remove NormCommand/NormFact.

yihozhang:
Update: This PR is self-contained and ready for review. The refactoring is not 100% done, but it is passing all the tests which is a milestone. So I propose to consider this PR as is for merging, and I plan to make follow-up PRs. More details:

* https://github.com/egraphs-good/egglog/pull/219#issuecomment-1701390640 is still true.
  * Compilation of actions is still done using the existing code, not the constraint-based type inference algorithm
  * The mentioned issues still exist- need to remove the front-end type check to do this
  * `gj.rs` and other places still use `query.funcs()` which is slow.
* Right now type error generated by the constraint system is a bit hacky. This is because constraint solving is implemented in a very general way (which is good) while the error information we need is very specific. I plan to refactor this by having a generic error type parameter in the constraint system.

TODOs for follow-up PRs (will open a tracking issue for them once this is merged).

* Add a renaming pass to resolve name conflicts/shadowing once and for all
* Improve error information for constraint solving.
* Add a congruence pass for optimization (so that `(= c1 (add a b)) (= c2 (add a b))` will be compiled into one atom)
* Refactor type inference for actions.
* Refactor the execution for actions to work over NormAction
* Use an opaque type instead of `AtomTerm` for `fn get_constraints` in the primitive function interface.

cc @mwillsey @oflatt

oflatt:
Are the new queries any slower now that we are not doing congruence?

yihozhang:
Oh, it’s congruence on query atoms not congruence in rebuilding’s sense.
This is just an easy and obvious optimization pass. If we are not
generating stupid queries like the one I mentioned there shouldn’t be a
performance issue.

I put it in TODOs because the old type checking code does this optimization.

src/constraint.rs
```diff
@@ -0,0 +1,399 @@
...
+    }
+}
+
+/// Construct a set of `Eq` constraints for the given arguments.
+/// If a sort is given, all arguments should also be equivalent to that sort.
+/// If a length is given, the arguments should have the exact length.
+/// If an output is given, the last element of arguments should have this output.
+/// useful for polymorphic and var-arg functions.
+/// Requires arguments.len() > 0
+pub fn all_equal_constraints(
```
oflatt, line 350: This helper function is a bit confusing because it does a lot at once.
What if we had a "contraint builder" system? It would have functions like `with_exact_length` and `with_all_arguments_type` and `with_output_type`.

src/gj.rs
```diff
@@ -247,7 +249,7 @@ impl<'b> Context<'b> {
                             let i = self.query.vars.get_index_of(v).unwrap();
                             self.tuple[i]
                         }
-                        AtomTerm::Value(val) => *val,
+                        AtomTerm::Literal(lit) => self.egraph.eval_lit(lit),
```
oflatt, line 252: isn't evaluating literals in the query a little slower and indirect than storing the value itself?

src/lib.rs
```diff
@@ -888,16 +909,22 @@ impl EGraph {
...
         let mut ctx = typecheck::Context::new(self);
-        let (query0, action0) = ctx
+        let core_rule = ctx
             .typecheck_query(&rule.body, &rule.head)
             .map_err(Error::TypeErrors)?;
-        let query = self.compile_gj_query(query0, &ctx.types);
+        let (query0, action0) = (core_rule.body, core_rule.head);
+
+        // TODO: We should refactor compile_actions later as well
+        let action0: Vec<Action> = action0.0.iter().map(|a| a.to_action()).collect();
```
oflatt, line 918: Here's where you can just preserve the NormAction

src/lib.rs
```diff
@@ -888,16 +909,22 @@ impl EGraph {
     ) -> Result<Symbol, Error> {
         let name = Symbol::from(name);
         let mut ctx = typecheck::Context::new(self);
-        let (query0, action0) = ctx
+        let core_rule = ctx
             .typecheck_query(&rule.body, &rule.head)
             .map_err(Error::TypeErrors)?;
-        let query = self.compile_gj_query(query0, &ctx.types);
+        let (query0, action0) = (core_rule.body, core_rule.head);
```
oflatt, line 915: please rename these variables

src/typecheck.rs
```diff
@@ -1,42 +1,109 @@
...
+// TODO: implement custom debug
+#[derive(Debug, Clone)]
+pub struct UnresolvedCoreRule {
+    pub body: Query<Symbol>,
+    pub head: Actions,
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct ResolvedCoreRule {
+    pub body: Query<ResolvedSymbol>,
```
oflatt, line 34: Since you are doing a bunch of refactoring, it would be awesome if the resolved query had type annotations. But that's future work

src/typecheck.rs
```diff
@@ -1,42 +1,109 @@
...
+    pub head: Actions,
+}
+
+#[derive(Debug, Clone)]
+pub enum ResolvedSymbol {
+    Func(Symbol),
+    Primitive(Primitive),
+}
+
+impl Query<ResolvedSymbol> {
```
oflatt, line 44: I think this can be generic over the parameter instead of specific to ResolvedSymbol?

src/typecheck.rs
```diff
@@ -48,20 +115,63 @@ impl<T: std::fmt::Display> std::fmt::Display for Atom<T> {
     }
 }
 
-#[derive(Default, Debug, Clone)]
-pub struct Query {
-    pub atoms: Vec<Atom<Symbol>>,
-    pub filters: Vec<Atom<Primitive>>,
+#[derive(Debug, Clone)]
+pub struct Query<T> {
+    pub atoms: Vec<Atom<T>>,
```
oflatt, line 120: What happened to filters? Wasn't it convenient to separate primitives from atoms?

src/typecheck.rs
```diff
@@ -109,337 +232,234 @@ impl PrimitiveLike for ValueEq {
...
-                    }
-                    Entry::Vacant(v) => {
-                        v.insert(Expr::Var(*var));
-                    }
-                },
-                _ => continue,
+    fn flatten_fact(&mut self, fact: &Fact) -> Query<Symbol> {
+        match fact {
+            Fact::Eq(exprs) => {
+                // TODO: currently we require exprs.len() to be 2 and Eq atom has the form (= e1 e2)
```
oflatt, line 278: I think we should make this requirement explicit in the long term, it just helps simplify the code

src/typecheck.rs
```diff
@@ -109,337 +232,234 @@ impl PrimitiveLike for ValueEq {
...
-                ENode::Var(var) => match leaves.entry(id) {
-                    Entry::Occupied(existing) => {
-                        canon.insert(*var, existing.get().clone());
-                    }
-                    Entry::Vacant(v) => {
-                        v.insert(Expr::Var(*var));
-                    }
-                },
-                _ => continue,
+    fn flatten_fact(&mut self, fact: &Fact) -> Query<Symbol> {
```
oflatt, line 275: Isn't this basically what desugaring already does? It seems like if you just use the existing `NormFact` you will be much closer to what you want.

src/typecheck.rs
```diff
@@ -109,337 +232,234 @@ impl PrimitiveLike for ValueEq {
...
-                                args.push(AtomTerm::Value(self.egraph.global_bindings[&v].1));
-                                continue;
-                            }
+    // TODO: should this be in ResolvedCoreRule
+    pub fn canonicalize(&self, rule: UnresolvedCoreRule) -> UnresolvedCoreRule {
+        let mut result_rule = rule;
+        loop {
+            let mut to_subst = None;
+            for atom in result_rule.body.atoms.iter() {
+                if atom.head == "value-eq".into() && atom.args[0] != atom.args[1] {
```
oflatt, line 330: I don't like this special case- instead of using the primitive `value-eq`, equality constraints should be specially represented in the data structure. For that reason, I have equality constraints in the `NormFact` data structure.

src/typecheck.rs
```diff
@@ -109,337 +232,234 @@ impl PrimitiveLike for ValueEq {
...
+    pub fn typecheck_query(
+        &mut self,
+        facts: &'a [Fact],
+        actions: &'a [Action],
+    ) -> Result<ResolvedCoreRule, Vec<TypeError>> {
+        let rule = Rule {
+            head: actions.to_vec(),
+            body: facts.to_vec(),
+        };
+        // dbg!(&rule);
```
oflatt, line 439: stray comments

oflatt (changes requested):
As far as I can tell, there are two big changes in here. (1) It adds a typechecker that is contraint-based. (2) It changes how query compilation works significantly- getting rid of the complex egraph stuff and just flattening the query.

In theory, I support both of these changes. But as they are implemented in this PR, I don't think it's worth merging. 

(1) The contraint-based typechecker is good, but happens too late to matter. I'm starting to think there is no point in merging it now and forcing you to replace the current typechecker with it. I think it will be easier for both of us if you replace the existing `typechecking.rs` with this as a separate PR.

(2) I like that this simplifies and improves query compilation, but it seems to re-invent the wheel. Most of what query compilation in this PR does is flatten things into atoms, which is exactly what desugaring already does. In fact, if you reuse the `NormFact` struct it basically already captures what you are doing here.

Again, great work on this PR! I think with more work this will be a big improvement to egglog. 
Let me know what you think of my suggestions

src/constraint.rs
yihozhang: That's a great idea! I've made the requested refactoring, which also makes the interface look nicer.

yihozhang (commented):
Thanks for the great review @oflatt and sorry for getting back late. Besides addressing some of your comments, I also made the following changes according to our discussion:

* Typechecking (for queries at least) happens only once now (in `typechecking.rs`), the backend CoreRule directly uses the type-checking result from it instead of redoing it.
* The backend takes a `NormRule` instead of a pair of `Query` and `Action` now, so we don't need to convert a `NormRule` first to an un-desugared `Rule` and then to `CoreRule`.
* Many codes are removed: We no longer need the old type-checking algorithm, we don't have to convert from un-desugared `Rule` to `CoreRule`, etc.

I am very happy about all of these changes

src/constraint.rs
```diff
@@ -0,0 +1,399 @@
...
+    Eq(Var, Var),
+    Assign(Var, Value),
+    And(Vec<Constraint<Var, Value>>),
+    // Exactly one of the constraints holds
+    // and all others are false
+    Xor(Vec<Constraint<Var, Value>>),
+    Impossible(ImpossibleConstraint),
+}
+
+pub enum ConstraintError<Var, Value> {
```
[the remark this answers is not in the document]
oflatt: Sure

oflatt:
Before I do my review, I just want to say amazing work @yihozhang and we make a great team! This is a big moment for egglog.

src/typecheck.rs
```diff
@@ -1,42 +1,109 @@
...
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct ResolvedCoreRule {
+    pub body: Query<ResolvedSymbol>,
+    pub head: Actions,
+}
+
+#[derive(Debug, Clone)]
+pub enum ResolvedSymbol {
```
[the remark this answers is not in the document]
oflatt: `Symbol` is used in a lot of places, where here you really mean a function call or primitive. I still think we should change it.

Is this supposed to include type information in the future?

src/typecheck.rs
```diff
@@ -1,42 +1,215 @@
...
+    Func(Symbol),
+    Primitive(Primitive),
```
oflatt, line 37: I'm still of the opinion that equality is special and should not be encoded as the `value-eq` primitive. It doesn't make sense because we never actually call `value-eq`, we compile it into equality checks. This will be confusing for new people in the code base.

src/typecheck.rs
```diff
@@ -1,42 +1,215 @@
...
+            atom.subst(subst);
+        }
+        self.head.subst(subst);
+    }
+}
+
+impl UnresolvedCoreRule {
+    // keep it here while it's untested
+    #![allow(unused)]
+    pub(crate) fn to_norm_rule(&self, egraph: &EGraph) -> NormRule {
```
oflatt, line 74: Why do we need this? Is it dead code?

src/typecheck.rs
```diff
@@ -1,42 +1,215 @@
...
+            }),
+        }
+    }
+    Query { atoms }
+}
+
+impl NormRule {
+    // keep it here while it's untested
+    #![allow(unused)]
+    pub(crate) fn to_core_rule(&self, typeinfo: &TypeInfo) -> UnresolvedCoreRule {
```
oflatt, line 173: Why is it unused? I thought the current pipeline required it

src/typecheck.rs
oflatt: Still think we shouldn't have `value-eq`, see my comment above.

src/typecheck.rs
```diff
@@ -48,20 +221,63 @@ impl<T: std::fmt::Display> std::fmt::Display for Atom<T> {
...
+
+impl<T> Default for Query<T> {
+    fn default() -> Self {
+        Self {
+            atoms: Default::default(),
+        }
+    }
+}
+
+impl Query<Symbol> {
```
oflatt, line 237: Can we rename `typecheck.rs` to something like `query-compile.rs`?
We should also move all the typechecking related stuff like this into `typechecking.rs`

src/typecheck.rs
```diff
@@ -78,368 +294,166 @@ impl<T> Atom<T> {
...
-                    if self.egraph.global_bindings.get(var).is_some() {
-                        match leaves.entry(id) {
-                            Entry::Occupied(existing) => {
-                                canon.insert(*var, existing.get().clone());
-                            }
-                            Entry::Vacant(v) => {
-                                v.insert(Expr::Var(*var));
-                            }
+    // TODO: should this be in ResolvedCoreRule
+    pub fn canonicalize(&self, rule: UnresolvedCoreRule) -> UnresolvedCoreRule {
```
oflatt, line 349: Another place where it would be good to add documentation- what's the purpose of canonicalize?

src/typecheck.rs
```diff
@@ -78,368 +294,166 @@ impl<T> Atom<T> {
...
-        while keep_going {
-            keep_going = false;
-            let nodes = std::mem::take(&mut self.nodes);
-            for (mut node, id) in nodes {
-                // canonicalize
-                let id = self.unionfind.find(id);
-                if let ENode::Func(_, children) | ENode::Prim(_, children) = &mut node {
-                    for child in children {
-                        *child = self.unionfind.find(*child);
+    pub(crate) fn resolve_rule(
```
oflatt, line 377: More typechecking stuff that probably should be in `typechecking.rs`

src/typechecking.rs
```diff
@@ -259,8 +266,25 @@ impl TypeInfo {
     }
 
     fn typecheck_facts(&mut self, ctx: CommandId, facts: &Vec<NormFact>) -> Result<(), TypeError> {
-        for fact in facts {
-            self.typecheck_fact(ctx, fact)?;
+        // ROUND TRIP TO CORE RULE AND BACK
```
oflatt, line 269: Good place for a TODO

oflatt (changes requested):
Did an initial review, plan to make another pass after you respond.
Looks like great progress.

src/typecheck.rs
yihozhang: I am convinced. After thinking about it, I introduced another rule variant and the current core definition looks like this:

```
#[derive(Debug, Clone)]
pub struct UnresolvedCoreRule {
    pub body: Query<SymbolOrEq>,
    pub head: Actions,
}

#[derive(Debug, Clone)]
pub struct CanonicalizedCoreRule {
    pub body: Query<Symbol>,
    pub head: Actions,
}
```

And `canonicalize` transforms an `UnresolvedCoreRule` into a `CanonicalizedCoreRule`. This makes equalities explicit in our IR definition.

src/typecheck.rs
yihozhang: I removed it.

src/typecheck.rs
yihozhang: Good catch. It is indeed used in our pipeline. I removed the `allow(unused)` annotation.

src/typecheck.rs
yihozhang: I made an issue for this (#268). I'll do it after the PR is merged but not now to make the diff clean.

src/typecheck.rs
yihozhang: Good point. I have documented its purpose.
```

    /// Transformed a UnresolvedCoreRule into a CanonicalizedCoreRule.
    /// In particular, it removes equality checks between variables and
    /// other arguments, and turns equality checks between non-variable arguments
    /// into a primitive equality check `value-eq`.
```

yihozhang: FWIW, `value-eq` cannot be completely removed from our codebase since queries may contain equality checks between e.g., global variables, which cannot be compiled away.

src/typechecking.rs
yihozhang: Done

src/typecheck.rs
oflatt: Great!

src/constraint.rs
```diff
@@ -0,0 +1,442 @@
...
+
+                            if updated {
+                                swap(&mut result_assignment, assignment);
+                            }
+                            assignment_updated = updated;
+                        }
+                        Err(error) => errors.push(error),
+                    }
+                }
+                // If update is successful for only one sub constraint, then we have nailed down the only true constraint.
```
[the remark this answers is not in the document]
oflatt: I see, thanks

oflatt (approved):
Great work Yihong!

PR #220, opened 2023-08-29
Update link to zulip chat

saulshanabrook:
There is now a full egglog stream and the egg-smol topic is not used anymore.


oflatt (approved):
Thanks!

PR #223, opened 2023-09-03
Distinguish datatypes from other tables

oflatt:
Tables with merge functions are fundamentally different than those without- we have places in the code where we check for a merge function or defaults. This PR clarifies that distinction.

It also enforces that the `set` keyword is only used on non-datatype tables, clarifying the confusion between `set` and `union` and making it clear when `set` is needed.

Depends on #222


src/ast/mod.rs
```diff
@@ -335,7 +335,29 @@ pub enum Command {
         sort: Symbol,
     },
     Sort(Symbol, Option<(Symbol, Vec<Expr>)>),
+    /// Declare an egglog function.
+    /// The function is a datatype when:
+    /// - The output is not a primitive
+    /// - No merge function is provided
+    /// - No default is provided
     Function(FunctionDecl),
```
saulshanabrook, line 343: Oh I wasn't aware you could have functions with non primitive return values and custom merge functions...

oflatt: It's currently a very under-tested feature! I plan to use it more in the future though, it's essential for encoding proofs.

saulshanabrook:
I am giving this branch a test with the Python test suite to see if anything breaks and will let you know this week!

oflatt:
> I am giving this branch a test with the Python test suite to see if anything breaks and will let you know this week!

Thanks! That's super helpful

src/ast/mod.rs
yihozhang: It can be used to simulate the [`choice`](https://souffle-lang.github.io/choice) construct in Datalog literature by setting `:merge old` (as is used in proofs)

yihozhang:
Does this PR disallow `:merge old` for datatypes?

oflatt:
Yep, datatypes can't have merge functions

yihozhang:
Doesn't this cause issues for proofs? Proof terms are datatypes, right?

I like the ability to simulate the choice construct (for data types) in Datalog. I'm even thinking about having a `:choice` keyword in function declarations. Or maybe you are sure we don't need this anymore?

saulshanabrook:
I gave it a go on the Python bindings, and almost all the tests pass. The one that doesn't might be on my end, so I'll keep looking into it. It's a more complicated example that I am looking to refactor anyway a bit

https://github.com/metadsl/egglog-python/pull/46

src/ast/mod.rs
```diff
@@ -559,7 +616,7 @@ impl FunctionDecl {
             },
             merge: None,
             merge_action: vec![],
-            default: None,
+            default: Some(Expr::Lit(Literal::Unit)),
```
saulshanabrook, line 619: I am curious why the relation now needs a default to work? It does make it a little more cumbersome to define relations with the function syntax. Would you ever want to define a function that returns `Unit` but doesn't have this as a default?

oflatt:
> Doesn't this cause issues for proofs? Proof terms are datatypes, right?
> 
> I like the ability to simulate the choice construct (for data types) in Datalog. I'm even thinking about having a `:choice` keyword in function declarations. Or maybe you are sure we don't need this anymore?

Actually, this PR supports proofs. For proofs we need a way to talk about stable ids for terms, so we need some kind of table which is stable (I call those datatypes). If you want something else, just use a table instead.

src/ast/mod.rs
oflatt: Good question- 
`Unit` is a primitive just like `i64`. Before, we had special logic for making default units, making them special in a weird way. Now, I make it explicit that the `relation` syntax is sugar for this function for convenience.

You can still have a function that returns `Unit` and use `set` everywhere.

saulshanabrook: Ah I see.

Is it true then that using a function that does not return a default as an action is a no-op? And that's why you needed the special case before and now with it removed you have added a default?

As a test, this rule seems to have no effect, when there is no default (based on the viz):

```
(datatype Expr)

(function trigger () Expr)

(function foo () i64)

(trigger)

(rule ((= x (trigger))) ((foo)))

(run 1)
```

But it does when there is a default:


```
(datatype Expr)

(function trigger () Expr)

(function foo () i64 : default 10)

(trigger)

(rule ((= x (trigger))) ((foo)))

(run 1)
```

yihozhang (approved):
Thanks! This makes sense now. I left a small comment but looks great!

src/ast/mod.rs
oflatt: Hmm, shouldn't you get an exception when there isn't a default?
Otherwise, the action might fail halfway-through!

src/typecheck.rs
```diff
@@ -747,9 +752,6 @@ impl EGraph {
                     let value = if let Some(out) = function.nodes.get(values) {
                         out.value
                     } else if make_defaults {
-                        if function.merge.on_merge.is_some() {
```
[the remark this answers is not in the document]
oflatt: `make_defaults` implies that this isn't a datatype. Either there is a merge function, or you get an exception because you tried to merge two primitives.

src/ast/mod.rs
saulshanabrook: Yeah, so if I am understanding that correctly, only functions with defaults can be used as actions?

PR #225, opened 2023-09-06
Better rule and ruleset timing statistics

oflatt:
This PR improves the run report to report timing information for individual rules and also for rulesets.
It also adds, in addition to the existing run report for the most recently run schedule, a report that collects statistics across all the runs.


saulshanabrook (approved):
Seems reasonable without testing anything out!

yihozhang (approved):

PR #226, opened 2023-09-06
Add a replace action

oflatt:
When writing rules where one row in a table subsumes another, it's often hard to write the rules. This PR adds a prototype for a `replace` action to solve this problem.

In particular, the only way to do it right now is to use `delete`. However, if the two rows are actually the same, the problem is that we never saturate.
Stupid example that runs forever:
[https://egraphs-good.github.io/egglog/?program=XQAAgACeAAAAAAAAAAAFYJzIYToE-74urWhpV-IA_[…]IY2FIeFaKKvnanDvE9ubGR2gphrxqDRu1WAeyBGxkXMQP_-wXIAA%253D](https://egraphs-good.github.io/egglog/?program=XQAAgACeAAAAAAAAAAAFYJzIYToE-74urWhpV-IA_FDhGu3fewDZx-wvrHQ77dBuxcNtTt8GO8U3aYj9B8GU5s5_6Xs03CIMEusIY2FIeFaKKvnanDvE9ubGR2gphrxqDRu1WAeyBGxkXMQP_-wXIAA%253D)

This PR fixes this problem by adding a `replace` action, allowing for one row to replace another (and does nothing when they are the same).

### DIFF
--- a/src/ast/desugar.rs
+++ b/src/ast/desugar.rs
@@ -411,6 +411,17 @@ fn add_semi_naive_rule(desugar: &mut Desugar, rule: Rule) -> Option<Rule> {
                     new_head_atoms.push(Fact::Eq(vec![fresh_var, expr]));
                 };
             }
+            Action::Replace { new_output, .. } => {
+                var_set.extend(new_output.vars());
+                if let Expr::Call(_, _) = new_output {
+                    add_new_rule = true;
+
+                    let fresh_symbol = desugar.get_fresh();
+                    let fresh_var = Expr::Var(fresh_symbol);
+                    let new_output = std::mem::replace(new_output, fresh_var.clone());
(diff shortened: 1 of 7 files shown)


oflatt:
Maybe we should get rid of `delete` in favor of `replace`?

yihozhang:
Since replace desugars to a `set` in actions, do you need to also duplicate it in `query` to make semi-naive sound?

oflatt:
> Since replace desugars to a `set` in actions, do you need to also duplicate it in `query` to make semi-naive sound?

Yes, great catch! Fixed.

oflatt:
To be clear, we shouldn't merge this yet- still playing around with it.

mwillsey:
Let’s keep not-ready prs as drafts, it helps reduce notification noise a bit

oflatt:
No longer necessary for terms- could still be a good idea but closing for now.

PR #227, opened 2023-09-08
Small gj refactor

oflatt:
This PR breaks out the core loop of seminaive generic join into a function

### DIFF
--- a/src/gj.rs
+++ b/src/gj.rs
@@ -403,6 +403,8 @@ impl EGraph {
         self.make_trie_access_for_column(atom, column, timestamp_range)
     }
 
+    // Returns `None` when no program is needed,
+    // for example when there is nothing in one of the tables.
     fn compile_program(
         &self,
         query: &CompiledQuery,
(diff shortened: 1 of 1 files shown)


src/gj.rs
```diff
@@ -620,6 +620,78 @@ impl EGraph {
...
+        mut f: F,
+    ) where
+        F: FnMut(&[Value]) -> Result,
+    {
+        let atom = &cq.query.atoms[atom_i];
+        // do the gj
+        if let Some((mut ctx, program, cols)) = Context::new(self, cq, timestamp_ranges) {
+            let start = Instant::now();
+            log::debug!(
+                        "Query:\n{q}\nNew atom: {atom}\nTuple: {tuple}\nJoin order: {order}\nProgram\n{program}",
```
yihozhang, line 637: nits: `atom`/`atom_i` only makes sense for semi-naive. Moreover, if not doing semi-naive, `New atom: ...` does not make sense. It would be great if you could do a favor by also refactoring this part.

oflatt: done!

yihozhang (approved):

PR #230, opened 2023-09-09
Add count match and replace string functions

saulshanabrook:
This PR adds two new methods for strings, counting the number of occurrences of a substring and replacing a substring.


oflatt (approved):
Looks good to me, won't merge yet so other egglog people can take a look if they want to.

Issue #228, opened 2023-09-09
RFC: fix semi-naive and table problems


oflatt:
Right now, we suffer from several related awkward problems.
The biggest is #142, where users of egglog can write rules where things in the right-hand side of a rule might change.

However, there's another problem. In this example, we see that users can write rules that access tables with no default, causing a rule execute only half of it's action:
https://egraphs-good.github.io/egglog/?program=XQAAgADQAAAAAAAAAAAUGQgnfMUD9dO10C6gYDgH7EMCDsP59lqRF2etuT_we0rC4-zdrRMdtQlgkwnRHou9aMqzrfUygj6rjTtUT_55lzsMcxqoclfQM7aYX12UDnF9ACgHhH7nPTjYxz3Y8rMOJ0YuOAqy89w9KfjT-1l2WN_734IA

I propose a simple fix: allow creating new *datatypes* in the actions of the rules. Also allow using `set` to set tables using these datatypes. However, do not allow looking up a value from a table.

See #223 for the difference between datatypes and tables

oflatt:
For further motivation, consider that these "look-ups" in tables on the right hand side of rules must be justified, complicating the proof system significantly.

Alex-Fischman:
I expected to get an error saying that "there's no value for (Num 1) in mayfail", and stopping execution when I ran that example. Why is this not what happened?

oflatt:
Hmm is that what happens on my PR? Either way it's bad

Alex-Fischman:
I don't know what happens on your PR, I'm saying that this is the behavior I expect as a user.

saulshanabrook:
xposting the original discussion I think that led to this issue: https://github.com/egraphs-good/egglog/pull/223#discussion_r1320605325

Thanks for opening a dedicated issue for it!

Alex-Fischman:
Oliver, wasn't this also your first thought in the other discussion? The correct thing to do, if the user does a mutable lookup on a table with no default action, and there is no output for the inputs they gave, is to throw an error. Also, we can't just restrict users to only put non-datatype lookups in the query, because of the :default option.

oflatt:
I think throwing an error is certainly better than what we do now.
However we could just prevent this problem with this proposal. I also realized, as you said, that this proposal necessitates getting rid of the :default option!

I think that if you want a default you can always use a helper rule.

Another idea would be to change the semantics of default so that it takes the default value in queries without being in the database if that makes any sense.

[...]

PR #234, opened 2023-09-13
Panic instead of partially running actions

Alex-Fischman:
This PR changes the ignores in lib.rs to panics, as a stopgap against half-run rules. It also fixes the resulting failing tests by doing the demand transformation manually.


Alex-Fischman:
This resolves the most urgent concern from #228, which is that rules can fail silently.

mwillsey:
Looks good. Can you add an explicit panic message?

Alex-Fischman:
Added.

Issue #229, opened 2023-09-09
Rule Applied Twice Bug


saulshanabrook:
I am encountering a bug where the actions for the actions in a rule will trigger twice in one execution. Normally, this is fine, if the rules are monotonic. But if you have a function that returns a primitive and has a merge function, then setting it twice could lead to incorrect results.

## Reproducible use case

Here is a way to reproduce this bug. It's not super minimal, but it is a lot more minimal than the Python code I extracted it from.

What it does is turns an expression graph into a string of statements of each expression. 

It works bottom-up, generating a string for each expression and also adding to the statements, if the expression defines a new variable, and incrementing the gensym.

It basically works, but it seems like the actions in the multiply rule are run twice on the same nodes instead of just once.

```lisp
(function statements () String :merge (+ old new) :default "")
(function gensym () i64 :merge (+ old new) :default 0)
(sort Math)
(function Num (i64) Math)
(function Var (String) Math)
(function Add (Math Math) Math)
(function Mul (Math Math) Math)
(function expr (Math) String)
(rule ((= x (Var s)))
      ((set (expr x) s))
         )
(rule ((= x (Num i)))
      ((set (expr x) (to-string i)))
         )
(rule ((= x (Add y z))
       (= y_expr (expr y))
       (= z_expr (expr z)))
      ((set (expr x) (+ "_" (to-string (gensym))))
       (set (statements) (+ (+ "_" (to-string (gensym))) " = " y_expr " + " z_expr "
"))
       (set (gensym) 1))
         )
(rule ((= x (Mul y z))
       (= y_expr (expr y))
       (= z_expr (expr z)))
      ((set (expr x) (+ "_" (to-string (gensym))))
       (set (statements) (+ (+ "_" (to-string (gensym))) " = " y_expr " * " z_expr "
"))
       (set (gensym) 1))
         )
(Mul (Num 2) (Add (Var "x") (Num 3)))

(run 3)

(extract (statements))
(extract (expr (Mul (Num 2) (Add (Var "x") (Num 3)))))
```

It should print out:

```
"_0 = x + 3
_1 = 2 * _0
"
"_1"
```

But it duplicates the second statement instead:

```
"_0 = x + 3
_1 = 2 * _0
_1 = 2 * _0
"
"_1"
```

saulshanabrook:
I am starting to debug this issue.

I added a few `println`'s to see try to pinpoint what's going on. Here is a bit of the trace that I think is thee problem area:

```
[...]

PR #238, opened 2023-09-28
Clarify that merge should be monotonic

saulshanabrook:
Resolves the question in #229 by stating that merge functions must be monotonic, otherwise there is undefined behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,8 @@ If you define a `:merge` expression, you can update specific values in the funct
 (extract (KeepMax 1)) ; this is 2
 ```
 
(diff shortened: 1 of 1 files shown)


oflatt (changes requested):
This is an improvement, but it doesn't really say what happens if you break this rule. Maybe we just say that it's undefined behavior or you will get unexpected results due to rules firing multiple times?

We really need proper documentation (I'll be working on that soon) but this is still better than nothing!

oflatt (approved):
Thanks!

PR #239, opened 2023-09-28
Add `vec-set`

rtjoa:
Add primitive `(vec-set vec index val)`

### Testing
Added basic test to [`tests/vec.egg`](https://github.com/egraphs-good/egglog/blob/cec39afc8100a167cc9e8326825c229db2485a8c/tests/vec.egg)


oflatt (approved):
LGTM

PR #242, opened 2023-10-01
Fix `query-extract` for `vec`

hatoo:
Currently, `query-extract` for `vec` prints the reverse order of the argument

```
(sort IVec (Vec i64))

(let e (vec-of 1 2 3 4))
(query-extract e) ; => (vec-push (vec-push (vec-push (vec-push (vec-empty) 4) 3) 2) 1)

(check (= e (vec-push (vec-push (vec-push (vec-push (vec-empty) 4) 3) 2) 1))) ; fail
```

This PR fixes it and prints by `vec-of`

```
(sort IVec (Vec i64))

(let e (vec-of 1 2 3 4))
(query-extract e) ; => (vec-of 1 2 3 4)
(query-extract (vec-empty)) ; => (vec-empty)

(check (= e (vec-of 1 2 3 4))) ; OK
```

### DIFF
--- a/src/sort/vec.rs
+++ b/src/sort/vec.rs
@@ -468,3 +468,35 @@ impl PrimitiveLike for Set {
         vec.store(&self.vec)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_vec_make_expr() {
+        let mut egraph = EGraph::default();
+        let outputs = egraph
+            .parse_and_run_program(
+                r#"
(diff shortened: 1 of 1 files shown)


oflatt (approved):
Good catch!

oflatt (changes requested):
Great catch, could you please also add a repro test for this?

hatoo:
Added.
I'm not sure it's the best way to test this though.

oflatt (approved):
Thanks!

PR #245, opened 2023-10-01
Add option serialization option to separate primitive results into their own e-classes

### LINKED ISSUE #244: Add option serialization option to separate primitive results into their own e-classes
I am opening this issue because I would like to add an option to the current serialization route to separate out all primitive results into their own e-classes instead of using own e-class for all values which return or take that primitive.

A motivating example is this large visualization from some of my Python work:

Source gv

If you zoom in, you can see a few places where this change would improve the readability. 

1. There are many `Program_statements` which are all in the same e-class as the empty string `""`. If those were all in separate e-classes, each with their own `""`, it would result in more nodes, but the placement could be closer to the nodes they are coming from, so the length and overlap of the lines could be minimized.
2. All of the relations `Program_compile` are in one e-class (all return `Unit`), which makes it very hard to follow which programs have been compiled yet and which haven't. If each were in their own e-class, then they could be laid out closer to the nodes they refer to.
3. Many of the `Program_compile` nodes take `0` as an argument, but `0` is also the value of many `Program_next_sym`. Since it's not in it's own e-class, it can't be inlined into the `Program_compile`, like it would if it was in its own. This would reduce the number of edges greatly.

## Proposal

I plan on working on a PR to add a boolean option to the serialization config, with the default preserving existing behavior. If it is switched, all primitive nodes which are returned from function will be in their own e-classes, with just the function and primitive node, instead of being grouped together with all other versions of that value. 

## Questions

Q: Instead of adding this as an option, would you want to just change the behavior to this behavior exclusively?
A: I thought for now the least invasive change would be to add a config, since there may be cases where the current behavior is desirable.  For extraction purposes, this will produce a greater number of nodes, so might not be helpful.

Q: Could you instead implement this in the upstream [`egraph-serialize`](https://github.com/egraphs-good/egraph-serialize) package, where we have already added algorithms to shift around the serialized form for better visualization?
A: I had hoped to do this, to make the functionality more general purpose. However, to split an e-class like this into many requires knowing which node out of the e-class is the actual primitive value, and which are functions. We could assume that the node without any children is the primitive value, but this isn't true for collection primitives, and also function which take no args but return a primitive would also be confused. So I couldn't think of a way to distinguish it, once all e-classes are serialized, without trying to parse the string (bad) or adding additional metatada to the JSON file, which seemed more confusing than just implementing it downstream in this library.

saulshanabrook:
Fixes #244

## Example from Python repo:

You can see that in this example, splitting the primitive outputs makes the structure of the graph **much** clearer and the edges are legible, even though it is just as large:

### Original

Source gv

### Split outputs

Source gv

## TODO:

- [x] Try on my particular example to see if it fixes it
- [x] Create PR to serialize-egraph repo with new JSON files produced by this visualized https://github.com/egraphs-good/egraph-serialize/pull/8

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1355,8 +1355,15 @@ impl EGraph {
     }
 
     /// Serializes the egraph for export to graphviz.
-    pub fn serialize_for_graphviz(&self) -> egraph_serialize::EGraph {
-        let mut serialized = self.serialize(SerializeConfig::default());
+    pub fn serialize_for_graphviz(
+        &self,
+        split_primitive_outputs: bool,
+    ) -> egraph_serialize::EGraph {
+        let config = SerializeConfig {
+            split_primitive_outputs,
+            ..Default::default()
+        };
+        let mut serialized = self.serialize(config);
         serialized.inline_leaves();
         serialized
     }
--- a/src/main.rs
+++ b/src/main.rs
@@ -164,7 +164,7 @@ fn main() {
             }
         }
         // if we are splitting primitive outputs, add `-split` to the end of the file name
-        let serialiaze_filename = if args.serialize_split_primitive_outputs {
+        let serialize_filename = if args.serialize_split_primitive_outputs {
             input.with_file_name(format!(
                 "{}-split",
                 input.file_stem().unwrap().to_str().unwrap()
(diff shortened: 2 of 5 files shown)


oflatt (approved):
Looks like a good option to me!

PR #247, opened 2023-10-04
Update Fact accept CallExpr instead of Expr

### LINKED ISSUE #243: The `:when` syntax is easy to make mistakes
When someone writes `:when` clause wrongly, such like

```
(rewrite
...
  :when (> x y) ; What I want is :when ((> x y))
)
```

It is syntactically correct but doesn't make sense because as far as I know there is no way to write a falsy variable or literal to code so it will always be true.  (> x y) => (true true true) => true

I think `egglog` should only accept `CallExpr` instead of `Expr` as a `Fact` in `:when` clause.

hatoo:
Fix #243 

This PR

- Fix `Command::Extract` to contain `Expr` instead of `Fact` because inputting `Fact::Eq` into it leads to a runtime error
- Disallow variable and literal itself as a fact
  - Since `Expr::Call` doesn't have its own type yet, `Fact` still has an `Expr`.

### DIFF
--- a/src/ast/desugar.rs
+++ b/src/ast/desugar.rs
@@ -453,7 +453,7 @@ pub(crate) fn desugar_simplify(
         desugar_command(
             Command::Extract {
                 variants: 0,
-                fact: Fact::Fact(Expr::Var(lhs)),
+                expr: Expr::Var(lhs),
             },
             desugar,
             false,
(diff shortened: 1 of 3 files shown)


oflatt (approved):
LGTM, keeping #243 open because it doesn't actually change the fact data structure to fix this issue.

PR #248, opened 2023-10-04
Omit default values when serializing primitive outputs

saulshanabrook:
This PR changes the serialization logic to omit outputting a separate node for the the primitive outputs of functions, if that return value is equal to the default.

This removes the extraneous `()` (Unit) nodes from the serialized e-graph as well as any other nodes equal to their defaults, which helps reduce the size of visualizations and makes them more readable.

### DIFF
--- a/src/serialize.rs
+++ b/src/serialize.rs
@@ -2,7 +2,7 @@ use ordered_float::NotNan;
 use std::collections::VecDeque;
 
 use crate::{
-    ast::{FunctionDecl, Id},
+    ast::{Expr, FunctionDecl, Id},
     function::{table::hash_values, ValueVec},
     util::HashMap,
     EGraph, Value,
@@ -52,7 +52,8 @@ impl EGraph {
     ///
     /// This is to achieve the following properties:
     /// - Equivalent primitive values will show up once in the e-graph.
-    /// - Functions which return primitive values will be added to the e-class of that value.
+    /// - Functions which return primitive values will be added to the e-class of that value (if `split_primitive_returns` is false, otherwise they will have a separate e-class)
+    ///   - If the value is equal to its default, it will not be added to the e-graph
     /// - Nodes will have consistant IDs throughout execution of e-graph (used for animating changes in the visualization)
     /// - Edges in the visualization will be well distributed (used for animating changes in the visualization)
     ///   (Note that this will be changed in `<https://github.com/egraphs-good/egglog/pull/158>` so that edges point to exact nodes instead of looking up the e-class)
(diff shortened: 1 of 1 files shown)


oflatt:
What happens to default values? We still get a node in the egraph for the row but it just doesn't point to anything for that value?

saulshanabrook:
We don't get a node in the e-graph for the output, we only get one for the function itself, for example:

```
(datatype Math
  (Var String))

(function Min (Math) i64 :default 0)
(function IsOdd (Math) Unit :default ())

(Min (Var "x"))
; same as (set (Min (Var "x")) 0)
(set (Min (Var "y")) 10)

(IsOdd (Var "x"))
``` 
eqsat-basic-split

oflatt:
I find that a bit confusing, especially for a new egglog user.
What about putting them in-line with the constructor like I've seen you do before?

saulshanabrook:
I could add it as an option, with the default as false, instead of on by default. It's a way to increase information density, omitting non-essential stuff, but does make things less explicit. 

@mwillsey [suggested](https://github.com/egraphs-good/egglog/pull/147#issuecomment-1646041649) at least omitting `()`  nodes for relations

oflatt:
I would say omitting the `()` is much less confusing, so I'd be a proponent of that change
But shouldn't that be something we change on the visualization end? I'd like them to still show up in the data for clarity

saulshanabrook:
I could instead add an `is_default` metadata to the serialized data and then add a way to filter on that in serialized repo.

oflatt:
Hmm after more consideration it's not great to make the format more complex. This PR that does it on the egglog side sounds good

oflatt:
Are you satisfied with just doing the change with the unit?

saulshanabrook:
Yeah I agree adding more metadata might not be the best! 

I would be curious if others have thoughts on special casing unit.  I thought that maybe since the default is no longer automatically set for functions that return unit, it might be best to not special case it in this case either. 

I have found omitting all defaults does make the graph slightly more succinct in a particular use case I have, but it’s not essential.

oflatt:
Unit is still a default when using the `relation` syntax sugar, right? I think in that case it might make sense to omit it.
But yeah happy to hear other opinions

saulshanabrook:
> Unit is still a default when using the relation syntax sugar, right? I think in that case it might make sense to omit it.

Yes, that is true!  In the Python bindings I don't use relation as much as just manually specifying `Unit` as the return types that I can use the normal function definitions in Python.

mwillsey:
After reading this, I think the clearest thing is just to leave `()` in the visualization. The less special casing, the better.

saulshanabrook:
@mwillsey Should I just close this PR then? Or would you accept a change to make this an option that defaults to false when serializing to omit defaults?

mwillsey:
I think we should probably just close it. It's really a very small amount of visual clutter (unless I'm missing something), plus it's more consistent.

saulshanabrook:
On small graphs, it's not a big deal, but I have noticed that on big graphs it can make the difference between being able to debug what's going on and not! It wouldn't be as big of a problem if our graph visualization was more interactive.

I was using this earlier today to debug something, if I end up needing to pick it up again I'll post an example here.

PR #249, opened 2023-10-04
Clean up and migrate all documentation to cargo docs

oflatt:
This PR cleans up our documentation and moves it to cargo doc.
It also gets rid of one foot gun which is forgetting to specify how many function rows to print.

Future work includes
- Run all our examples in the docs to make sure they are valid code
- Syntax highlighting in the docs
- runnable code in the docs
- egglog tutorial


src/ast/mod.rs
```diff
@@ -323,83 +323,346 @@ impl Display for NormSchedule {
...
+    /// [INFO ] Command failed as expected.
+    /// ```
     Check(Vec<Fact>),
+    /// Currently unused, this command will check proofs when they are implemented.
     CheckProof,
-    PrintTable(Symbol, usize),
+    /// Print out rows a given function, extracting each of the elements of the function.
+    /// Example:
+    /// ```text
+    /// (print Add 20)
```
yihozhang, line 640: I believe `print` is renamed to `print-table`.

src/ast/mod.rs
```diff
@@ -773,17 +1046,36 @@ impl Display for Fact {
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub enum Action {
+    /// Bind a variable to a particular datatype or primitive.
+    /// At the top level (in a [`Command::Action`]), this defines a global variable.
+    /// In a [`Command::Rule`], this defines a local variable in the actions.
     Let(Symbol, Expr),
-    /// `set` a table to a particular result.
+    /// `set` a function to a particular result.
```
yihozhang, line 1053: I like how you are consistent about table vs function

yihozhang (approved):
LGTM! Thanks for taking the initiative to make this long-awaited PR happen!

PR #254, opened 2023-10-09
Add booleans to egglog

oflatt:
Booleans are useful, this PR adds them. It also adds `bool-<` which returns a boolean. It is different from `<`, which is a filter that fails when the first argument is not less than the second.

I think `bool-<` and `<` are very confusing, but both are useful. Maybe they need better names? We should also disallow `<` in actions because it might fail. Opening an issue for that.


yihozhang:
I recently found a unifying mechanism for coping with failed computation (None) and boolean (false). The idea is that for the left-hand side `(atom1 ... atomn)`, we require every match to satisfy `atom1 /\ ... /\ atomn`, where the semantics of `/\` is defined as follows: the result is `#f` if any of the top-level atom *failed or is #f*. 

This way, comparison operators like `<` can have the natural type `(i64, i64) -> bool`, and you still allow failed computation (like `div`). Moreover, the negation operator can be defined in the usual way (`(!) :: bool -> bool`), which is currently not definable as a primitive. Queries like `((R x y) (= (= x y) false))` are also semantically meaningful under this view.

Not that we should refactor egglog to reflect this view- just thought I should share this.

oflatt:
I like this trick, but I worry it's a little confusing to users. Booleans would have a different meaning at the top level

src/sort/bool.rs
```diff
@@ -0,0 +1,55 @@
...
+        self.name
+    }
+
+    fn as_arc_any(self: Arc<Self>) -> Arc<dyn Any + Send + Sync + 'static> {
+        self
+    }
+
+    #[rustfmt::skip]
+    fn register_primitives(self: Arc<Self>, eg: &mut TypeInfo) {
+        add_primitives!(eg, "not" = |a: bool| -> bool { !a });
```
yihozhang, line 27: Have you considered symbol names like `!`, `&&`, and `||` to avoid conflict with potentially common names?

src/sort/i64.rs
```diff
@@ -54,6 +54,11 @@ impl Sort for I64Sort {
         add_primitives!(typeinfo, "<=" = |a: i64, b: i64| -> Opt { (a <= b).then(|| ()) });
         add_primitives!(typeinfo, ">=" = |a: i64, b: i64| -> Opt { (a >= b).then(|| ()) });
 
+        add_primitives!(typeinfo, "bool-<" = |a: i64, b: i64| -> bool { a < b });
```
yihozhang, line 57: An alternative set of names could be `is-less`, `is-greater`, `is-not-greater`, and `is-not-less` which makes it clear that they return booleans.

yihozhang (approved):
LGTM! I also suggested another set of names for boolean operations in case they are useful.

yihozhang:
> Booleans would have a different meaning at the top level

That is the intention! For atoms to be "conjunctive" all top-levels need to be true.

src/sort/bool.rs
oflatt: Good idea, I'll do that

src/sort/i64.rs
oflatt: Okay that sounds better

oflatt: Hmm, actually I like this for now because it tips you off on the bool output

src/sort/bool.rs
oflatt: I don't like these names because they are similar to `=` and `!=` and `>` which are filters

src/sort/bool.rs
```diff
@@ -0,0 +1,55 @@
...
+        add_primitives!(eg, "and" = |a: bool, b: bool| -> bool { a && b });
+        add_primitives!(eg, "or" = |a: bool, b: bool| -> bool { a || b });
+        add_primitives!(eg, "xor" = |a: bool, b: bool| -> bool { a ^ b });
+        add_primitives!(eg, "=>" = |a: bool, b: bool| -> bool { !a || b });
```
saulshanabrook, line 31: Is there a name for this operator? Sorry, I am not familiar with it.

saulshanabrook: Ah it's implication, nevermind.

PR #260, opened 2023-10-13
Add doctest to document extract()

gussmith23:


oflatt (changes requested):
Really great improvement to the docs-
If you would be so kind, would you add a small text description of what the code does for the documentation? We shouldn't add code examples without descriptions.
(Optional) Also would be good to link to this function from the existing documentation of the `query-extract` command

gussmith23:
Will do, but may have to wait til next week!

gussmith23:
@oflatt done. Please feel free to suggest different wording in both changes!

oflatt (approved):
Thanks a bunch!

PR #261, opened 2023-10-15
FIx bool's tag symbol

hatoo:
```
; Test bool's tag
(relation R (i64))
(function F (i64) bool)

(rule
    ((R i))
    ((set (F i) true))
)

(R 0)

(run 3)
```

fails
```
thread 'main' panicked at /home/hatoo/.cargo/git/checkouts/egglog-0aabc6dbd51d8ac4/a4559d2/src/function/mod.rs:183:17:
assertion `left == right` failed
  left: "bool"
 right: "Bool"
```

I guess this is because of a typo for bool's tag.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -70,7 +70,7 @@ impl From<Symbol> for Value {
 impl From<bool> for Value {
     fn from(b: bool) -> Self {
         Self {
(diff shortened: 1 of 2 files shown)


oflatt (approved):
Great catch!

PR #263, opened 2023-10-17
Support `(print-size)` to print all size of functions

### LINKED ISSUE #262: Feature request: `(print-size)` to list the size of all functions
When I'm writing `egglog`, ensuring no `functions` have a big size is a good indication of efficient code.

I think adding `(print-size)` command to list all function's sizes will be good.

hatoo:
fixes #262

Perhaps making another `Command` variant like `Command::PrintSizeAll` would be better.
Currently, only resuse ` Command::PrintSize(Symbol)` to `Command::PrintSize(Option<Symbol>)`


oflatt (approved):
Good change, thanks!

PR #265, opened 2023-10-24
Add support for matching on rationals

saulshanabrook:
This PR adds `numer` and `denom` functions to deconstruct rationals.

Otherwise, I couldn't seem to match on the pieces of them in rules, so this lets you do that now.


oflatt (approved):
Nice change!

PR #266, opened 2023-10-24
Add `list.egg` test file

Alex-Fischman:
Adds a test to egglog. I've been thinking about replacing Vecs with Lists, and realized that I could partially implement Lists in egglog.


tests/list.egg
```diff
@@ -0,0 +1,97 @@
...
+
+(function list-length (List) i64)
+(relation list-length-demand (List))
+(rule
+	((list-length-demand (Nil)))
+	((set (list-length (Nil)) 0))
+	:ruleset list)
+(rule
+	((list-length-demand (Cons head tail)))
+	((set (list-length-demand tail) ()))
```
yihozhang, line 15: Can you just write `((list-length-demand tail))` in place of this line?

tests/list.egg
```diff
@@ -0,0 +1,97 @@
...
+
+(function list-get (List i64) i64)
+(relation list-get-demand (List i64))
+(rule
+	(	(list-get-demand list 0)
+		(= list (Cons head tail)))
+	((set (list-get list 0) head))
+	:ruleset list)
+(rule
+	(	(list-get-demand list n) (!= n 0)
```
yihozhang, line 31: nits: `(> n 0)` instead of `(!= n 0)` which can cause potential non-termination

tests/list.egg
```diff
@@ -0,0 +1,97 @@
...
+	:ruleset list)
+(rule
+	(	(list-get-demand list n)
+		(= list (Cons head tail))
+		(= item (list-get tail (- n 1))))
+	((set (list-get list n) item))
+	:ruleset list)
+
+(function list-append (List List) List)
+(relation list-append-demand (List List))
```
yihozhang, line 43: For `list-append`, I actually think you don't need a demand relation. `list-append` itself can be used as demand, and then the semantics is just a few `rewrite`s like `(rewrite (list-append (Nil) list) list)`

tests/list.egg
```diff
@@ -0,0 +1,97 @@
...
+	 (= (list-append tail list) appended))
+	((union (list-append (Cons head tail) list) (Cons head appended)))
+	:ruleset list)
+
+; list-contains Nil _ => false
+; list-contains (Cons item tail) item => true
+; list-contains (Cons head tail) item => assert(head != item); (list-contains tail item)
+; list-contains needs inequality
+
+(function list-set (List i64 i64) List)
```
yihozhang, line 63: I bet you also don't need a separate demand relation for `list-set`.

yihozhang (changes requested):

yihozhang (approved):
LGTM!

PR #270, opened 2023-10-29
Small fixes to lambda.egg tests

yihozhang:


oflatt (approved):
Great change, now that check performs a query

PR #272, opened 2023-10-30
Update cargo.lock and toolchain so we can release

oflatt:


mwillsey (approved):
Looks good. Consider using a tool like cargo release to automate some of this. https://github.com/crate-ci/cargo-release

PR #278, opened 2023-11-01
Change s-expressions to use Symbol instead of String

oflatt:
`String` was a bad name. It corresponds more to symbol, since they are not wrapped in quotes.
This PR renames and updates to a newer version of my s-expr library that supports basic pretty printing


yihozhang (approved):
LGTM

PR #279, opened 2023-11-03
Make s-expression conversion public

oflatt:
This PR makes s-expression conversion public for convenience. This also allows users to pretty-print their s-expressions, which we need in the eggcc project.


yihozhang (approved):
Looks good! And thanks for the documentation

PR #280, opened 2023-11-03
Expose interface to add primitives

yihozhang:
Right now it is impossible to register a primitive without registering a user-defined sort.

This PR allows users to 1) directly register primitives and 2) get a handle of registered sorts (for primitive definition).

This is a breaking change since it changes the signature of `get_sort` in `TypeInfo` (cc @saulshanabrook ).

### DIFF
--- a/src/extract.rs
+++ b/src/extract.rs
@@ -123,7 +123,7 @@ impl<'a> Extractor<'a> {
     fn expr_from_node(&self, node: &Node, termdag: &mut TermDag) -> Option<Term> {
         let mut children = vec![];
         for value in node.inputs {
-            let arcsort = self.egraph.get_sort(value).unwrap();
+            let arcsort = self.egraph.get_sort_from_value(value).unwrap();
             children.push(self.find_best(*value, termdag, arcsort)?.1)
         }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -550,7 +550,7 @@ impl EGraph {
                 value
             }
         } else {
-            if let Some(sort) = self.get_sort(&value) {
+            if let Some(sort) = self.get_sort_from_value(&value) {
                 if sort.is_eq_sort() {
                     return Value {
                         tag: value.tag,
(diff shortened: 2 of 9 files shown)


oflatt (changes requested):
Looks good to me, but it's hard to see the implications without an example of a user adding a sort.
Would you mind adding a test? This could be a small crate in the repository that adds a new sort.
Would also be awesome to get a little bit of documentation while you are changing this.

yihozhang:
@oflatt This is a great suggestion. I tweaked the interface a little bit and this allows us to introduce a new primitive as follows (see `src/lib.rs`):

```

    struct InnerProduct {
        ele: Arc<I64Sort>,
        vec: Arc<VecSort>,
    }

    impl PrimitiveLike for InnerProduct {
        fn name(&self) -> symbol_table::GlobalSymbol {
            "inner-product".into()
        }

        fn get_type_constraints(&self) -> Box<dyn crate::constraint::TypeConstraint> {
            SimpleTypeConstraint::new(
                self.name(),
                vec![self.vec.clone(), self.vec.clone(), self.ele.clone()],
            )
            .into_box()
        }

        fn apply(&self, values: &[crate::Value], _egraph: &EGraph) -> Option<crate::Value> {
            let mut sum = 0;
            let vec1 = Vec::<Value>::load(&self.vec, &values[0]);
            let vec2 = Vec::<Value>::load(&self.vec, &values[1]);
            assert_eq!(vec1.len(), vec2.len());
            for (a, b) in vec1.iter().zip(vec2.iter()) {
                let a = i64::load(&self.ele, a);
                let b = i64::load(&self.ele, b);
                sum += a * b;
            }
            sum.store(&self.ele)
        }
    }

    #[test]
    fn test_user_defined_primitive() {
        egraph
            .parse_and_run_program(
                "
                (sort IntVec (Vec i64))
            ",
            )
            .unwrap();
        let i64_sort: Arc<I64Sort> = egraph.get_sort().unwrap();
        let int_vec_sort: Arc<VecSort> = egraph
            .get_sort_by(|s: &Arc<VecSort>| s.element_name() == i64_sort.name())
            .unwrap();
        egraph.add_primitive(InnerProduct {
            ele: i64_sort,
            vec: int_vec_sort,
        });
        egraph
            .parse_and_run_program(
                "
                (let a (vec-of 1 2 3 4 5 6))
                (let b (vec-of 6 5 4 3 2 1))
                (check (= (inner-product a b) 56))
            ",
            )
            .unwrap();
    }
```
A caveat here is that for container sorts, you need to first instantiate the container sort before introducing the primitive, which needs to get a handle on the sort.

oflatt (approved):
Great work!
This would be *even better* if the test you added was in the documentation, and explained it!

PR #281, opened 2023-11-03
Fix unexpected assertion failure when arity is mismatched

yihozhang:


oflatt (approved):
Great, and even includes a test case.

PR #283, opened 2023-11-06
Fix bugs in the constraint solver for `And`

yihozhang:


tests/type-constraints-tests.egg
```diff
@@ -0,0 +1,10 @@
+(datatype Operand)
+(sort VecOperandBase (Vec Operand))
+(datatype VecOperand (VO VecOperandBase))
```
saulshanabrook, line 3: Sorry I think I missed it maybe in an earlier PR, what does the `V0` do?

yihozhang: Ah it's a wrapper trick we used in eggcc that allows us to union two vectors. It's very specific to our application.

saulshanabrook: Ah sorry, I was just misreading the datatype constructor as a sort constructor and thought that `V0` was a presort. Nevermind!

oflatt (approved):
Looks good to me

PR #284, opened 2023-11-07
Better statistics for runs

oflatt:
This PR sorts the rules by time, and also adds a number of matches statistic.


yihozhang (approved):
looks good!

PR #285, opened 2023-11-08
Add integer equal boolean function

clyben:
This PR add `bool-=` function, which is similar to existing `bool-<` but missing. I found `bool-=` is needed in eggcc.


oflatt (approved):
great!

PR #287, opened 2023-11-08
Short-circuit calls to merge functions with the same argument.

ezrosent:
egglog runs rebuilding until a fixed point: while egglog programs do not always terminate, rebuilding really always should.

@oflatt wrote an egglog program in which rebuilding didn't terminate. The idea was:

* Have some function `(function F (X X) X)`
* Make `F` the merge function of some other function `(function G (Y) X :merge (F old new))`
* Have an entry `(G arg1) => x`, `x`'s canonical member changes to `y`, and `arg1`'s canonical member changes to `arg2`. and we also have `(G arg2) => y`. 

During rebuilding, we notice that these two entries are equal (not just their keys, but also their values). In the current egglog code we _still run the merge function_,  creating a new entry `(G arg2) => (F y y)`. Furthermore, we union `y` and `(F y y)`. Then:

* We rebuild `F`, potentially changing the canonical id for `y`.
* That causes us to rebuild `G`, creating `(F y (F y y))`.
* Then we rebuild `F` again...

This process continues _ad infinitum_.

This PR implements a short-term fix, which is to never run the merge function if we are replacing a tuple with itself: merge functions already do not behave well if they aren't idempotent. In this case `(F x x)` always equaled `x` _eventually_, just not until we got a chance to union it. I think this behavior should be correct for any useful merge function, but it feels like a hack. I'm open to other fixes (maybe we can initialize `(F y y)` to `y` and never call `make_set`?)

cc @yihozhang 

@oflatt do you mind if I add your test file to the egglog tests? It's a large file but it still takes less time than some of the math benchmarks.

### DIFF
--- a/src/function/mod.rs
+++ b/src/function/mod.rs
@@ -410,7 +410,7 @@ impl Function {
                         prev
                     }
                     MergeFn::Expr(_) => {
-                        if !appended {
+                        if !appended && prev != out_val {
                             deferred_merges.push((scratch.clone(), prev, out_val));
                         }
                         prev
--- a/tests/eggcc-extraction.egg
+++ b/tests/eggcc-extraction.egg
@@ -1,0 +1,1709 @@
+
+(datatype Literal)
+(datatype Expr)
+(datatype Operand)
+(datatype Body)
+
+(sort VecOperandBase (Vec Operand))
+(datatype VecOperand (VO VecOperandBase))
+
+(sort VecVecOperandBase (Vec VecOperand))
+(datatype VecVecOperand (VVO VecVecOperandBase))
+
+;; Type
+(datatype Type
+    (IntT)
+    (BoolT)
+    (FloatT)
+    (CharT)
+    (PointerT Type))
+(datatype EffectType
+    (Bril Type)
+    (PrintState))
+(sort FuncSigs (Vec EffectType))
+(datatype OptionType
+    (SomeType Type)
+    (NoneType))
+;; Literal
+(function Num (i64) Literal)
+(function Float (f64) Literal)
+(function Char (String) Literal)
+(function Bool (bool) Literal)
+
+;; Expr
+(datatype ConstOps (const))
+(function Const (Type ConstOps Literal) Expr)
+;; Call may return multiple values but at most one of them
+;; is a value type, which is stored in OptionType.
+;; The last fields denotes how many return values it have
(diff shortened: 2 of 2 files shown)


ezrosent:
That's right: this is a deferred merge. I'm actually not sure where the two get unioned. I thought it would be in `apply_merges` but that just calls `insert` on the result.

If we change `apply_merges` to union the previous value with the output of the merge function, that also fixes the bug. Do you prefer that code? It's making the same assumption about merge functions, but now we always populate the merge function table if we do a merge.

oflatt:
Do you mean only when the merge function is `MergeFn::Union`?
It doesn't seem right otherwise

ezrosent:
Yeah I think I agree. Take a look at the latest commit?

oflatt:
I think we already perform the union on line `404` in `function/mod.rs`?
It's just that the thing we insert is not updated?

ezrosent:
That's just when we have `MergeFn::Union`, this code fires when we have `MergeFn::Expr`, so line 404 doesn't run

oflatt:
Oh hmm, I'm not so sure I'm down for this change- maybe I still don't quite understand the problem.
My current guess at what is happening-
`(F y y)` and `y` are getting unioned already- so this change doesn't change the problem you mentioned in this PR. 
The actual but is that we are never using the union-find to look up the canonical version of the newly merged value.


If you wrap `to_insert` in a lookup does this solve the problem?

ezrosent:
Wrapping `to_insert` in a lookup does not solve the problem, sadly. But I don't think canonicalizing upon insert is enough to guarantee that the value will be canonical by the time you actually get to rebuilding `F`. I think the main thing you want may just be the code from the initial commit: just don't try and merge two tuples if they are already the same. Thoughts?

oflatt:
Interesting- what about looking up the two values that are being used by the merge function? I think I'm still confused as to how this bug is coming about if it's not a forgotten lookup somewhere.

ezrosent:
> what about looking up the two values that are being used by the merge function?

That's what the initial code in this PR does, I think?

oflatt:
Okay let's go with that then!

ezrosent:
Alright, I reverted to the initial code and DAG-ified the initial example. Lets see if CI is happy...

oflatt (approved):
Thanks for the fix! And good to get an eggcc test in there.

PR #288, opened 2023-11-13
Small improvements to egglog's primitive system

yihozhang:
* Expose into_box method
* Add a `map-length` primitive
* Add a `vec-remove` primitive


oflatt (approved):
lgtm

PR #290, opened 2023-11-14
Implement greedy dag extraction, as well as context-wrapper

oflatt:
This PR implements greedy dag extraction (inefficiently). It also allows users to "break out of" sharing sub-terms in the cost model using `context-wrapper`. Things in a context wrapper and outside of it are not shared in the cost function.


oflatt:
It's unclear if this is what we want in egglog, and it changes the egglog rust interface slightly. The current implementation is also pretty slow (though it's tricky to make it faster using functional/persistent data structures)

saulshanabrook:
Am I understanding it right that this PR changes the extraction algorithm to from considering the lowest cost total tree (with nodes that show up twice counting twice for example) with the lowest cost graph (where duplicate nodes are only counted once in the cost)?

And that it also adds a flag on functions `context-wrapper` to not have it share any cost between the descendents of instance of that function and other values outside that function?

I.e. if `context-wrapper` was the default for all functions, then would this be the same as our current behavior of looking at tree cost?

oflatt:
Yeah, I think if context-wrapper was the default it would replicate the current behavior

oflatt:
Closing for now, since it is unused
I think long term a better strategy would be to allow other crates to implement custom extractors on top of egglog

PR #291, opened 2023-11-20
docs: fix reference to vs code plugin to match repo

mgree:
The docs refer to an `vscode/egglog-1.0.0` or `vscode/vscode` directory, but neither of these exist: it should be (it seems) `vscode/eggsmol-1.0.0`.


mwillsey:
Thanks! I think we missed this in the great rename.

PR #293, opened 2023-11-20
Add option for eclass outputs to egglog serialize

oflatt:


oflatt:
@saulshanabrook I can't add you as a reviewer (yet)
But this one's for you!

src/serialize.rs
```diff
@@ -135,6 +139,14 @@ impl EGraph {
                 },
             );
         }
+
+        let roots = config
+            .root_eclasses
+            .iter()
+            .map(|v| self.serialize_value(&mut egraph, &mut node_ids, v, None).0)
```
oflatt, line 146: Is the None here correct?

saulshanabrook: Yeah that's correct. You only need to set the prim_node_id when we are emitting an e-clas for a primitive with its function as a pair. In this case, there wouldn't be any function to pair it with. 

It does mean that if `split_primitive_outputs` is true and you pass a primitive value as root_eclass, it will render to a class ID that's not in the graph. It isn't clear what the "right" answer would be in this case, because there might be multiple "e-classes" that all have that same primitive value in them, and there isn't neccesarsily a canonical one.

saulshanabrook (approved):
Looks good! I was wondering when we would need to add this, thanks!

PR #297, opened 2023-11-21
Add i64-fresh!

oflatt:
`i64-fresh!` is a primitive that allows you to get new integers each time it is called.
It's bad from a semantic point of view- running a rule multiple times gives different results!
However it is useful in practice.

Lets not merge yet until we are sure it is necessary.


saulshanabrook:
Are you using it for symbol generation?

oflatt:
Yeah, we need fresh integers as unique ids for an experimental version of eggcc.

saulshanabrook:
Nice.

If it's helpful at all, happy to go into the way I generated a unique integer per node, when [generating Python code in the e-class](https://github.com/egraphs-good/egglog-python/blob/main/python/egglog/exp/program_gen.py). [This](https://github.com/egraphs-good/egglog-python/blob/3e58a999c08807d3e3351ff8c89c2d74d0c464e6/python/tests/test_program_gen.py#L62-L74) is an example of how it's tested and the [generated output](https://github.com/egraphs-good/egglog-python/blob/main/python/tests/__snapshots__/test_program_gen/test_to_string.py).

You probably already looked into that sort of method tho.

oflatt:
Closing for now, as we are not using it anymore
I think that something has gone quite wrong if you need this

PR #299, opened 2023-11-25
User Defined Generics RFC

saulshanabrook:
I wanted to share this draft of an "RFC" (although I based the structure loosely off the Python PEP format, like [this one on generics](https://peps.python.org/pep-0695/#summary-examples)) on adding user-defined generic types and functions to egglog.

Happy for any feedback here or to set up a call if that's easier.

i.e. Does it make sense? Are there any questions left after reading it? Any confusion about the rationale or the syntax?
What do you think of it? Do you think implementing it is viable? Would it be useful for any of your work?

### DIFF
--- a/rfc/0001-generics.md
+++ b/rfc/0001-generics.md
@@ -1,0 +1,130 @@
+# Generic user defined types and functions
+
+## Abstract
+
+This is a proposal to add support for allowing users to add generic types and functions.
+
+## Motivation
+
+Currently, only primitive types and functions in `egglog` can be generic, i.e. `(Vec i64)`, but user defined types cannot be.
+
+User defined container types are helpful to implement, compared to the primitives, because they can be used in cases where the value of the expression is not known eagerly. However, implementing them currently requires re-writing their definitions and rewrite rules for every inner type.
+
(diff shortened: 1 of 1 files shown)


oflatt:
Thanks for the RFC @saulshanabrook!
Quick question: would it be possible to desugar all these problems with generics on the python side?
As in, the python bindings would generate egglog code, one function per instantiation.
That way, you could experiment with generics and we won't have to modify egglog (yet!)

Egglog is about to go through lots of cleanup, so it's not a great time to add complexity to the type system

saulshanabrook:
Hey @oflatt yeah that makes sense. I believe it would be doable by generating new datatypes for each instantiation of a generic type.

For rewrites/rules that use generic types, they are generated for every pairwise combination of generic types that they use and are instantiated.

saulshanabrook:
I don't think this needs to be considered right now, but re opening this bc I thought of an issue with the desugaring approach... 

If there is a rule that has facts that use type T and actions that create type List[T] what concrete types do you instantiate this with? It seems like at the limit it could be infinite if it applies recursively.

saulshanabrook:
I'll re-close this actually and open an issue if I want to look more into how to implement generics as pre-processing step.

PR #301, opened 2023-11-27
Adds support marking exprs as subsumed

### LINKED ISSUE #256: How to do directional transform?
Say I have a `a*b + c` and I then decide to add an `fma` instruction to my program. I want to directionally **fuse** the original expr as `fma(a, b c)` because it is faster.

Maybe assigning them with properly crafted costs will solve the problem. Then this becomes an unmanageable ordering problem.

In a real world program, especially in the field of deep learning, there are numerious fusion patterns can be used. Properly ordering all of then with cost is not pratical. So the idea here is to decompose all pattern into steps, apply them one by one (with human insight on the order of applying).

```
--->: apply fusion then extract or discard (don't change the program) if not apply

original program --> program with fusion pattern 1 --> program with fusion pattern 1 and 2 --> ...
````

And this also allows us to maintain the rules of each step seperately in a standalone library/file.

Is this kind of directional transfrom achievable? If yes how?

[1] [ægraphs: Acyclic E-graphs for Efficient Optimization in a Production Compiler](https://pldi23.sigplan.org/details/egraphs-2023-papers/2/-graphs-Acyclic-E-graphs-for-Efficient-Optimization-in-a-Production-Compiler)
[2] https://github.com/bytecodealliance/rfcs/blob/main/accepted/cranelift-egraph.md

### PR DESCRIPTION
**Pull request by saulshanabrook**
**EDIT**: This PR has been changed to allow "subsuming" nodes, which are both unextractable and cannot be matched during rules: https://github.com/egraphs-good/egglog/pull/301#issuecomment-1934665417

---

Previously, we had support for marking a whole function as unextractable. This PR extends that to also allow marking individual rows of a function as unextractable as well.

It also adds a high-level `:unextractable` parameter for the `rewrite` command to mark the LHS as unextractable after its been applied.

This is one way to support directional transformations/subsumption (closes #256 for now).

As I mentioned in that issue, one concrete use case for this is in my rewrites for numba compatibility of array expressions. In that example, I have to [translate `mean` into `sum` expressions, when an `axis` parameter is passed to mean](https://github.com/egraphs-good/egglog-python/blob/e8bcca69079d39514703d72bd2077896b3077791/python/egglog/exp/array_api_numba.py#L15-L30). I want the rewritten version extracted, but by default, the cost would be higher so it wouldn't be. My workaround at the moment was to [give the whole `mean` function a huge cost](https://github.com/egraphs-good/egglog-python/blob/main/python/egglog/exp/array_api.py#L1217-L1219) so it would always be avoided if possible.

However, that solution isn't ideal for a few reasons:

1. It mixes the rewrite logic with the original function definition. I would like all the numba logic to be contained in that separate module.
2. It makes the extraction take longer because it's considering these large costs instead of just ignoring those options altogether.

With this PR, I could move the logic entirely to the rewrite.

## Implementation

This PR implements the feature by storing a set of "unextractable" input values to each function. I am not sure if this is like safe for rebuilding, so I would love some feedback on that. If there is a better way of storing this information, I am happy to change it.

### CODE REVIEW

**Comment by saulshanabrook** [comment]
Thanks for taking a look @oflatt!

> This probably doesn't work with rebuilding- the approach I would take would be to store in the tables in the back-end which rows are unextractable. Later, any e-node equal to it should also be marked unextractable.


Will do, I will move it into the table, as part of the `vals`.

> It would be cool, while we are add it, to add a way to mark things as subsumed- they will be extractable but not e-match-able.
> We could also have things that are neither e-matchable nor extractable if they have both flags

Oh yeah, I like this behavior. I was already worried that if like you mark something as unextractable, then it could still be matched and return a similar expression that would then be extractable. What if we just had one flag called `subsumed` that made rows un-extractable and un-matchable? Cna you think of a use case where you would only want one and not the other?

This would also give a better name to the rewrite flag, of `:subsume` instead of `:unextractable` which is a bit ambiguous and confusing I think

**Comment by saulshanabrook** [comment]
> This probably doesn't work with rebuilding- the approach I would take would be to store in the tables in the back-end which rows are unextractable. Later, any e-node equal to it should also be marked unextractable.

I added a few failing tests for this. I am realizing it's actually a bit more complicated than I imagined... Because you might already have canonicalized two expressions into the same value, but then you just want to mark one as unextractable, so you effectively have to un-canonicalize them. I think it will have to happen in the [`check_action`](https://github.com/saulshanabrook/egg-smol/blob/190dcd54940b3d1895eb0436ec4b33c6e4969d42/src/typecheck.rs#L522-L529) or in the [`run_actions`](https://github.com/saulshanabrook/egg-smol/blob/190dcd54940b3d1895eb0436ec4b33c6e4969d42/src/typecheck.rs#L838-L839) block where you translating `Expr` to `Value`.

**Comment by oflatt** [comment]
I'm guessing there are cases where we might want something to be not extractable, but still matchable or vice-versa.
For example, if you find something you know is better, but you still want to be able to match on this e-node to find something even better than that.

**Comment by saulshanabrook** [comment]
Ok I can add actions then for both of them and store the info separately. 

Maybe the high level rewrite flag though could enable both and be called like ":replace", since the combination semantically is like you delete the old one, but it just can also never be re-added.

**Comment by saulshanabrook** [comment]
This is ready for review again. I have updated it to:

1. Add a `subsume` action which marks a row as subsumed and unable to be queried when running rules. It can still be queried when running a check.
2. Add the `:replace` instead of the `:unextractable` argument to `rewrite`, which marks the LHS as subsumed and unextractable.

File: src/ast/mod.rs [hunk 1]
```diff
@@ -1107,14 +1123,31 @@ pub enum Action {
...
+}
+
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub enum NormAction {
     Let(Symbol, NormExpr),
     LetVar(Symbol, Symbol),
     LetLit(Symbol, Literal),
     Extract(Symbol, Symbol),
     Set(NormExpr, Symbol),
-    Delete(NormExpr),
```
saulshanabrook, line 1117: I collapsed this into one normalized action, since much of the behavior is the same for all three actions.

src/function/mod.rs
```diff
@@ -61,6 +63,21 @@ impl ResolvedSchema {
     }
 }
 
+impl Debug for Function {
```
saulshanabrook, line 66: I added some debug implementations since they were helpful when debugging failing test cases.

src/function/mod.rs
```diff
@@ -390,36 +429,42 @@ impl Function {
...
-                    appended = true;
-                }
-                match &self.merge.merge_vals {
-                    MergeFn::Union => {
-                        debug_assert!(self.schema.output.is_eq_sort());
-                        uf.union_values(prev, out_val, self.schema.output.name())
+        self.nodes.insert_and_merge(
+            scratch,
+            timestamp,
+            |prev| {
```
saulshanabrook, line 435: This looks like a big change, but I think the function remained the same. The formatted just changed the indentation of the function when I added another arg.

src/function/table.rs
```diff
@@ -46,12 +47,31 @@ struct TableOffset {
...
+    }
+}
+
 #[derive(Default, Clone)]
 pub(crate) struct Table {
     max_ts: u32,
     n_stale: usize,
     table: RawTable<TableOffset>,
-    pub(crate) vals: Vec<(Input, TupleOutput)>,
+    pub(crate) vals: Vec<Row>,
```
saulshanabrook, line 74: I wasn't sure if the unextractable and subsumed metadata should be on the input or output, so I just put them on the row itself and made it a struct.

src/gj.rs
```diff
@@ -885,12 +909,20 @@ impl<'a> std::fmt::Display for TrieAccess<'a> {
...
+    ) -> impl Iterator<Item = usize> + 'a {
+        ixs.iter().copied().filter_map(move |ix| {
             let ix = ix as usize;
             let (inp, out) = self.function.nodes.get_index(ix)?;
             if self.timestamp_range.contains(&out.timestamp)
                 && self.constraints.iter().all(|c| c.check(inp, out))
+                // If we are querying to run a rule, we should not include subsumed expressions
+                // but if we are querying to run a check we can include them.
+                // So we have a flag to control this behavior and pass it down to here.
+                && (include_subsumed || !self.function.nodes.get_index_row(ix).unwrap().subsumed)
```
saulshanabrook, line 925: Unfortunately, this was the cleanest way I found to make the query engine ignore the subsumed values, only when we were running rules, but not when running a check.

tests/files.rs
```diff
@@ -125,22 +125,28 @@ fn generate_tests(glob: &str) -> Vec<Trial> {
             test_terms_encoding: false,
         };
         let should_fail = run.should_fail();
+        // Marking as subsumed and non extractable is not supported for eqsat values with the term encoding
```
saulshanabrook, line 128: Delete is also unsupported in the term encoding with eqsat return values.

src/ast/parse.lalrpop
```diff
@@ -60,9 +60,10 @@ Command: Command = {
     LParen "ruleset" <name:Ident> RParen => Command::AddRuleset(name),
     LParen "rule" <body:List<Fact>> <head:List<Action>> <ruleset:(":ruleset" <Ident>)?> <name:(":name" <String>)?> RParen => Command::Rule{ruleset: ruleset.unwrap_or("".into()), name: name.unwrap_or("".to_string()).into(), rule: Rule { head, body }},
     LParen "rewrite" <lhs:Expr> <rhs:Expr>
+        <replace:(":replace")?>
```
oflatt, line 63: I think I would prefer to have both `:subsume` and `:unextractable` here.
I think it's more clear, and users don't have to learn that `:replace = :subsume + :unextractable`

src/ast/desugar.rs
```diff
@@ -257,16 +277,48 @@ fn flatten_actions(actions: &Vec<Action>, desugar: &mut Desugar) -> Vec<NormActi
                 let added_variants = add_expr(variants.clone(), &mut res);
                 res.push(NormAction::Extract(added, added_variants));
             }
+            // TODO: Reduce duplication in these three cases
```
oflatt, line 280: I think you can do this TODO by matching on `Action::Delete(symbol, exprs) | Action::Unextractable(symbol, exprs)...`

src/ast/desugar.rs
```diff
@@ -550,9 +602,13 @@ pub(crate) fn desugar_command(
...
         Command::Declare { name, sort } => desugar.declare(name, sort),
         Command::Datatype { name, variants } => desugar_datatype(name, variants),
-        Command::Rewrite(ruleset, rewrite) => {
-            desugar_rewrite(ruleset, rewrite_name(&rewrite).into(), &rewrite, desugar)
-        }
+        Command::Rewrite(ruleset, rewrite, replace) => desugar_rewrite(
+            ruleset,
+            rewrite_name(&rewrite).into(),
+            &rewrite,
+            replace,
```
oflatt, line 609: nit: rename to `is_replace`

src/ast/mod.rs
```diff
@@ -1080,6 +1090,12 @@ pub enum Action {
     /// Be wary! Only delete entries that are subsumed in some way or
     /// guaranteed to be not useful.
     Delete(Symbol, Vec<Expr>),
+    /// Set an entry to be `unextractable`
```
oflatt, line 1093: I think this documentation should clarify what an "entry" is.
What we mean is an "e-node" or a constructor and it's children e-classes.
In particular, any e-node that becomes equal to it will also be unextractable

src/ast/mod.rs
```diff
@@ -1080,6 +1090,12 @@ pub enum Action {
...
+    /// so that it cannot be extracted.
+    /// Note that this cannot be an expr but has to be a symbol and args, because we need to refer to a specific row
+    Unextractable(Symbol, Vec<Expr>),
```
oflatt, line 1096: I think in typechecking we should disallow calling `Unextractable` on non-datatype tables, since they can't be extracted anyways. That way they get a good error message.

src/ast/mod.rs
```diff
@@ -1080,6 +1090,12 @@ pub enum Action {
...
+    /// `subsume` an an entry so that it cannot be queries for during rules or rewrites.
+    Subsume(Symbol, Vec<Expr>),
```
oflatt, line 1098: On the other hand all table entries can be subsumed.

src/ast/mod.rs
oflatt: great!

src/function/mod.rs
oflatt: Nice

src/gj.rs
oflatt: I agree that passing include_subsumed everywhere is not ideal- but it's okay, I can't think of another way either.
If we need to have more metadata in the future we can package it up.

tests/files.rs
oflatt: I'll make this work when I re-write the terms encoding again (:

oflatt (commented):
Really impressive work @saulshanabrook, nice job with this one. This has been a wishlist feature for many people.
I want to check with the egglog team on this change since it'll have a long-term impact. Also @yihozhang wants to review

src/function/table.rs
```diff
@@ -153,23 +217,27 @@ impl Table {
...
             let k = mem::take(&mut inp.data);
             let new_offset = self.vals.len();
-            self.vals.push((
+            self.vals.push(Row::new(
                 Input::new(k),
                 TupleOutput {
                     value: next,
                     timestamp: ts,
                 },
+                unextractable,
```
yihozhang, line 226: If I understand the behavior here correctly, it would throw away the current `unextractable` and `subsumed` status of the row and always use the passed-in statuses instead. This does not feel right to me. For example, if a row is originally marked as unextractable but not subsumed, and later someone calls insert_and_merge with `unextractable=false` but `subsumed=true`. This would make the row extractable again. Instead, the correct behavior is to view unextractable and subsumed as a boolean lattice and take the lattice join of the old and current status (i.e., `||`).

src/gj.rs
yihozhang: `include_subsumed` seemed to be an intrinsic property of `TrieAccess`, just like `timestamp_range`. It defines whether TrieAccess should build a trie only based on unsubsumed rows. So ideally, it should only be passed at places where `timestamp_range` is also passed. Can't we make include_subsumed a field of TrieAccess instead of passing it everywhere?

src/function/table.rs
```diff
@@ -137,14 +195,20 @@ impl Table {
         inputs: &[Value],
         ts: u32,
         on_merge: impl FnOnce(Option<Value>) -> Value,
+        unextractable: bool,
```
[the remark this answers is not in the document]
yihozhang: Giving it a second thought, I think `unextractable` and `subsumed` can also be part of `TupleOutput` and we can simply get rid of `Row` struct. The rationale being TupleOutput is supposed to store everything that is determined by the input columns, that is, (meta-)data that are dependent on the input. This is why both timestamp and output are stored in this struct. 

I think either way, it would be great to make sure that `unextractable` and `subsumed` are treated and organized the same way as `timestamp`, since both are row metadata.

Another nice thing about them being encapsulated within the same abstraction is that we can bit-compress these meta data (e.g., store in a single `u64`).

src/typecheck.rs
```diff
@@ -516,7 +516,26 @@ impl<'a> ActionChecker<'a> {
                 let (_, _ty) = self.infer_expr(&fake_call)?;
                 let fake_instr = self.instructions.pop().unwrap();
                 assert!(matches!(fake_instr, Instruction::CallFunction(..)));
-                self.instructions.push(Instruction::DeleteRow(*f));
+                self.instructions
+                    .push(Instruction::ChangeRow(ChangeRow::Delete, *f));
+                Ok(())
+            }
+            Action::Unextractable(f, args) => {
```
yihozhang, line 523: Nits: code duplications that could potentially be avoided by extracting a common function.

The refactoring Oliver and I are working on will get rid of this type checker for actions, so this part will need to be rewritten anyway, so this is not a very big issue.

yihozhang (commented):
Thanks for making this PR. This is a well-desired feature, and the implementation is nice: I'm working on a big refactoring of the front-end with Oliver, and thanks to the clarity and modularity of this PR, I think it is not that hard to port this PR on top of the refactoring after that is done.

However, I do have some questions. See my comments. The most outstanding issue is that the current implementation is not sound.
```
(datatype Expr
  (f Expr)
  (Num i64)
  )

(f (Num 1))
(unextractable (f (Num 1)))
(f (Num 2))

(push)
(union (Num 2) (Num 1))
(extract (f (Num 2))) ;; produces (f (Num 2))
(extract (f (Num 1))) ;; produces (f (Num 2))
(pop)

(push)
(union (Num 1) (Num 2))
(extract (f (Num 2))) ;; crashes
(pop)
```
This is because insert_and_merge always overrides the old unextractable and subsumed status.

yihozhang:
Also, now that we have subsume, do we still need delete? Subsume seems to be the safe alternative of delete.

saulshanabrook:
Thank you both for the comprehensive reviews! I appreciate it. I will work on resolving all the comments and the unsound behavior. 

> Also, now that we have subsume, do we still need delete? Subsume seems to be the safe alternative of delete.

I was thinking thinking something similar. I wasn't sure if there were use cases that deleted things and then re-added them later on purpose, since subsuming is like a permanent delete.

saulshanabrook:
@yihozhang Thanks for the example: 

```
(datatype Expr
  (f Expr)
  (Num i64)
  )

(f (Num 1))
(unextractable (f (Num 1)))
(f (Num 2))

(push)
(union (Num 2) (Num 1))
(extract (f (Num 2))) ;; produces (f (Num 2))
(extract (f (Num 1))) ;; produces (f (Num 2))
(pop)

(push)
(union (Num 1) (Num 2))
(extract (f (Num 2))) ;; crashes
(pop)
```

My understanding of the expected behavior is that, actually, I believe both branches should crash, not just the second one. Or at least they should both fail to extract, I am not sure if there is a different way to signal that that all equivalent expressions are unextractable.


Why? Because as I understand it, the semantics of `(unextractactable (f (g))` is that "We should never extract the function `f` if its argument is **equivalent** to `(g)`". 

Why should this be the semantics instead of the alternative "We should never extract the function `(f (g))`"? I originally thought that's what the semantics should be, but I changed my mind when I realized that it would be much more complicated to implement and differ from how `delete` works.

Implementing it that way would require making sure that every node in the tree of the whole expression is not canonicalized as the same value as any other node. If they have been canonicalized to the same node id, then they would have to be split before they can be registered as unextractable. 

Let me know what you think about this. If it would be helpful to jump on a call to discuss if I am not making sense, happy to do that, it might be easier to talk about with pictures.

yihozhang:
This sounds right and is exactly what I was proposing: when you merge two
rows, you also want to merge their unextractable flags, which exactly
corresponds to this behavior. And as you said, we also don’t want to split
a canonicalized row.

src/ast/desugar.rs
saulshanabrook: Yeah but then how do I differentiate which branch it went down to match to decide which `ChangeRow` enum to use?

src/lib.rs
```diff
@@ -1694,4 +1697,351 @@ mod tests {
             )
             .unwrap();
     }
+    #[test]
+    fn test_unextractable_action_extract() {
+        // Test when an expression is set as unextractble, it isn't extracted, even if its the cheapest
+        let mut egraph = EGraph::default();
```
saulshanabrook, line 1703: Probably should move these tests to a standalone file at some point.

oflatt (changes requested):
Couple small tweaks

oflatt:
The semantics of egglog are getting hard, and subsumption is kind of tricky
We should consider putting this PR on hold until after we make a semantics file
https://github.com/egraphs-good/egglog/issues/314

src/ast/desugar.rs
```diff
@@ -25,12 +25,35 @@ fn desugar_rewrite(
...
+    unextractable: bool,
+    subsume: bool,
     desugar: &mut Desugar,
 ) -> Vec<NCommand> {
     let var = Symbol::from("rewrite_var__");
     // make two rules- one to insert the rhs, and one to union
     // this way, the union rule can only be fired once,
     // which helps proofs not add too much info
+    let mut head = vec![Action::Union(Expr::Var(var), rewrite.rhs.clone())];
+    if unextractable {
```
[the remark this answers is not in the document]
saulshanabrook: I don't believe so? For example, this works:

```
(datatype Math
  (Num i64))

(let a (Num 10))

(rewrite a (Num 11))

(run 1)
```

yihozhang:
I actually think merging this is fine given that this PR also removes `delete`, and the official semantics can just be based on `subsume` (instead of `delete`). We are already fairly confident that `subsume` is a better replacement for `delete`.

saulshanabrook:
@oflatt I moved the tests I added to a separate file in the `tests` dir. This did require me to make a few things public, so that the the tests could then access them. 

@yihozhang I didn't update this PR to remove`delete`, but I am happy to update it to do that. I wasn't sure if there were still some use cases for delete, but I don't have any personally.

oflatt:
I think we should consider carefully before we remove delete-
It's the only way to free memory, and I use it in the terms encoding.

I also still prefer to merge a semantics file (I'm working on it) and update it with subsume semantics before merging this.

saulshanabrook:
Is the plan then to merge #320 before this? It will require some refactoring then on this branch to adapt to the refactor.

oflatt:
I think we should probably merge yours first, and not make your life so hard

oflatt:
I'm thinking
1. semantics.md (coming soon!)
2. this
3. #320 
Would that be a good plan?

yihozhang:
It's easier to re-implement this PR on top of #320 (by copying things over
and adapting) than to merge this first and deal with hideous merge
conflicts, so we should probably merge #320 first. I'm happy to do the
re-implementation.

Since it could require several weeks work to get semantics.md right/reach
consensus, we probably also don't want everything to be blocked by
semantics.md.

saulshanabrook:
> It's easier to re-implement this PR on top of [#320](https://github.com/egraphs-good/egglog/pull/320) (by copying things over and adapting) than to merge this first and deal with hideous merge conflicts, so we should probably merge [#320](https://github.com/egraphs-good/egglog/pull/320) first. I'm happy to do the re-implementation.

That sounds good. I'm also happy to do it!

> Since it could require several weeks work to get semantics.md right/reach consensus, we probably also don't want everything to be blocked by semantics.md.

I agree that getting the semantics written down in a reasonable might take a bit and blocking based on it might rush that process.

oflatt:
Here's a concern I have with this PR right now: How does/should subsumption work in the presence of merge functions?

For example, if I mark `(lower-bound a 1)` as subsumed, but then I find also that `(lower-bound a 2)`, should the result `(lower-bound a 2)` also be marked as subsumed? Probably not, I think.
This seems fine since `(lower-bound a 1)` will also never be found added again due to the well-behaved merge function.

But then I can think of the following case which seems bad.
Define a function `(function MyLower (Math) Math :merge (MyMax old new))`
We find a lower bound `(set (MyLower (Num 2)) (Num 2))`
Then `(Subsume (MyLower (Num 2) (Num 2)))` for some reason
Then we find another lower bound `(set (MyLower (Num 2)) (Add (Num 1) (Num 1)))`
This gets rid of the subsumption 
Later we find that `(Add (Num 1) (Num 1)) = (Num 2)`
And now we have a term that we subsumed earlier, marked as not subsumed.

src/function/table.rs
```diff
@@ -226,15 +262,23 @@ impl Table {
...
+        if !inp.live() || (!include_subsumed && out.subsumed) {
             return None;
         }
         Some((inp.data(), out))
     }
 
+    // pub(crate) fn get_index_row(&self, i: usize) -> Option<&Row> {
+    //     self.vals.get(i)
+    // }
+
```
saulshanabrook, line 281:
```suggestion

```

saulshanabrook:
Hey @oflatt I was trying to make an example for the behavior you are describing and wasn't exactly sure how to do so. Here is how I far I got, trying to cover your first paragraph:

```
(datatype Math
  (Num i64)
  (Var String)
  (Add Math Math)
  (MyMax Math Math))

(function MyLower (Math) Math :merge (MyMax old new))

(rewrite (MyMax (Num a) (Num b)) (Num a) :when ((>= a b)))
(rewrite (MyMax (Num a) (Num b)) (Num b) :when ((< a b)))


(let a (Var "a"))
(set (MyLower a) (Num 1))
(subsume (MyLower a))
(set (MyLower a) (Num 2))

(run 1)

```

I am confused when you say "if I mark `(lower-bound a 1)` as subsumed" because we don't include the output in the subsume command, just the inputs, so you would mark `(lower-bound a)` as subsumed, not `(lower-bound a 1)`.

oflatt:
Oh, I see- I think I was just confused. I thought subsumption included the output.
Subsumption doesn't include the output, so I guess it'll always be subsumed regardless.

oflatt:
Do you think we should add some sort of warning if a whole e-class has been subsumed? I think it might be a potential foot-gun, and hard to prevent.

Example:
```
(datatype Math
  (Num i64)
  (Add Math Math))

;; lets subsume this  
(subsume (Add (Num 2) (Num 1)))
;; for some reason we like this version better
(Add (Add (Num 1) (Num 1)) (Num 1))

;; later we do constant folding
(union (Add (Num 1) (Num 1)) (Num 2))

;; now as far as I can tell we only have (Add (Num 2) (Num 1)) in the e-class, and it is subsumed.
;; for unextractable, that means we can't even extract anything
```

saulshanabrook:
> Do you think we should add some sort of warning if a whole e-class has been subsumed? I think it might be a potential foot-gun, and hard to prevent.

Maybe we could expose the unextractable and subsumed status in the visualization? I was also thinking this would be helpful for debugging. 

If we do this though, I would prefer to do it in a follow up PR, since it requires a bit of coordination with the egraph-serialized repo.

saulshanabrook:
I joined the egglog meeting today and we came up with a proposal that we:

* Remove the unextractable keyword
* Never extract things that are subsumed
* Don't allow marking customs with custom merge functions as subsumed.

Thank you all!

I will update this PR to reflect these comments.

---

We also talked about some other possible workarounds, including separating unextractable/costs for functions from the function definitions so you could use different ones.

We also remarked that if you have multiple rules that subsume an expression, those could match at the same time and they would both be matched even though it is subsumed. It was expressed this was OK because this is similar to other actions that also act simultaneously after matching, even if it's unexpected. You could avoid this if you want by splitting rules into rulesets and running them one at a time.

saulshanabrook:
I have updated the PR to address the points from the discussion. This is ready for another review.

src/ast/mod.rs
```diff
@@ -961,10 +981,8 @@ pub enum GenericAction<Head, Leaf, Ann> {
         Vec<GenericExpr<Head, Leaf, Ann>>,
         GenericExpr<Head, Leaf, Ann>,
     ),
-    /// `delete` an entry from a function.
-    /// Be wary! Only delete entries that are subsumed in some way or
-    /// guaranteed to be not useful.
-    Delete(Ann, Head, Vec<GenericExpr<Head, Leaf, Ann>>),
+    /// Subsume or delete an entry from a function.
```
yihozhang, line 984: optional: leave a note that `delete` is more dangerous than `subsume`?

src/function/table.rs
```diff
@@ -226,10 +248,14 @@ impl Table {
...
-    pub(crate) fn get_index(&self, i: usize) -> Option<(&[Value], &TupleOutput)> {
+    /// Returns the entries at the given index if the entry is live (and possibly not subsumed) and the index in bounds.
+    pub(crate) fn get_index(
+        &self,
+        i: usize,
+        include_subsumed: bool,
+    ) -> Option<(&[Value], &TupleOutput)> {
         let (inp, out) = self.vals.get(i)?;
-        if !inp.live() {
+        if !inp.live() || (!include_subsumed && out.subsumed) {
```
yihozhang, line 258: I'm a little less sure about the changes to `table.rs`. Should `Table` internalize the knowledge of whether some tuple is subsumed, or we always handle them at GJ/Query level? If it is the former, does it suffice to just change this one place? Other methods like `iter` and `iter_range` do not take an `include_subsumed` flag. Could you maybe give more rationale on this change?

yihozhang (commented):
This looks great! Thanks for adapting the PR to the new refactored codebase and incorporating conclusions from discussions! I left a small question. I believe once @oflatt approves, we should be able to merge this PR.

src/ast/mod.rs
saulshanabrook: I updated the docs to make it more explicit what subsume does. There is also a longer docstring on the `Change` enum above.

src/function/table.rs
saulshanabrook: Sometimes we want to filter by subsumption and sometimes not... In GJ for example it depends on whether we are running a check or matching on a rule.

I didn't include that check in the `iter` since most of the use cases for `iter` (and its relatives) do include subsumed I believe, such as making the tries and indices.

I think I will update it though to make that arg explicit to `iter` et al, to mirror the API.

saulshanabrook: Thanks for this feedback! I added the flag to the other iter methods as well and I think it's cleaner now.

src/actions.rs
```diff
@@ -134,7 +135,7 @@ enum Instruction {
     CallPrimitive(Primitive, usize),
     /// Pop function arguments off the stack and delete the corresponding row
```
oflatt, line 136: This documentation is not correct.
Could you please change it to give an example of `delete`, and example of using `subsume`, and what each of them means?

src/ast/mod.rs
```diff
@@ -430,7 +432,13 @@ pub enum GenericCommand<Head, Leaf> {
     ///          :when ((= a (Num 0)))
     /// ```
     ///
-    Rewrite(Symbol, GenericRewrite<Head, Leaf, ()>),
+    /// Add the`:subsume` flags to apply those to the left hand side on a rewrite when it's mtched.
+    ///
+    /// ```text
+    /// (rewrite (Mul a 2) (bitshift-left a 1) :subsume)
```
oflatt, line 438: Could you show the desugared version here, and link to the documentation for subsume? (see prev comment)

oflatt (changes requested):
Looks almost ready to merge!

src/ast/mod.rs
saulshanabrook: Thanks for catching this! Just added a desugared version and linked to the `Change`/

src/actions.rs
saulshanabrook: I fixed this line. The docs for `Change` should cover what each arm does in more detail, so it looks sufficient now to me compared to the docs for the other Instructions.

oflatt (approved):
Looks great! 🎉 🎉

saulshanabrook:
I wanted to note that I am now integrating these changes into the Python code and I can now remove my custom magic cost numbers! https://github.com/egraphs-good/egglog-python/pull/127/files#diff-9a9a81a04022815662adeb365e546523cb4376d5c1c797be3131968605f1c4c8

Thank you all for the time/energy spent refining this.

PR #302, opened 2023-11-27
Add slide mode to egglog website, good for demos

oflatt:


web-demo/static/index.html
```diff
@@ -77,66 +78,185 @@
...
-                height,
+                width: width,
+                height: height,
                 // Don't animate transitions between shapes for performance
                 tweenPaths: false,
                 tweenShapes: false,
             })
                 .transition(d3.transition().duration(2000))
                 .dot(message.data.dot).render();
-            // If we have made a zoom selection, reset that before transitioning
```
oflatt, line 107: @saulshanabrook why did we reset zoom here? it seems to work better in general without it

saulshanabrook (changes requested):
When I try this locally with `make serv` and `open http://[::]:8080/?example=eqsat-basic` the page seems to crash (possibly an infinite loop somewhere?) when I click `Slidemode` button.

oflatt:
Good find! I fixed the infinite loop.
This PR unfortunately includes a small s-expression parser.

saulshanabrook (dismissed):
Looks good! I gave it a go and seemed to work mostly as expected, letting you flip through the commands and see the output change.

https://github.com/egraphs-good/egglog/assets/1186124/737c6aed-c6f4-4634-b2b1-d28225bd1181

mwillsey:
Why do you need an sexp parser? seems like splitting on \n\n would work?

oflatt:
> Why do you need an sexp parser? seems like splitting on \n\n would work?

That's a good point, I'll just make that change!

mwillsey:
Otherwise I love the idea! Makes for super simple presentations.

oflatt:
I remember now- for the herbie demo, the code is too dense to add an extra newline in-between each command.
It works a lot better to split by s-expression.

mwillsey:
I'm a little spooked by the sexp parser. Perhaps we could choose a different separator then? Something like `\n\n\n` or `\n\n;;`? Or even `;;;;` or something? The sexp parser will be brittle to parens in other places probably, right?

oflatt:
Yep, the sexp parser is very evil
So three newlines makes my problem worse- the code has even more space in between
Having a comment that separates every "slide" could work, kind of ugly though

mwillsey:
Wait maybe i'm misunderstanding. You want to build in each top-level command individually? That seems like a bit much, you probably want to define rewrites in a big group, for example. Seems like `\n\n` or even 3 should be fine, it will just reduce the number of slides you have.

oflatt:
I'm happy to click though many lines at once- but I don't want is to introduce several lines at once without explaining them

mwillsey:
Ok! Sounds good then!

oflatt:
I'll merge this as-is then? Could I get an approval @mwillsey?

mwillsey:
What will this do for parens in comments?

oflatt:
It will treat the comment as a slide
I can make it skip comments though!

mwillsey:
Ok as long as it doesn’t break! Sounds good to me!

saulshanabrook (approved):

PR #304, opened 2023-11-30
Merge egglog language server

hatoo:
Resolve https://github.com/hatoo/egglog-language-server/issues/4

Added https://github.com/hatoo/egglog-language-server and https://github.com/hatoo/tree-sitter-egglog to this repo

I think it's better to fix the directory location first and update README and miscellaneous such as vscode/egglog-language-server/package.json later.


vscode/egglog-language-server/tree-sitter-egglog/src/tree_sitter/parser.h
```diff
@@ -0,0 +1,224 @@
+#ifndef TREE_SITTER_PARSER_H_
```
oflatt, line 1: Quick question before the egglog team takes a look at this: do we have to commit these header and c files? Can they be re-generated each time?

hatoo: I think it's possible to generate those files each time perhaps using custom `build.rs`. However, this makes it harder for people just need the parser only.
So, I think it's good to commit generated files and it seems a standard way in the tree-sitter community e.g. https://github.com/tree-sitter/tree-sitter-c

oflatt:
Thank you for making this PR @hatoo.
After consideration, the egglog team has decided that we don't have the cycles to merge and maintain this for now. We have several refactors of the code base that we want to do in the near term.

While we won't merge this, we've decided to make your repo the recommended egglog extension. Here's the issue I made:
https://github.com/egraphs-good/egglog/issues/312

Thank you so much for your excellent vscode extension, which we are using!

PR #306, opened 2023-12-01
Add `set-length` and `set-get`

rtjoa:
Without these, we can't walk over a set nor match its contents.

See [tests/set.egg](https://github.com/rtjoa/egglog/blob/set-length-and-get/tests/set.egg) for examples.

### DIFF
--- a/src/sort/set.rs
+++ b/src/sort/set.rs
@@ -60,6 +60,8 @@ impl SetSort {
             "set-union".into(),
             "set-diff".into(),
             "set-intersect".into(),
+            "set-get".into(),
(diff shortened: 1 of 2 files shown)


yihozhang:
I'm a little bit unsure about `set-get` since this seems un-idiomatic. In eggcc I had a user-defined `set->vec` primitive that can achieve something similar.

oflatt (approved):
Yihong and I think this is bad and inefficient, but the easiest thing for now.
In the future, we should either change these to index sets or write a set->vec function or both.

PR #317, opened 2023-12-06
Fix build on stable

### LINKED ISSUE #311: Build error, maybe involving `Rule` and `ast::Rule` conflict?
Hi all. Getting an error on a fresh build of egglog on GitHub Actions runners. Weirdly, cannot replicate on my own machine even after clean. Using stable Rust on my machine.

```
Compiling egglog v0.1.0 (https://github.com/egraphs-good/egglog?rev=41562802315eccc9136c9d2ae6967d2df4071420#41562802)
error[E0560]: struct `Rule` has no field named `head`
   --> /root/.cargo/git/checkouts/egglog-0aabc6dbd51d8ac4/4156280/src/terms.rs:391:13
    |
391 |             head: self.instrument_actions(&rule.head),
    |             ^^^^ `Rule` does not have this field
    |
    = note: available fields are: `query`, `program`, `matches`, `times_banned`, `banned_until`, `todo_timestamp`

error[E0560]: struct `Rule` has no field named `body`
   --> /root/.cargo/git/checkouts/egglog-0aabc6dbd51d8ac4/4156280/src/terms.rs:392:13
    |
392 |             body: self.instrument_facts(&rule.body),
    |             ^^^^ `Rule` does not have this field
    |
    = note: available fields are: `query`, `program`, `matches`, `times_banned`, `banned_until`, `todo_timestamp`

error[E0308]: mismatched types
    --> /root/.cargo/git/checkouts/egglog-0aabc6dbd51d8ac4/4156280/src/terms.rs:397:13
     |
397  |             rule,
     |             ^^^^ expected `ast::Rule`, found `Rule`
     |
     = note: `Rule` and `ast::Rule` have similar names, but are actually distinct types
note: `Rule` is defined in module `crate` of the current crate
    --> /root/.cargo/git/checkouts/egglog-0aabc6dbd51d8ac4/4156280/src/lib.rs:423:1
     |
423  | struct Rule {
     | ^^^^^^^^^^^
note: `ast::Rule` is defined in module `crate::ast` of the current crate
    --> /root/.cargo/git/checkouts/egglog-0aabc6dbd51d8ac4/4156280/src/ast/mod.rs:1244:1
     |
1[244](https://github.com/uwsampl/lakeroad/actions/runs/7073200166/job/19252861215#step:4:245) | pub struct Rule {
     | ^^^^^^^^^^^^^^^
```

hatoo:
fix #311 

There are 2 `Rule`s are defined in `crate::*`. Which are `crate::ast::Rule` and `crate::Rule`. And rustc is confused about it.

We should specify it clearly and I think it's good to remove `use crate::*;` and specify each `use`s

### DIFF
--- a/src/terms.rs
+++ b/src/terms.rs
@@ -1,4 +1,12 @@
-use crate::*;
+use crate::{
+    ast::{
+        desugar::Desugar, Action, Command, CommandId, Expr, Fact, NCommand, NormAction,
+        NormCommand, NormExpr, NormFact, NormFunctionDecl, NormRule, NormSchedule, Rule, RunConfig,
+        Schedule,
+    },
+    util::ListDisplay,
+    ArcSort, EGraph, Error, Symbol, TypeInfo,
+};
 
 /// The state of the term encoding is simply a current context
 /// so that it can look up type information.


saulshanabrook (commented):
This PR seems safe, but I am not that familiar with rust imports, so I am not sure if this is a good solution, so not sure if I should be the best one to review this, sorry!

From what I can see, there are two `Rule` structs defined, one in [`ast`](https://github.com/egraphs-good/egglog/blob/801b01817a3da128d595bfc87ff3b82d16866248/src/ast/mod.rs#L1244) and on in [`lib`](https://github.com/egraphs-good/egglog/blob/801b01817a3da128d595bfc87ff3b82d16866248/src/lib.rs#L423). 

In `lib`, the `Rule` from `ast` is imported through [the `use ast::*`](https://github.com/egraphs-good/egglog/blob/801b01817a3da128d595bfc87ff3b82d16866248/src/lib.rs#L42). As far as I understand it, this type [of "name shadowing" is allowed in Rust](https://github.com/rust-lang/rfcs/blob/master/text/1560-name-resolution.md#explicit-names-may-shadow-implicit-names).

Then when we go to the `use crate::*` statement in `terms.rs`, I am unsure which one it will pull in. I guess it ends up using the one from `ast`, maybe since that one is `pub` and the one in `lib` isn't?

So it would be nice to have a sense of why this bug arises in the first place, and if there is a more comprehensive fix for it, like adding a clippy check or renaming some things. If someone else feels like this is a good enough solution though for now, it also seems reasonable.

PR #320, opened 2023-12-08
Adding type annotations to intermediate representation

### LINKED ISSUE #80: Type Inference Failure

yihozhang:
This PR implements some important refactoring. At a high level, this PR gets rid of many legacy code and use a set of generic IRs which allows greater code sharing (thus reduced codebase size).

* Fully remove the old `NormExpr` and `NormAction`, use the new CoreIR instead.
* Generalizes front-end IR (e.g., `Rule`) and backend-IR (e.g., `CoreRule`) to be parametric over the head and leaf symbols.
* Add type resolution so that typechecking returns a type-annotated expression.
* Replace the type checker for actions to use the constraint-based one.
* Refactor the backend `Action->Program` compiler so that it no longer does unnecessary type checking and works with `CoreAction` instead of `Action`.
* `desugar` and `resugar` flags are replaced with a `--show desugared-egglog` flag, which prints an egglog program produced by the desugaring phase.

The current compilation workflow is

1. `Command -[desugar]-> NCommand`: does some lightweight desugaring, e.g., turns a function into a relation
2. `Rule in NCommand -[lower]-> CoreRule`: lowering to `CoreRule` for typechecking
3. `(Rule, CoreRule) -[typecheck]-> ResolvedRule`: `ResolvedRule` is a `Rule` with all its function symbols and variables type annotated.
4. Other planned transformations (e.g. proofs) over ResolvedRule
5. `ResolvedRule -[lower]-> ResolvedCoreRule`
6. `ResolvedCoreRule -[compilation]-> Program`

Fixes #80, #113, #196.


src/ast/expr.rs
```diff
@@ -140,23 +169,60 @@ impl Expr {
...
+            GenericExpr::Var(..) => f(self),
+            GenericExpr::Call(ann, op, children) => {
                 let children = children.iter().map(|c| c.map(f)).collect();
-                f(&Expr::Call(*op, children))
+                f(&GenericExpr::Call(ann.clone(), op.clone(), children))
+            }
+        }
+    }
+
+    // TODO: cannon function may want to take the annotation of variables
```
oflatt, line 181: what does this comment mean?

src/ast/mod.rs
```diff
@@ -1101,405 +978,340 @@ pub enum Action {
...
-            NormAction::LetVar(symbol, other) => {
-                NormAction::LetVar(fvar(*symbol, true), fvar(*other, false))
-            }
-            NormAction::LetLit(symbol, lit) => NormAction::LetLit(fvar(*symbol, true), lit.clone()),
-            NormAction::Set(expr, other) => {
-                NormAction::Set(expr.map_def_use(fvar, false), fvar(*other, false))
-            }
-            NormAction::Extract(var, variants) => {
-                NormAction::Extract(fvar(*var, false), fvar(*variants, false))
+    pub(crate) fn to_norm_actions<FG: FreshGen<Head, Leaf>>(
```
oflatt, line 1017: we should re-name this and add documentation
- explain why it returns both core actions and generic actions
- explain what the (Head, Leaf) is in the genericactions

src/ast/mod.rs
```diff
@@ -1101,405 +978,340 @@ pub enum Action {
...
-}
+pub type NormAction = CoreAction<Symbol, Symbol>;
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
-pub struct NormRule {
-    pub head: Vec<NormAction>,
-    pub body: Vec<NormFact>,
-}
+pub struct CoreActions<Head, Leaf>(pub(crate) Vec<CoreAction<Head, Leaf>>);
+pub(crate) type NormActions = CoreActions<Symbol, Symbol>;
```
oflatt, line 1169: do we use this anywhere? we could delete it

tests/files.rs
```diff
@@ -6,55 +6,34 @@ use libtest_mimic::Trial;
 #[derive(Clone)]
 struct Run {
     path: PathBuf,
-    test_proofs: bool,
-    resugar: bool,
-    test_terms_encoding: bool,
+    // test_proofs: bool,
```
oflatt, line 9: Lets not comment these out, just remove them for now

tests/files.rs
```diff
@@ -120,28 +99,28 @@ fn generate_tests(glob: &str) -> Vec<Trial> {
     for entry in glob::glob(glob).unwrap() {
         let run = Run {
             path: entry.unwrap().clone(),
-            test_proofs: false,
-            resugar: false,
-            test_terms_encoding: false,
+            // test_proofs: false,
+            // resugar: false,
```
oflatt, line 103: We should make re-sugaring work and make sure to test it- it's been very valuable in the past for catching bugs

src/typecheck.rs
```diff
@@ -1,172 +1,246 @@
 use std::ops::AddAssign;
```
oflatt, line 1: Can we pls rename this file to `query-compiler.rs` or something like this?
And move out a few of the typechecking related things to typechecking.rs

oflatt (changes requested):
Had some small ideas for how to improve this. Good stuff
I think now would be a great time to do a careful pass over the code and add documentation and comments.

src/ast/expr.rs
yihozhang: Currently, `subst_leaf` takes a leaf but not an annotation over the leaf, so it has to "make up" annotations for the returned `GenericExpr`. A better approach is for `subst_leaf` to also take the annotation, which we should implement after we use real non-`()` annotations

(also added the above explanation to the code)

src/ast/mod.rs
yihozhang: It's used in `typechecking.rs`. Do you think if it's a good idea to change this to the `GenericCoreActions/CoreActions/ResolvedCoreActions` schema?

yihozhang (commented):
Addressed some comments. I still need to make resugaring work.

src/ast/mod.rs
yihozhang: done

src/typecheck.rs
yihozhang: Renamed to `core.rs`, moved some core-related code to this file, and moved Action executor and compiler to a separate `action.rs`.

tests/files.rs
yihozhang: It works now!

src/ast/mod.rs
oflatt: Yeah that would make more sense

src/typecheck.rs
oflatt: Amazing change, thank you!

src/ast/desugar.rs
```diff
@@ -423,15 +109,14 @@ fn add_semi_naive_rule(desugar: &mut Desugar, rule: Rule) -> Option<Rule> {
 /// type_info field is used for that.
```
oflatt, line 109: This comment is not true- I don't think type_info actually has to live in desugar anymore.
It also doesn't keep track of global variables any more.

yihozhang: Great catch. I have moved `type_info` out of `Desugar`.

src/actions.rs
```diff
@@ -0,0 +1,412 @@
...
+    Union(usize),
+    Extract(usize),
+    Panic(String),
+}
+
+#[derive(Clone, Debug)]
+pub struct Program(Vec<Instruction>);
+
+impl EGraph {
+    /// Takes a list of variables bound to `subst` (variables bound during matching),
```
ezrosent, line 137: bound to `binding` ?

src/actions.rs
```diff
@@ -0,0 +1,412 @@
...
+    Global(Symbol),
+    // function to call, and whether to make defaults
+    CallFunction(Symbol, bool),
+    CallPrimitive(Primitive, usize),
+    DeleteRow(Symbol),
+    Set(Symbol),
+    Union(usize),
+    Extract(usize),
+    Panic(String),
+}
```
ezrosent, line 131: I think documentation for each of the variants here would be helpful. Some of these _seem_ self-explanatory, but a short sentence explaining what each does would still be helpful. 

E.g. I believe these are all instructions manipulating a stack, but how they manipulate it is worth explaining here.

src/actions.rs
```diff
@@ -0,0 +1,412 @@
...
+    fn perform_set(
+        &mut self,
+        table: Symbol,
+        new_value: Value,
+        stack: &mut Vec<Value>,
+    ) -> Result<(), Error> {
+        let function = self.functions.get_mut(&table).unwrap();
+
+        let new_len = stack.len() - function.schema.input.len();
+        // TODO would be nice to use slice here
```
ezrosent, line 202: What does this TODO mean? we are using a slice for `args`, yes?

src/actions.rs
```diff
@@ -0,0 +1,412 @@
...
+            types.insert(var.name, var.sort.clone());
+        }
+        let mut compiler = ActionCompiler {
+            egraph: self,
+            types: &types,
+            locals: IndexSet::default(),
+            instructions: Vec::new(),
+        };
+
+        for a in actions.0.iter() {
```
ezrosent, line 185: optional: any chance this is the same as `for a in &actions.0 { ...` ?

src/ast/desugar.rs
```diff
@@ -403,81 +89,57 @@ fn add_semi_naive_rule(desugar: &mut Desugar, rule: Rule) -> Option<Rule> {
...
-/// type_info field is used for that.
 pub struct Desugar {
     next_fresh: usize,
-    next_command_id: usize,
     // Store the parser because it takes some time
     // on startup for some reason
     parser: ast::parse::ProgramParser,
-    pub(crate) expr_parser: ast::parse::ExprParser,
-    pub(crate) action_parser: ast::parse::ActionParser,
+    // yz (dec 5): Comment out since they are only used in terms.rs which are deleted
```
ezrosent, line 107: Do we still need these comments? Should we just remove these fields for now?

src/ast/desugar.rs
```diff
@@ -403,81 +89,57 @@ fn add_semi_naive_rule(desugar: &mut Desugar, rule: Rule) -> Option<Rule> {
...
     fn default() -> Self {
-        let type_info = TypeInfo::default();
         Self {
             next_fresh: Default::default(),
-            next_command_id: Default::default(),
             // these come from lalrpop and don't have default impls
             parser: ast::parse::ProgramParser::new(),
-            expr_parser: ast::parse::ExprParser::new(),
-            action_parser: ast::parse::ActionParser::new(),
+            // expr_parser: ast::parse::ExprParser::new(),
```
ezrosent, line 120: (see other comment)

src/ast/desugar.rs
```diff
@@ -708,13 +346,10 @@ impl Clone for Desugar {
     fn clone(&self) -> Self {
         Self {
             next_fresh: self.next_fresh,
-            next_command_id: self.next_command_id,
             parser: ast::parse::ProgramParser::new(),
-            expr_parser: ast::parse::ExprParser::new(),
-            action_parser: ast::parse::ActionParser::new(),
+            // expr_parser: ast::parse::ExprParser::new(),
```
ezrosent, line 350: (same question)

src/ast/expr.rs
```diff
@@ -56,66 +58,93 @@ impl Display for Literal {
...
+    }
+}
+
+impl ToSexp for ResolvedVar {
+    fn to_sexp(&self) -> Sexp {
+        Sexp::Symbol(self.name.to_string())
+    }
+}
+
+pub type Expr = GenericExpr<Symbol, Symbol, ()>;
```
ezrosent, line 100: Add documentation here (ResolvedExpr too, but particularly for a `pub` type)

src/ast/expr.rs
```diff
@@ -56,66 +58,93 @@ impl Display for Literal {
...
+pub(crate) type ResolvedExpr = GenericExpr<ResolvedCall, ResolvedVar, ()>;
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
+pub enum GenericExpr<Head, Leaf, Ann> {
```
ezrosent, line 104: Add documentation explaining what a GenericExpr is, what the different type parameters mean, etc.

Also, consider giving this a shorter name (I don't think we need `Generic` in the name: anyone using it will see the type takes type parameters.). Maybe we can call this `Expr` and what is currently called `Expr` could be `RawExpr` or similar?

src/ast/expr.rs
```diff
@@ -56,66 +58,93 @@ impl Display for Literal {
...
+    Lit(Ann, Literal),
+    Var(Ann, Leaf),
```
ezrosent, line 106: optional:

We might consider adding helper functions for the case when `Ann` is `()`.  It could make some of the code with leading `()`s in all the data constructors a bit less busy. But on the other hand it may be less explicit. Up to you.

Something like:

```rust
impl<Head, Leaf> GenericExpr<Head, Leaf, ()> {
    /// construct a `Var` expression with a trivial annotation.
    pub(crate) fn var(leaf: Leaf) -> Self { GenericExpr::Var((), leaf) }
    pub(crate) fn lit(literal: Literal) ....
}
```

This would generalize some of the functions that we have on `Expr` further down.

src/ast/mod.rs
```diff
@@ -40,232 +43,110 @@ impl Display for Id {
...
-            ),
-        }
+impl<Head, Leaf, Ann> GenericSchedule<Head, Leaf, Ann> {
+    pub fn saturate(self) -> Self {
+        GenericSchedule::Saturate(Box::new(self))
     }
 }
 
-trait ToSexp {
+pub trait ToSexp {
```
ezrosent, line 149: Add some documentation here (the name is self-explanatory, but it'd be good to explain why this trait is important.

src/ast/mod.rs
```diff
@@ -764,13 +641,20 @@ impl Display for IdentSort {
...
-pub struct RunConfig {
+pub struct GenericRunConfig<Head, Leaf, Ann> {
     pub ruleset: Symbol,
-    pub until: Option<Vec<Fact>>,
+    pub until: Option<Vec<GenericFact<Head, Leaf, Ann>>>,
 }
 
-impl ToSexp for RunConfig {
+impl<Head: Display, Leaf: Display, Ann> ToSexp for GenericRunConfig<Head, Leaf, Ann>
+where
```
ezrosent, line 654: why do we need the `Head: Display`, `Leaf: Display` bounds twice?

src/ast/mod.rs
```diff
@@ -785,64 +669,18 @@ impl ToSexp for RunConfig {
...
-            schema: self.schema.clone(),
-            default: self.default.clone(),
-            merge: self.merge.clone(),
-            merge_action: self.merge_action.iter().map(|a| a.to_action()).collect(),
-            cost: self.cost,
-            unextractable: self.unextractable,
-        }
-    }
-}
+pub type FunctionDecl = GenericFunctionDecl<Symbol, Symbol, ()>;
```
ezrosent, line 672: Documentation (in particular, explain the difference between FunctionDecl and ResolvedFunctionDecl)

ezrosent (changes requested):
Minor comments (this is the first round. I think there's more to do but this may be a good place to start)

src/ast/mod.rs
```diff
@@ -959,127 +802,157 @@ impl ToSexp for FunctionDecl {
...
 
-impl Fact {
-    pub fn map_exprs(&self, f: &mut impl FnMut(&Expr) -> Expr) -> Fact {
+impl<Head, Leaf, Ann> GenericFact<Head, Leaf, Ann>
+where
+    Ann: Clone,
+    Head: Clone + Display,
+    Leaf: Clone + PartialEq + Eq + Display + Hash,
+{
+    pub(crate) fn map_exprs<Head2, Leaf2>(
```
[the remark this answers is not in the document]
yihozhang: It does not. This was because some legacy implementations of `Expr` require these things (without a good reason), and all code built on top just followed this. I removed these unnecessary dependencies here and at some other places.

src/ast/mod.rs
```diff
@@ -959,127 +802,157 @@ impl ToSexp for FunctionDecl {
...
-impl Display for NormFact {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.to_fact())
+impl<Head, Leaf> GenericFact<Head, Leaf, ()>
+where
+    Leaf: Clone + PartialEq + Eq + Display + Hash,
+    Head: Clone + Display,
+{
+    pub(crate) fn to_unresolved(&self) -> Fact
+    where
```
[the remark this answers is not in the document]
yihozhang: Otherwise it will err with the message "can't leak crate-private trait" since SymbolLike is a crate-private trait.

src/ast/mod.rs
```diff
@@ -1101,405 +978,162 @@ pub enum Action {
...
-
-        let mut unbound_vars = HashSet::<Symbol>::default();
-        for fact in facts {
-            fact.map_def_use(&mut |var, def| {
-                if !def && !bound_vars.contains(&var) {
-                    unbound_vars.insert(var);
-                }
-                var
-            });
+    pub fn map_def_use(&self, fvar: &mut impl FnMut(&Leaf, bool) -> Leaf) -> Self {
```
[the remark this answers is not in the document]
yihozhang: It is actually funny that every method in this `impl` are dead code if not `pub`. I left a note here. I think `map_def_use` is referred to in `term.rs` so it might be needed in the near future. All these functions are either from the very original codebase or from the `NormFact` refactor. I'm also happy to just delete all of them.

src/constraint.rs
```diff
@@ -216,13 +384,134 @@ where
...
+        }
+        Ok(())
+    }
+
+    pub(crate) fn add_rule(
+        &mut self,
+        rule: &CoreRule,
+        typeinfo: &TypeInfo,
+    ) -> Result<(), TypeError> {
+        let CoreRule { head, body } = rule;
```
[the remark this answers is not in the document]
yihozhang: Very nice!

src/core.rs
```diff
@@ -0,0 +1,755 @@
...
+
+impl ResolvedCall {
+    pub fn output(&self) -> &ArcSort {
+        match self {
+            ResolvedCall::Func(func) => &func.output,
+            ResolvedCall::Primitive(prim) => &prim.output,
+        }
+    }
+
+    // Different from `from_resolution`, this function only considers function types and not primitives.
```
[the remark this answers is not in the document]
yihozhang: Done; I changed them to pub(crate)

src/core.rs
```diff
@@ -0,0 +1,755 @@
...
+    // As a result, it only requires input argument types, so types.len() == func.input.len(),
+    // while for `from_resolution`, types.len() == func.input.len() + 1 to account for the output type
+    pub fn from_resolution_func_types(
+        head: &Symbol,
+        types: &[ArcSort],
+        typeinfo: &TypeInfo,
+    ) -> Option<ResolvedCall> {
+        if let Some(ty) = typeinfo.func_types.get(head) {
```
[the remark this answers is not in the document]
yihozhang: Done

src/core.rs
```diff
@@ -0,0 +1,755 @@
...
+                    resolved_call.push(ResolvedCall::Primitive(SpecializedPrimitive {
+                        primitive: primitive.clone(),
+                        input: inp.to_vec(),
+                        output: out.clone(),
+                    }));
+                }
+            }
+        }
+
+        assert!(resolved_call.len() == 1);
```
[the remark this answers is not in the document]
yihozhang: Done

src/core.rs
```diff
@@ -0,0 +1,755 @@
...
+                }
+            }
+        }
+        Ok((
+            GenericCoreActions::new(norm_actions),
+            GenericActions::new(mapped_actions),
+        ))
+    }
+}
+
```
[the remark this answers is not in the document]
yihozhang: I added a `MappedExpr` type (with brief documentation) as a shorthand

yihozhang (commented):
@ezrosent Thanks for the awesome review. The review is super helpful for improving this PR and I really appreciate it. I should have addressed most of your comments.

src/core.rs
```diff
@@ -0,0 +1,754 @@
...
+                        [GenericAtomTerm::Var(x), y] | [y, GenericAtomTerm::Var(x)] => {
+                            to_subst = Some((x, y));
+                            break;
+                        }
+                        _ => (),
+                    }
+                }
+            }
+            if let Some((x, y)) = to_subst {
+                let subst = HashMap::from_iter(vec![(x.clone(), y.clone())]);
```
ezrosent, line 651: nit: I don't think you need `vec!` here (you should be able to iterate over the array, or at least pass a `iter::once(`)

src/core.rs
```diff
@@ -0,0 +1,754 @@
...
+        &self,
+        typeinfo: &TypeInfo,
+        fresh_gen: impl FreshGen<Head, Leaf>,
+        value_eq: impl Fn(&GenericAtomTerm<Leaf>, &GenericAtomTerm<Leaf>) -> Head,
+    ) -> Result<GenericCoreRule<Head, Head, Leaf>, TypeError>
+    where
+        Leaf: SymbolLike,
+    {
+        let rule = self.to_core_rule(typeinfo, fresh_gen)?;
+        let rule = rule.canonicalize(value_eq);
```
ezrosent, line 734: nit, optional: I think `Ok(rule.canonicalize(value_eq))` is a little more readable here.

src/lib.rs
```diff
@@ -19,8 +19,9 @@ mod gj;
 mod serialize;
 pub mod sort;
 mod termdag;
-mod terms;
-mod typecheck;
+// mod terms;
```
ezrosent, line 22: remove this? (and `use terms::TermState`)

src/lib.rs
```diff
@@ -1210,125 +1207,94 @@ impl EGraph {
...
-                    NormCommand {
-                        metadata,
-                        command: *c,
-                    },
-                    should_run,
-                );
+            ResolvedNCommand::Fail(c) => {
+                let result = self.run_command(*c);
                 if let Err(e) = result {
                     log::info!("Command failed as expected: {}", e);
```
ezrosent, line 1289: nit: does the `"Command failed as expected: {e}"`syntax work?

src/lib.rs
```diff
@@ -1395,28 +1362,33 @@ impl EGraph {
...
+        let num_facts = actions.len();
+        let commands = actions
+            .into_iter()
+            .map(NCommand::CoreAction)
+            .collect::<Vec<_>>();
+        let commands: Vec<_> = self.type_info_mut().typecheck_program(&commands)?;
+        for command in commands {
+            self.run_command(command)?;
+        }
+        log::info!("Read {} facts into {func_name} from '{file}'.", num_facts);
```
ezrosent, line 1391: same question: `{num_facts}`

src/lib.rs
```diff
@@ -1426,89 +1398,48 @@ impl EGraph {
...
-        let should_run = true;
-
         for command in program {
             // Important to process each command individually
             // because push and pop create new scopes
-            for processed in self.process_command(command, CompilerPassStop::All)? {
-                self.run_command(processed, should_run)?;
+            for processed in self.process_command(command)? {
+                if self.run_mode.show_egglog() {
+                    // This is a hack: we have to do this because process_commands are indeed
```
ezrosent, line 1423: Having trouble following what's happening here.

My read here is that `show_egglog()` is essentially a "dry run" mode where the only commands we want to run are the ones that we need to make everything well-scoped (Push/Pop). 

But then how does that relate to typechecking a function twice? When would we do this twice? And aren't we still typechecking in `process_command?` ?

src/main.rs
```diff
@@ -109,67 +110,49 @@ fn main() {
...
+                                    line_num + 1,
+                                    // TODO: Show utf8 aware character count
+                                    byte_offset - sum_offset + 1
+                                )
+                            }
+                        })
+                        .to_string(),
+                    err => err.to_string(),
+                };
+                log::error!("{}", err);
```
ezrosent, line 154: `log::error!("{err}")`

src/sort/mod.rs
```diff
@@ -145,3 +146,29 @@ impl<T: IntoSort> IntoSort for Option<T> {
 
 pub type PreSort =
     fn(typeinfo: &mut TypeInfo, name: Symbol, params: &[Expr]) -> Result<ArcSort, TypeError>;
+
+pub(crate) struct ValueEq {
+    pub unit: Arc<UnitSort>,
+}
+
+impl PrimitiveLike for ValueEq {
+    fn name(&self) -> Symbol {
```
ezrosent, line 155: optional: 

A while back, I noticed some performance problems related to calling "string constant".into() all over the place. It was faster / cheaper to have a lazy_static! that cached the intern value and then loading that here.

ezrosent (approved):

src/lib.rs
yihozhang: You are right. I was specifically talking about "typechecking a function twice" because, in our test, we have code like
```
(sort Expr)
(push)
(function f () Expr)
(pop)
(push)
(function f () Expr)
(pop)
```
It is only when `(push)` and `(pop)` are run that this does not lead to an error. But I agree that "typechecking twice" is not fundamental to the issue here. I'll update the comment to reflect this.

src/sort/mod.rs
yihozhang: Done!

oflatt (approved):
Ready to merge! Amazing work 🦸

PR #321, opened 2023-12-10
fix egglog python link

majidaldo:


saulshanabrook (approved):
Thanks for this. I had updated the docs to default to English, removing the `/en/` path prefix.

PR #322, opened 2023-12-12
Fix clippy errors and extraction in yihong's typechecking branch

oflatt:


yihozhang:
Great work! ALL tests are passing :) There are a number of places where the type can be made more specialized/concise (e.g., `Expr::xxx` vs `GenericExpr::xxx` and `GenericExpr<Symbol, Symbol, ()>` vs `Expr`), which I'll be working on in my branch.

PR #330, opened 2024-01-13
Typechecker tests for variable shadowing 

oflatt:
The local-global test actually currently fails, but passes on #320


tests/shadow-global-constructor.egg
```diff
@@ -0,0 +1,4 @@
+(datatype MySort
+  (Wrap i64))
+
+(let Wrap 3)
```
saulshanabrook, line 4: lol this is a fun one! 😬

saulshanabrook: I assume we want this one and the next one to fail ideally? seems a bit worrisome to allow this kind of shadowing if we disallow the other kind

oflatt:
Looks like we fail a test and we are not strict enough on shadow-global-constructor
Leaving this up for now until fixed

Alex-Fischman:
Reviving this because we just merged #565.

Alex-Fischman:
I think these tests should be in Rust so that we can check that the right error message is returned.

Alex-Fischman:
Subsumed by #590.

Issue #335, opened 2024-01-31
error [E0659] when `make all`


cyk2018:
when I was installing egglog, I clone this reposity and I install nextest through cargo. when I use "make all", this information is shown.

```bash
$ cargo nextest run --release
   Compiling egglog v0.1.0 (/home/test/cyk/egglog)
error[E0659]: `core` is ambiguous
  --> src/lib.rs:46:5
   |
46 | use core::{AtomTerm, ResolvedCall};
   |     ^^^^ ambiguous name
   |
   = note: ambiguous because of multiple potential import sources
   = note: `core` could refer to a built-in crate
   = help: use `::core` to refer to this crate unambiguously
note: `core` could also refer to the module defined here
  --> src/lib.rs:17:1
   |
17 | mod core;
   | ^^^^^^^^^
   = help: use `crate::core` to refer to this module unambiguously

For more information about this error, try `rustc --explain E0659`.
error: could not compile `egglog` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `egglog` (lib test) due to previous error
error: command `/usr/bin/cargo test --no-run --message-format json-render-diagnostics --release` exited with code 101
make: *** [Makefile:16: test] Error 101
```

does anyone known how to fix it?

oflatt:
I can't seem to reproduce but I can solve the problem probably
Here's the PR:
https://github.com/egraphs-good/egglog/pull/336
Lmk if it works! If you have any idea why I can't reproduce that would be interesting

cyk2018:
> I can't seem to reproduce but I can solve the problem probably Here's the PR: #336 Lmk if it works! If you have any idea why I can't reproduce that would be interesting

Thank you for your quick reply. I have try your method and fix this error. It seems in the long information I have compiled it successfully. In the end, 
make: rustup: No such file or directory
make: *** [Makefile:21: nits] Error 127
is shown. I think this doesn't matter, but do you have some idea about this.

About why you can not reproduce, maybe I can only give you some information as a cpp player. First I have no Rust tool and environment in my computer. The only work is I install cargo through apt. and my cargo version is cargo 1.71.1, my make version is GNU Make 4.3. I use this tool in wsl. my wsl version is 2.0.9.0.
When I search this in website. I find core is one mod in standard lib, I think does this problem you can not reproduce is that you have install some Rust tool in your computer?

Later, I will try this again in my ubuntu laptop, I think it will show that wsl doesn't matter.

[...]

PR #336, opened 2024-02-01
Fix ambiguity in `core` module

oflatt:
Addresses #335


saulshanabrook (approved):
I believe @cyk2018 mentioned that this PR changed fixed the build locally for them in https://github.com/egraphs-good/egglog/issues/335#issuecomment-1920299031 so it seems good.

I am not familiar enough with rust modules to understand what the issue was or why it fixed it.

yihozhang:
Probably because Rust also has [a `core` module](https://doc.rust-lang.org/core/) (which is the basis of `std`)

Issue #334, opened 2024-01-30
Extracted Declarations are Temporary Variables


saulshanabrook:
Currently, declarations are a bit odd since they cannot be extracted. Or they can, but what you end up with is a temporary variable that it's unclear how it relates to the original declaration. For example:

```
(datatype Hi)
(declare x Hi)
(extract x)
# (v0___)
```

## Why

This is how `declare` is currently desugared:

https://github.com/egraphs-good/egglog/blob/b78f69ca1f7187c363bb31271c8e8958f477f15d/src/ast/desugar.rs#L392-L409

It makes a fresh function (which is what is extracted) and creates a `let` binding with that name to the function.

There @yihozhang had a previous PR that changes this behavior to make declare's unextractable (https://github.com/egraphs-good/egglog/pull/141) but it was closed.

Personally, it would be nice if there would be a way to make them extractable. In the Python bindings, I compile the "constant" (aka a "declare") into a nullary function, but I make the function name the name that the user passes in and return it called. Then, on pretty printing, I just show it as a constant, instead of as a function call. However, if I could compile directly to the `declare` that would be nice!

oflatt:
Good catch
I see two solutions:
1. Get rid of declare all together because it's annoying and currently broken
2. Make declare actually work as you mentioned, by transforming all code after it to refer to a nullary function

I'm currently working on desugaring all globals to use nullary functions again internally
We could desugar declare similarly

oflatt:
I'm actually in favor of getting rid of declare, since it causes confusion.

yihozhang:
`declare` is convenient for declaring constants like `True` and `False` (similar to `declare-const` in z3).

IMO it might be more consistent to desugar both `declare` and `let` to extractable `function` with a special flag `can_be_used_without_parentheses` (I'm sure we can come up with a shorter name). This is also aligned with our goal to remove globals from the backend implementation.

saulshanabrook:
> IMO it might be more consistent to desugar both declare and let to extractable function

That sounds good to me, but with `let` functions being marked unextractable.

saulshanabrook:
That sounds good. Previously I had the types of each e-class as a tag visually, so I don't think it should be too hard to get the tag(s) to show up there if we can get them in the serialized format somehow.

oflatt:
I don't like the idea of `can_be_used_without_parenthesis`
[...]

PR #338, opened 2024-02-07
Desugar global variables away

oflatt:
This PR adds a brand new, type-preserving pass that gets rid of all global variables by tuning them into functions.
This does not yet resolve the issue about declare: #334 
Example:
```
(let x 3)
(Add x x)
 ```

becomes

```
(function x () i64)
(set (x) 3)
(Add (x) (x))
```

Most of this PR is helpers, such as `map_exprs`, which allows us to map over all exprs in the whole egglog program.

Here are some questions we need to resolve before this is merged:
- [ ] While refactoring `map_exprs`, I had to add a lot of type constraints throughout the ast. It's a lot of boilerplate- is there a way to get rid of them?
- [x] How does this affect visualizations? Right now the function is unextractable, but probably still shows up in the viz which is incorrect
- [ ] Can we remove `AtomTerm::Global`? Seems like the typechecker uses core actions so maybe not? Right now we have several panics in the back end when we see a global.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -168,10 +168,14 @@ where
                 ruleset,
                 rule: rule.map_exprs(f),
             },
-            GenericNCommand::RunSchedule(schedule) => GenericNCommand::RunSchedule(schedule),
+            GenericNCommand::RunSchedule(schedule) => {
+                GenericNCommand::RunSchedule(schedule.map_exprs(f))
+            }
             GenericNCommand::PrintOverallStatistics => GenericNCommand::PrintOverallStatistics,
             GenericNCommand::CoreAction(action) => GenericNCommand::CoreAction(action.map_exprs(f)),
-            GenericNCommand::Check(facts) => GenericNCommand::Check(facts),
+            GenericNCommand::Check(facts) => {
+                GenericNCommand::Check(facts.into_iter().map(|fact| fact.map_exprs(f)).collect())
+            }
             GenericNCommand::CheckProof => GenericNCommand::CheckProof,
             GenericNCommand::PrintTable(name, n) => GenericNCommand::PrintTable(name, n),
             GenericNCommand::PrintSize(name) => GenericNCommand::PrintSize(name),
@@ -242,6 +246,31 @@ macro_rules! list {
         let list: Vec<Sexp> = vec![ $($e.to_sexp(),)* ];
         Sexp::List(list)
     }};
+}
+
+impl<Head, Leaf, Ann> GenericSchedule<Head, Leaf, Ann>
+where
+    Ann: Clone + Default,
+    Head: Clone + Display,
+    Leaf: Clone + PartialEq + Eq + Display + Hash,
+{
+    fn map_exprs(
+        self,
+        f: &mut impl FnMut(GenericExpr<Head, Leaf, Ann>) -> GenericExpr<Head, Leaf, Ann>,
+    ) -> Self {
+        match self {
+            GenericSchedule::Saturate(sched) => {
+                GenericSchedule::Saturate(Box::new(sched.map_exprs(f)))
+            }
+            GenericSchedule::Repeat(size, sched) => {
+                GenericSchedule::Repeat(size, Box::new(sched.map_exprs(f)))
+            }
+            GenericSchedule::Run(config) => GenericSchedule::Run(config.map_exprs(f)),
+            GenericSchedule::Sequence(scheds) => {
+                GenericSchedule::Sequence(scheds.into_iter().map(|s| s.map_exprs(f)).collect())
+            }
+        }
+    }
 }
 
 impl<Head: Display, Leaf: Display, Ann> ToSexp for GenericSchedule<Head, Leaf, Ann> {
@@ -721,6 +750,25 @@ pub struct GenericRunConfig<Head, Leaf, Ann> {
     pub until: Option<Vec<GenericFact<Head, Leaf, Ann>>>,
 }
 
+impl<Head, Leaf, Ann> GenericRunConfig<Head, Leaf, Ann>
+where
+    Ann: Clone + Default,
+    Head: Clone + Display,
+    Leaf: Clone + PartialEq + Eq + Display + Hash,
+{
+    pub fn map_exprs(
+        self,
+        f: &mut impl FnMut(GenericExpr<Head, Leaf, Ann>) -> GenericExpr<Head, Leaf, Ann>,
+    ) -> Self {
+        Self {
+            ruleset: self.ruleset,
+            until: self
+                .until
+                .map(|until| until.into_iter().map(|fact| fact.map_exprs(f)).collect()),
+        }
+    }
+}
+
 impl<Head: Display, Leaf: Display, Ann> ToSexp for GenericRunConfig<Head, Leaf, Ann>
 where
     Head: Display,
@@ -991,7 +1039,9 @@ where
         f: &mut impl FnMut(GenericExpr<Head, Leaf, Ann>) -> GenericExpr<Head, Leaf, Ann>,
     ) -> GenericFact<Head, Leaf, Ann> {
         match self {
-            GenericFact::Eq(exprs) => GenericFact::Eq(exprs.into_iter().map(f).collect()),
+            GenericFact::Eq(exprs) => {
+                GenericFact::Eq(exprs.into_iter().map(|expr| expr.map(f)).collect())
+            }
             GenericFact::Fact(expr) => GenericFact::Fact(expr.map(f)),
         }
     }
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -21,6 +21,7 @@ use crate::{
 mod expr;
 pub use expr::*;
 pub mod desugar;
+mod remove_globals;
 
 #[derive(Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord, Debug)]
 pub struct Id(usize);
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -1266,11 +1266,11 @@ where
     }
 }
 
-impl<Head: Display + ToSexp, Leaf: Display + ToSexp, Ann> ToSexp for GenericAction<Head, Leaf, Ann>
+impl<Head, Leaf, Ann> ToSexp for GenericAction<Head, Leaf, Ann>
 where
     Ann: Clone + Default,
-    Head: Clone + Display,
-    Leaf: Clone + PartialEq + Eq + Display + Hash,
+    Head: Clone + Display + ToSexp,
+    Leaf: Clone + PartialEq + Eq + Display + Hash + ToSexp,
 {
     fn to_sexp(&self) -> Sexp {
         match self {
@@ -1346,6 +1346,7 @@ where
             }
             // TODO should we refactor `Set` so that we can map over Expr::Call(lhs, args)?
             // This seems more natural to oflatt
+            // Currently, visit_exprs does not apply f to the first argument of Set.
             GenericAction::Set(ann, lhs, args, rhs) => {
                 let args = args.into_iter().map(|e| e.visit_exprs(f)).collect();
                 GenericAction::Set(ann.clone(), lhs.clone(), args, rhs.visit_exprs(f))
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

**Comment by oflatt** [comment]
Assigning @yihozhang to review, added @saulshanabrook in case they wanted to take a look

**Comment by saulshanabrook** [comment]
Is there a feature/bug this resolves or is it mainly meant to make the implementation cleaner?

**Comment by oflatt** [comment]
It just makes it cleaner 
Also will help with adding proofs

File: src/ast/mod.rs [hunk 1]
```diff
@@ -753,8 +865,33 @@ impl FunctionDecl {
...
+where
+    Ann: Clone + Default,
+    Head: Clone + Display,
+    Leaf: Clone + PartialEq + Eq + Display + Hash,
+{
+    pub fn map_exprs(
+        self,
+        f: &mut impl FnMut(GenericExpr<Head, Leaf, Ann>) -> GenericExpr<Head, Leaf, Ann>,
+    ) -> GenericFunctionDecl<Head, Leaf, Ann> {
+        GenericFunctionDecl {
```
yihozhang, line 878: Using [this syntax](https://doc.rust-lang.org/book/ch05-01-defining-structs.html#creating-instances-from-other-instances-with-struct-update-syntax) might make this (and a few other places) look cleaner

src/ast/mod.rs
```diff
@@ -893,26 +1030,44 @@ where
...
-    Ann: Clone,
+    Ann: Clone + Default,
     Head: Clone + Display,
     Leaf: Clone + PartialEq + Eq + Display + Hash,
 {
-    pub(crate) fn map_exprs<Head2, Leaf2>(
-        &self,
-        f: &mut impl FnMut(&GenericExpr<Head, Leaf, Ann>) -> GenericExpr<Head2, Leaf2, Ann>,
-    ) -> GenericFact<Head2, Leaf2, Ann> {
+    pub(crate) fn map_exprs(
```
yihozhang, line 1037: This now seems to be a very different function than the original declaration. Should we keep the original method, which also seems useful? The proposed `map_exprs` seems more like `normalize` in the sense it does a bottom up normalizing pass.

src/ast/mod.rs
```diff
@@ -893,26 +1030,44 @@ where
...
-            GenericFact::Eq(exprs) => GenericFact::Eq(exprs.iter().map(f).collect()),
-            GenericFact::Fact(expr) => GenericFact::Fact(f(expr)),
+            GenericFact::Eq(exprs) => {
+                GenericFact::Eq(exprs.into_iter().map(|expr| expr.map(f)).collect())
+            }
+            GenericFact::Fact(expr) => GenericFact::Fact(expr.map(f)),
         }
     }
 
-    pub(crate) fn subst<Leaf2, Head2>(
```
yihozhang, line 910: Same question here. The original `subst` method seems more powerful, and it is very common to substitute a leaf (variable) with an expression (this is how the substitution works for lambda calculus), while map_leafs is more like renaming than substitution. Could we still do what the original `subst` did in the new proposed interface?

src/ast/mod.rs
```diff
@@ -942,12 +1096,47 @@ where
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct CorrespondingVar<Head, Leaf>
```
yihozhang, line 1100: Having a struct for CorrespondingVar is a great idea! 

Optional: the name is a little bit long, can we think of a simpler name? E.g., `WithWitness`?

src/ast/mod.rs
```diff
@@ -1048,45 +1263,45 @@ where
...
-        f: &mut impl FnMut(&GenericExpr<Head, Leaf, Ann>) -> GenericExpr<Head, Leaf, Ann>,
+        self,
+        f: &mut impl FnMut(GenericExpr<Head, Leaf, Ann>) -> GenericExpr<Head, Leaf, Ann>,
     ) -> Self {
         match self {
             GenericAction::Let(ann, lhs, rhs) => {
-                GenericAction::Let(ann.clone(), lhs.clone(), f(rhs))
+                GenericAction::Let(ann.clone(), lhs.clone(), rhs.map(f))
             }
+            // TODO should we refactor `Set` so that we can map over Expr::Call(lhs, args)?
```
yihozhang, line 1276: I like `Set` to have an `lhs` and `args` more so that the constraints that the first argument to `Set` is not a literal nor variable is enforced. It is just a good practice to make illegal states unrepresentable. Here, we essentially ensure that the substituted result produced by `f` is not a literal/variable.

yihozhang (changes requested):
Thanks for making this PR! Getting rid of global variables would definitely help make our lives easier in the future.

I made a pass through the interface changes to `map_exprs` and other related functions, and will make another pass through the rest of the PR soon. 

I have some concerns about the proposed changes though, mostly because it cuts functionalities from before and it is not clear to me that the new interface subsumes the old one in terms of generality. In the future it would be great for us to agree on some interface/semantic change plans before we make any commitment, but great work!

src/ast/mod.rs
oflatt: keeping as is for now

src/ast/mod.rs
oflatt: can't think of a better one

src/ast/mod.rs
oflatt: added old subst

src/ast/mod.rs
oflatt: added old map_exprs, re-named mine to visit

src/ast/mod.rs
oflatt: True, but then you might forget to update this code if things change in the future
Keeping as is for now

oflatt:
This PR is ready to merge
The two other remaining TODOs are optional

yihozhang (changes requested):
Just had another pass. This is great and the code is easy to follow. Besides questions in the comments, I have a higher-level question:

* Have you thought about removing global variables during the desugaring phase before the type checking? This way `ResolvedVar` does not need to carry an `is_global` field around, which is only useful for typechecking and globals removal. 
  * If we can do this, then the typechecking and passes after this will only be dealing with local variables, and we can entirely remove `AtomTerm::Global`.
  * Instead of having `TypeInfo` to track the set of global variables, we can have `Desugar` responsible for managing a set of globals to be desugared.

oflatt:
> Just had another pass. This is great and the code is easy to follow. Besides questions in the comments, I have a higher-level question:
> 
> * Have you thought about removing global variables during the desugaring phase before the type checking? This way `ResolvedVar` does not need to carry an `is_global_ref` field around. Currently, this field makes sense only for the typechecking and the globals removal passes.
>   
>   * If we can do this, then the typechecking and passes after this will only be dealing with local variables, and we can entirely remove `AtomTerm::Global`.
>   * Instead of having `TypeInfo` to track the set of global variables, we can have `Desugar` responsible for managing a set of globals to be desugared.

I've thought of this, and I agree it would be much nicer. The problem is, this pass actually does depend on type information: it needs to know the type of the global being desugared.

src/ast/remove_globals.rs
```diff
@@ -0,0 +1,107 @@
...
+                };
+                vec![
+                    GenericNCommand::Function(func_decl),
+                    GenericNCommand::CoreAction(GenericAction::Set(
+                        ann,
+                        ResolvedCall::Func(FuncType {
+                            name: name.name,
+                            input: vec![],
+                            output: ty,
+                            is_datatype: false,
```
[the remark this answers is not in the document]
oflatt: That's a good point, but this isn't a datatype because we are using `Set` on it.
You can think of it having a merge function of panic, since we only set it once.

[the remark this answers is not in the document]
oflatt: Using `Set` instead of union is nice because we don't have to touch the equivalence relation, and it works for closed datatypes if we decide to add them

src/typechecking.rs
```diff
@@ -184,6 +184,8 @@ impl TypeInfo {
                 let var = ResolvedVar {
                     name: *var,
                     sort: output_type,
+                    // not a global reference, but a global binding
+                    is_global_ref: false,
```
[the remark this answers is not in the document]
oflatt: Hmm that's true. Maybe we should override equality? Or just keep it in mind?

src/ast/remove_globals.rs
```diff
@@ -0,0 +1,107 @@
...
+pub(crate) fn remove_globals(
+    type_info: &TypeInfo,
+    prog: Vec<ResolvedNCommand>,
+) -> Vec<ResolvedNCommand> {
+    prog.into_iter()
+        .flat_map(|cmd| remove_globals_cmd(type_info, cmd))
+        .collect()
+}
+
+fn replace_global_var(expr: ResolvedExpr) -> ResolvedExpr {
```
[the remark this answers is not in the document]
oflatt: I would have to implement `subst` on commands, which is a lot more work... I'll add a todo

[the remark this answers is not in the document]
oflatt: `visit_exprs` is also more efficient actually because it uses an owned value

src/ast/mod.rs
```diff
@@ -687,14 +796,22 @@ pub(crate) type ResolvedFunctionDecl = GenericFunctionDecl<ResolvedCall, Resolve
...
     pub name: Symbol,
     pub schema: Schema,
     pub default: Option<GenericExpr<Head, Leaf, Ann>>,
     pub merge: Option<GenericExpr<Head, Leaf, Ann>>,
     pub merge_action: GenericActions<Head, Leaf, Ann>,
     pub cost: Option<usize>,
     pub unextractable: bool,
+    /// Globals are desugared to functions, with this flag set to true.
+    /// This is used by visualization to handle globals differently.
+    pub is_global: bool,
```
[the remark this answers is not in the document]
oflatt: Replaced with `ignore_viz`

src/ast/remove_globals.rs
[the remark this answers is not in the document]
yihozhang: We can possibly make `subst` also own its argument, but a TODO sounds good! After all this is just a minor implementation detail (although taste-wise I tend to think `subst` as being slightly more principled).

oflatt:
@yihozhang is this ready to merge?

src/ast/remove_globals.rs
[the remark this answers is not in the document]
yihozhang: Wait, if globals are not datatypes, since we only allow looking up datatypes in actions, this means we can't refer to globals in actions?

yihozhang (approved):
I see some of the nit comments are marked resolved but not reflected in the code, so I went ahead fixing them and pushing a commit to your branch, assuming you were planning to address those nits. I also added some small clarification comments from our discussion. So if my commit looks good to you, we can merge this PR 🎉 (also see my question above)

oflatt:
Thanks for fixing up, I must have lost the commit where I fixed those

src/ast/remove_globals.rs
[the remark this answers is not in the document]
oflatt: That's a great point.
I used `set` so that this works with primitives and containers.
One thing I could do is inline primitives but this is inefficient for containers.
I could also add them to the query instead...

oflatt:
@yihozhang I've refactored this PR to not look up values in actions.
It adds to the query instead.
I'm going to refactor generating fresh symbols next.

src/ast/remove_globals.rs
```diff
@@ -25,88 +32,176 @@ use crate::{
...
+/// (rule ((Neg y)
+///        (= fresh_var_for_x (x)))
+///       ((Add fresh_var_for_x fresh_var_for_x)))
+/// ```
 pub(crate) fn remove_globals(
     type_info: &TypeInfo,
     prog: Vec<ResolvedNCommand>,
 ) -> Vec<ResolvedNCommand> {
+    let mut remover = GlobalRemover {
+        fresh: SymbolGen::new(),
```
yihozhang, line 52: This seems dangerous, since the counters are not shared across `SymbolGen`s so you may accidentally get conflict names.

yihozhang (approved):
I left a comment on SymbolGen but otherwise this looks great!

src/ast/remove_globals.rs
oflatt: Will fix soon! There are other problems with the symbolgen to address at the same time

PR #342, opened 2024-02-12
Remove terms file

### LINKED ISSUE #341: Terms file is not included
I believe that the `terms.rs` file is no longer included in the source tree. Maybe it should be removed?

saulshanabrook:
Closes #341 as a follow up from #320


oflatt (approved):
Thanks!

PR #344, opened 2024-02-15
Fix queries with no variables firing without checking query

### LINKED ISSUE #343: A rule testing a function evaluation against a literal always matches
The following program:
```
(function foo () i64 :merge (min old new))

(rule ((= (foo) 5)) ((set (foo) 4)))

(set (foo) 10)

(run 3)

(query-extract (foo))
```
prints `4`, meaning that the rule was matched. This seems unexpected, to me the rule could only match if the condition was `(= (foo) 10)`.

Strangely, replacing `5` in the condition by `(+ 1 4)` now gives the expected behavior (i.e. the rule does not match).

It is as if `5` was treated as a free variable in this case, instead of referring to the literal `5` (this might not make any sense, I have 0 knowledge of egglog's internals).

Am I missing something?

Thanks!

oflatt:
Fixes #343, which demonstrated that queries without variables always match. This PR fixes it by handling the edge case, also adding a `SearchResult` abstraction.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -473,7 +473,7 @@ pub struct NotFoundError(Expr);
 /// For each rule, we produce a `SearchResult`
 /// storing data about that rule's matches.
 /// When a rule has no variables, it may still match- in this case
-/// the `matched` field is used.
+/// the `did_match` field is used.
 struct SearchResult {
     name: Symbol,
     all_matches: Vec<Value>,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -470,6 +470,17 @@ impl Default for EGraph {
 #[error("Not found: {0}")]
 pub struct NotFoundError(Expr);
 
+/// For each rule, we produce a `SearchResult`
+/// storing data about that rule's matches.
+/// When a rule has no variables, it may still match- in this case
+/// the `matched` field is used.
+struct SearchResult {
+    name: Symbol,
+    all_matches: Vec<Value>,
+    did_match: bool,
+    rule_search_time: Duration,
+}
+
 impl EGraph {
     /// Use the rust backend implimentation of eqsat,
     /// including a rust implementation of the union-find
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -470,6 +470,17 @@ impl Default for EGraph {
 #[error("Not found: {0}")]
 pub struct NotFoundError(Expr);
 
+/// For each rule, we produce a `SearchResult`
+/// storing data about that rule's matches.
+/// When a rule has no variables, it may still match- in this case
+/// the `matched` field is used.
```
saulshanabrook, line 476: Do you mean the `do_match` field?

src/lib.rs
```diff
@@ -936,13 +947,15 @@ impl EGraph {
...
+            let mut all_matches = vec![];
             if rule.banned_until <= iteration {
                 let mut fuel = safe_shl(match_limit, rule.times_banned);
                 let rule_search_start = Instant::now();
+                let mut did_match = false;
                 self.run_query(&rule.query, rule.todo_timestamp, |values| {
+                    did_match = true;
                     assert_eq!(values.len(), rule.query.vars.len());
-                    all_values.extend_from_slice(values);
+                    all_matches.extend_from_slice(values);
```
saulshanabrook, line 958: Is `did_match` the same as `all_matches` being non empty?

saulshanabrook (commented):
Thanks for fixing this weird bug! Just had a few small questions

src/lib.rs
oflatt: yes, fixed

src/lib.rs
oflatt: No- since `all_matches` is a flat vector of matches, when there are no variables it is always empty.

saulshanabrook: Cool, that makes sense!

saulshanabrook (approved):

Issue #345, opened 2024-02-16
Serlalization not working as expected


cgjohannsen:
I'm trying to use egglog to do some optimizations on some fairly large benchmarks and the outputted JSON isn't as expected. For example, the datatype I'm using has an `"And"` variant that should always have two children, but in the JSON, many of these ENodes have no children at all. Also, I'm not sure if the `"root_eclasses"` field is ever populated, which would be useful if it were.

I'm not sure relying on the JSON serialization is the smartest route in the first place, but since I have a fairly complex cost function, dumping the JSON and doing the manually seemed like the easiest option.

Any help would be appreciated, thanks!

File to reproduce (with rules removed):  
[mvp.txt](https://github.com/egraphs-good/egglog/files/14305992/mvp.txt)

Command:

     cd egglog
     cargo run mvp.txt --to-json

cgjohannsen:
Here is a much smaller example that reproduces the issue (with the last expression not being serialized or constructed correctly)
[mvp_2.txt](https://github.com/egraphs-good/egglog/files/14311398/mvp_2.txt)

saulshanabrook:
Hey @cgjohannsen I believe the issue is that our default serialization configuration sets a max number of nodes to be output, so that the tests could run in a reasonable amount of time for large graphs:

https://github.com/egraphs-good/egglog/blob/76cb253a7f076b4690e62f1405bb42aa37c1448d/src/serialize.rs#L25-L29

Sorry about that! I think it will make sense to update it to set the default to be unlimited and then just make sure for our tests we limit it where we need.

cgjohannsen:
Great! Thanks so much for the workaround, that seems to work perfectly now with those bounds increased. 

Also, am I correct that the `root_eclasses` field is never populated? 

https://github.com/egraphs-good/egglog/blob/73a76dc9228ce24291b3e41bbbca14eddd10cd5e/src/main.rs#L173

oflatt:
`root_eclasses` not being populated is a known problem, we plan to add it soon!

saulshanabrook:
`root_eclasses` should be populated now. Please re-open or create a new issue if you have further problems here.

PR #347, opened 2024-02-16
Switch default serialization to output all nodes

saulshanabrook:
Still keeps visualization defaults to limit for testing and to make more readable graphviz files.

Partially addresses #345


oflatt (approved):
lgtm

PR #348, opened 2024-02-16
First class / higher order functions

saulshanabrook:
This PR is an attempt to support higher-order functions in egglog.

Compared to the earlier approach, in #155, this doesn't embed the bodies of unevaluated functions in the EGraph. Instead, all function values are just represented as lookups to egglog functions. An advantage of this approach is that we don't need to re-implement function application in anyway, we can just re-use our existing implementation. For defining the body of functions, we can just use rewrite rules.

Screenshot 2024-02-16 at 6 15 09 PM

This was inspired by [@yihozhang's comment](https://egraphs.zulipchat.com/#narrow/stream/375765-egg.2Fegglog/topic/embedded.20lambda.20functions/near/393293171) a while ago about how to embed functions in egglog.

As implemented, this supports functions that return primitives and eqsorts, since it uses the builtin mechanism for calling functions.

Although it doesn't support anonymous functions at the moment, adding them would be relatively trivial I believe as syntactic sugar, that would translate to a new global function with a random name and a rewrite rule to cover all cases.

I made sure that this implementation supports partial application, which is necessary for rewrite rules that transform functions.

The implementation is fairly self-contained as a primitive sort, requiring no changes to the parser or type checker. The only main breaking change is requiring a mutable EGraph pointer to be passed to primitive functions. This is necessary so that the function evaluation can possibly mutate the unionfind data structure, if new e-classes are created. I am unable to get access to a mutable egraph in the query checker, so these will panic if attempted to use in facts, they can only be used in action. This limitation seems reasonable.

One of the driving use cases for this on my end, besides just generally being cool to be able to optimize functional programs, is @sklam's example for [defining a higher level array API](https://gist.github.com/sklam/5e5737137d48d6e5b816d14a90076f1d?permalink_comment_id=4882443#gistcomment-4882443) and translating that within egglog.


saulshanabrook:
@ztatlock and others had wondered whether this PR could be simplified by not supporting curried functions. This would remove the storage of values in the sort, making the function sort not a container sort.

My sense from implementing it below is that it would be possible using codegen/macros, but it expands the amount of code and would require adding implementations for every combination of type of curried arguments for every function.

---


As currently implemented, this PR would allow you to define a map function over a custom user defined map data structure:

```
(datatype Math
  (Num i64)
  (Var String)
  (Add Math Math)
  (Mul Math Math))

(rewrite (Mul (Num x) (Num y)) (Num (* x y)))

(datatype MathList
  (Nil)
  (Cons Math MathList))

(sort MathFn (Fn (Math) Math))
(function list-map-math (MathFn MathList) MathList)
(rewrite (list-map-math fn (Nil)) (Nil))
(rewrite (list-map-math fn (Cons x xs)) (Cons (app fn x) (list-map-math fn xs)))

;; version where partial application is allowed

(function list-multiply-by (MathList Math) MathList)
(rewrite (list-multiply-by l i) (list-map-math (fn "Mul" i) l))
```

However, if partial application was not allowed, we could implement this by defining a custom map function that would partially apply an arg that is passed into it in the function invocation:

```
;; alternative version where no partial application is allowed

;; Must make a new map function that also takes a binary fn instead of a unary one
;; as well as an initial partialled applied arg
(sort MathMathFn (Fn (Math Math) Math))
(function list-map-math-partial (MathMathFn MathList Math) MathList)
(rewrite (list-map-math-partial fn (Nil)) (Nil))
(rewrite (list-map-math-partial fn (Cons x xs) y) (Cons (app fn y x) (list-map-math-partial fn xs y)))

(rewrite (list-multiply-by l i) (list-map-math-partial (fn "Mul") l i))
```

Or alternatively, we could write a wrapper for a unary math function that supports partial application and use this  (thanks @yihozhang for this version):

```
(sort MathMathFn (Fn (Math Math) Math))

(datatype MathFnWrapper (partial-app MathMathFn Math))
(function App (MathFnWrapper Math) Math)

(rewrite (App (partial-app f a) b) (app f a b))

;; rewrite
(function list-map-math (MathFnWrapper MathList) MathList)
(rewrite (list-map-math fn (Nil)) (Nil))
(rewrite (list-map-math fn (Cons x xs)) (Cons (App fn x) (list-map-math fn xs)))

(rewrite (list-multiply-by l i) (list-map-math (partial-app (fn "Mul") i)))
```

oflatt:
Here's a nice example of how you might do partial application in current egglog. It requires a little bit of boilerplate but it's not too bad:

https://egraphs-good.github.io/egglog/?program=XQAAgACDBAAAAAAAAAAUGQgnfMUD9dO10C6gYDgH7EMCDsP59lqRF2h73FWgyD8WAkCEiBUexoj9z4WGjCHzYIc2LwHce_oYevbmNoBoN_WQ0Ts129CFhQcgqLvd8Zghww4Y3IeW7pA3n8eY20nLHxZ4zXkw6JDSjaSDWK0QQJZ390IDSmZeEE3CmnobPAqTsX8BTdPj8QSF2zgKh5NEEX2EgxsFiBdyGhGDLqKjxejD7h2vbJ5lZDCckd5FQRFFAzgthseF8hCaVbe8kJU5fqVDBKluNgDdTE-eM6dXpgLktUJzigaujVbemliqhHNWu72dmDCnkMNGsm-ekguI93K4PDyt7mN076MFiYzL-v0xR9u0R3D9e1KKf-oWM5SSGIhFSUDFs2w5rVO0BHFZFXVvqDN9CLnk4qha7rVChJrEg9nolD6ZQ4NFfpGoEeGPqvQPc_n5NcCZvWLOXDLPCtkb6Qqpo7kcwRBzZoyl_h83cEZWVE2afqoNAP_Xx-5sPqJU_MQ75bM6ehIg-x14LK329vrl6OKZ8WbSZlNR7b9ztvwJDKIEJmpNxffZ4v4g-SJoF__sLfFz

The nice thing about this example is that it gives you an easy way to reference partial applications without any change to egglog. You do have to write conversion rules that go from a partial application to a normal one, but they are easily generated.

saulshanabrook:
A few comments I got yesterday from folks on this PR:

* Would it be possible to defer the canonicalization of the stored partial args until call time?
* This might not interact well with the (future) proofs features
* It would be nice to have a friendly syntax than a string quoted function name
* How does this relate to more generally implementing functional programming in egglog?

saulshanabrook:
At the meeting today, the consensus was this seems good to go to experiment on in the Python bindings. I prefaced all the public-facing functionality with `unstable` so that it's clear this could change.

If this is merged, I will work on adding a higher-level semantics to the Python bindings on top of this, for anonymous functions.

tests/unstable-fn.egg
```diff
@@ -0,0 +1,65 @@
...
+(let doubled-x (list-multiply-by x (Num 2)))
+(run 100)
+(check (= doubled-x (Cons (Num 2) (Cons (Num 4) (Cons (Num 6) (Nil))))))
+
+;; Test we can define a higher order compose function
+
+(function composed-math (MathFn MathFn Math) Math)
+(rewrite (composed-math f g v) (unstable-app f (unstable-app g v)))
+
+(let square-of-double (unstable-fn "composed-math" square-fn (unstable-fn "Mul" (Num 2))))
```
oflatt, line 50: We shouldn't use strings here for the identifier `composed-math`

src/typechecking.rs
```diff
@@ -45,10 +45,13 @@ impl Default for TypeInfo {
         res.presort_names.extend(MapSort::presort_names());
         res.presort_names.extend(SetSort::presort_names());
         res.presort_names.extend(VecSort::presort_names());
+        res.presort_names.extend(FunctionSort::presort_names());
```
oflatt, line 48: Should we make typechecking work for this PR? What happens if you try to run an ill-typed program- do you get ill-typed eclasses?

src/sort/fn.rs
```diff
@@ -0,0 +1,347 @@
...
+    fn get_type_constraints(&self) -> Box<dyn TypeConstraint> {
+        Box::new(FunctionCTorTypeConstraint {
+            name: self.name,
+            function: self.function.clone(),
+            string: self.string.clone(),
+        })
+    }
+
+    fn apply(&self, values: &[Value], egraph: Option<&mut EGraph>) -> Option<Value> {
+        let egraph = egraph.expect("Cannot create function in fact only in actions");
```
oflatt, line 255: This error message should be improved
Ideally, we should check that we don't use this primitive in queries at all at typechecking time

oflatt (changes requested):
Had a few concerns
The main question in my mind is if we need to make typechecking work for these functions

tests/unstable-fn.egg
saulshanabrook: AFAIK there isn't a way to use real identifiers without a large change to parsing/type checking, which I was trying to avoid for this sort of unstable feature.

src/typechecking.rs
saulshanabrook: I added a few type checking failure examples to show that it will make sure the input types are correct when calling a function and also that the return type is properly typed as well. I don't believe there is a way to get ill typed e-classes with this.

src/sort/fn.rs
saulshanabrook: I rewrote it a bit, let me know if you have any suggestions on how to make it clearer.

I think, although correct me if I am wrong, that the easiest way to make this fail during type-checking would be to modify [`Facts.to_query`](https://github.com/egraphs-good/egglog/blob/83a51a9f6f1ad69b5bd067558dd7e7a29cbefbbf/src/ast/mod.rs#L1000-L1007) and [`GenericExpr.to_query`](https://github.com/egraphs-good/egglog/blob/83a51a9f6f1ad69b5bd067558dd7e7a29cbefbbf/src/core.rs#L516-L523) to return `Result`s so that we can return a type-checking error if you try to create a query with these functions.

I am hesitant to make that larger change for just this special case, since it's still unstable.

saulshanabrook:
Hey @oflatt, thanks for taking a look at this!

I responded to your points, but overall I am hesitant to make any changes to much outside of the `fn.rs` sort definition unless they are really required since this is unstable and I wouldn't want to tie us to this implementation if we do want to move.

Also, even if we keep this, we might not want to expose it directly to users as we talked about, so if we made a higher level API, we could make sure that had all the proper affordances, like not representing functions as strings but as symbols.

I am excited to get this in so I can start experimenting with it in the Python bindings!

src/typechecking.rs
oflatt: Thanks! Nicely done

tests/unstable-fn.egg
oflatt: okay!

oflatt (approved):
This looks good to me, we should merge for now and keep thinking about functional programming

PR #349, opened 2024-02-24
Symbol generation fixes

oflatt:
This PR allows different code to set different reserved symbols for the symbol gen struct
This fixes all of our name conflicts. Desugaring using `__` by default and the middle of the compiler uses `$`


yihozhang (approved):
LGTM

PR #353, opened 2024-03-09
Add custom extraction cost functions

mwillsey:
This add support for custom cost function. You can write code like this now:

```

(datatype E
  (Foo String :cost FooCost))

(union (Foo "x") (Foo "y"))
(union (Foo "y") (Foo "z"))

(set (FooCost "x") 17)
(set (FooCost "y") 11)
(set (FooCost "z") 15)

(extract (Foo "y"))

```

I'm not 100% if what I did with typechecking is okay, but at least this simple example works.

### DIFF
--- a/src/ast/parse.lalrpop
+++ b/src/ast/parse.lalrpop
@@ -101,9 +101,12 @@ Schedule: Schedule = {
     <ident:Ident> => Schedule::Run(RunConfig { ruleset: ident, until: None }),
 }
 
-Cost: Option<usize> = {
-    ":cost" <UNum> => Some(<>),
-    => None,
+Cost: CostFn = {
+    ":cost" <UNum> => CostFn::Constant(<>),
+    ":cost" <Ident> => CostFn::Table(false, <>),
+    // dont use this, it's only for round-tripping desugaring
+    ":_declared_cost" <Ident> => CostFn::Table(true, <>),
+    => CostFn::Constant(1),
 }
 
 NonLetAction: Action = {
--- a/src/ast/desugar.rs
+++ b/src/ast/desugar.rs
@@ -1,26 +1,6 @@
 use super::{Rewrite, Rule};
 use crate::*;
 
-fn desugar_datatype(name: Symbol, variants: Vec<Variant>) -> Vec<NCommand> {
-    vec![NCommand::Sort(name, None)]
-        .into_iter()
-        .chain(variants.into_iter().map(|variant| {
-            NCommand::Function(FunctionDecl {
-                name: variant.name,
-                schema: Schema {
-                    input: variant.types,
-                    output: name,
-                },
-                merge: None,
-                merge_action: Actions::default(),
-                default: None,
-                cost: variant.cost,
-                unextractable: false,
-            })
-        }))
-        .collect()
-}
-
 fn desugar_rewrite(
     ruleset: Symbol,
     name: Symbol,
@@ -236,7 +216,7 @@ pub(crate) fn desugar_command(
             inputs,
         } => desugar.desugar_function(&FunctionDecl::relation(constructor, inputs)),
         Command::Declare { name, sort } => desugar.declare(name, sort),
-        Command::Datatype { name, variants } => desugar_datatype(name, variants),
+        Command::Datatype { name, variants } => desugar.desugar_datatype(name, variants),
         Command::Rewrite(ruleset, rewrite, subsume) => {
             desugar_rewrite(ruleset, rewrite_name(&rewrite).into(), &rewrite, subsume)
         }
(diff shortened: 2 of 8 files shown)


mwillsey:
This will solve Tyler Hou's issue [on Zulip](https://egraphs.zulipchat.com/#narrow/stream/328972-general/topic/How.20can.20I.20find.20the.20tree.20associated.20with.20an.20extraction.3F).

saulshanabrook (commented):
I love adding costs as a function table!

It is a nice way to add custom costs to different rulesets.

---

I had a go at thinking how to expose this on the Python side of things. One way would be to have users define their own cost functions, so that they are typed properly. With this kind of interface, it would statically verify that both functions take the same arguments and the cost function should return an `i64`:

```python
# mypy: disable-error-code="empty-body"

from collections.abc import Callable
from typing import TypeVar, TypeVarTuple, Unpack

from egglog import EGraph, Expr, StringLike, i64, set_, union

T = TypeVar("T", bound=Expr)
TS = TypeVarTuple("TS")


# typing of `function` wrapper that support cost keyword arg
def function(*, cost: Callable[[Unpack[TS]], i64]) -> Callable[[Callable[[Unpack[TS]], T]], Callable[[Unpack[TS]], T]]:
    raise NotImplementedError


# User code below...


class E(Expr):
    pass


def FooCost(x: StringLike) -> i64: ...


@function(cost=FooCost)
def Foo(x: StringLike) -> E: ...


egraph = EGraph()

egraph.register(
    union(Foo("x")).with_(Foo("y")),
    union(Foo("y")).with_(Foo("z")),
    set_(FooCost("x")).to(i64(17)),
    set_(FooCost("y")).to(i64(11)),
    set_(FooCost("z")).to(i64(15)),
)

egraph.extract(Foo("y"))
```

It does make me wonder about an alternate interface, with a global `cost` function that can be used on any value to set its cost. In egglog this might look like:

```
(datatype E
  (Foo String))

(union (Foo "x") (Foo "y"))
(union (Foo "y") (Foo "z"))

(set (cost (Foo "x")) 17)
(set (cost (Foo "y")) 11)
(set (cost (Foo "z")) 15)

(extract (Foo "y"))
```

In Python, it could look like:


```python
# mypy: disable-error-code="empty-body"

from typing import TypeVar, TypeVarTuple

from egglog import EGraph, Expr, StringLike, function, i64, set_, union

T = TypeVar("T", bound=Expr)
TS = TypeVarTuple("TS")


def cost(x: Expr) -> i64:
    raise NotImplementedError


# User code below...


class E(Expr):
    pass


@function
def Foo(x: StringLike) -> E: ...


egraph = EGraph()

egraph.register(
    union(Foo("x")).with_(Foo("y")),
    union(Foo("y")).with_(Foo("z")),
    set_(cost(Foo("x"))).to(i64(17)),
    set_(cost(Foo("y"))).to(i64(11)),
    set_(cost(Foo("z"))).to(i64(15)),
)

egraph.extract(Foo("y"))
```

From the Python side, an advantage of this version over the current one would be that user's don't need to create a new function with typing.

It does also change the semantics a bit, meaning that any function would support setting custom costs, instead of just the ones you specify.

mwillsey:
I like your idea Saul, but the reason that it is like `(FooCost args)` rather than `(cost (Foo args))` is because the latter "loses precision" as e-class gen unioned, i.e., you can't maintain a handle on a specific e-node. The first approach builds the constructor in, so to speak, so it's always clear which e-node a cost is referring to. Perhaps a new node-based architecture would make this possible, but for now I think it's necessary to be the `FooCost` way.

saulshanabrook:
Ah I see what you mean!

If I am understanding it correctly, the current implementation stores a cost associated with the tuple of e classes representing the args of the function. With my proposed change, it would store the cost with the eclass of the function return value instead. 

It makes me think it's a bit similar to subsumption or deletion of a node? Which is based on identity of the args not the output. 

What if we added a "cost" action that takes a call and an i64? It would be similar to subsume. When I implemented subsume, I made an enum with delete as two row level actions, where cost setting could be a third. 

Would that have the same expressivity do you think?

yihozhang:
On another note, it might be better to make `:cost FooCost` a parameter of
`extract` than to datatype/constructor, so that `extract` is parametrized
by the cost model used. This will be even nicer if we have a module system
in the future, and we will be able to write something like `(extract e
:cost-model CostModelModuleA)`. The current downside is that since
datatypes are open, it is not clear what parameters we should provide to
`extract`. We can simply say if something is not in `extract`'s parameters,
it's by default unextractable.

src/ast/parse.lalrpop
```diff
@@ -101,9 +101,12 @@ Schedule: Schedule = {
     <ident:Ident> => Schedule::Run(RunConfig { ruleset: ident, until: None }),
 }
 
-Cost: Option<usize> = {
-    ":cost" <UNum> => Some(<>),
-    => None,
+Cost: CostFn = {
+    ":cost" <UNum> => CostFn::Constant(<>),
+    ":cost" <Ident> => CostFn::Table(false, <>),
```
oflatt, line 106: Perhaps it would be simpler to force the user to declare the table themself, then typecheck it properly? This is pretty good too but may be confusing.

oflatt (approved):
Great work on this PR! I approve.
Would be great to have a more complex test though

mwillsey:
I thought of that, and I agree. The difficulty is that it’s hard to do for
datatypes, since the cost-table would have to be defined after the datatype
defn to mention the sort, but then that’s after its use in the variant as
well!

oflatt:
Yeah, that makes sense

saulshanabrook:
> What if we added a "cost" action that takes a call and an i64? It would be similar to subsume. When I implemented subsume, I made an enum with delete as two row level actions, where cost setting could be a third.


I had a go at implementing it this way: https://github.com/egraphs-good/egglog/pull/355

Let me know what you think.

PR #360, opened 2024-03-27
bump regex to 1.10

ajpal:
Needed for eggcc to be compatible with bril2json

### DIFF
--- a/src/gj.rs
+++ b/src/gj.rs
@@ -503,7 +503,7 @@ impl EGraph {
             log::debug!("Variable costs: {:?}", ListDebug(&var_cost, "\n"));
 
             let var = *var_cost[0].1;
-            let info = vars.remove(&var).unwrap();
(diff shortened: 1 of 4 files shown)


src/gj.rs
```diff
@@ -503,7 +503,7 @@ impl EGraph {
             log::debug!("Variable costs: {:?}", ListDebug(&var_cost, "\n"));
 
             let var = *var_cost[0].1;
-            let info = vars.remove(&var).unwrap();
+            let info = vars.swap_remove(&var).unwrap();
```
ajpal, line 506: Bumping rust version resulted in: 
```
warning: use of deprecated method `indexmap::IndexMap::<K, V, S>::remove`: `remove` disrupts the map order -- use `swap_remove` or `shift_remove` for explicit behavior.
```

oflatt (approved):
lgtm

PR #362, opened 2024-04-05
Add combined rulesets to egglog

oflatt:
This PR adds combined rulesets to egglog, allowing rulesets to be collected and run together.
We plan to use this feature in eggcc, since we declare a ruleset per optimization. We want all our optimizations to run together each iteration.

The PR also cleans up the main loop of egglog, breaking it into functions.
Finally, it deletes the backoff scheduler functionality, which nobody uses as far as I can tell.

### DIFF
--- a/src/ast/desugar.rs
+++ b/src/ast/desugar.rs
@@ -274,8 +274,10 @@ pub(crate) fn desugar_command(
             result
         }
         Command::Sort(sort, option) => vec![NCommand::Sort(sort, option)],
-        // TODO ignoring cost for now
         Command::AddRuleset(name) => vec![NCommand::AddRuleset(name)],
+        Command::CombinedRuleset(name, subrulesets) => {
+            vec![NCommand::CombinedRuleset(name, subrulesets)]
+        }
         Command::Action(action) => vec![NCommand::CoreAction(action)],
         Command::Simplify { expr, schedule } => desugar_simplify(desugar, &expr, &schedule),
         Command::Calc(idents, exprs) => desugar_calc(desugar, idents, exprs, seminaive_transform)?,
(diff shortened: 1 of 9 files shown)


saulshanabrook:
Seems to make sense to me to add this kind of feature. I was curious if this could be added as a schedule constructor instead to keep the change smaller?

oflatt:
Great question.
I would have liked to add this as a schedule constructor, but it doesn't compose with other schedule constructors. In other words, it has to be a leaf node of schedules.

What combined rulesets do is:
1. Run the searchers for all rules in all of the rulesets
2. Run the appliers for all of these rules

You can't run multiple iterations here as sub-schedules do: you have to phase the searchers and the appliers.

saulshanabrook:
> I would have liked to add this as a schedule constructor, but it doesn't compose with other schedule constructors. In other words, it has to be a leaf node of schedules.

Yeah that makes sense. What if we just extended the `run` schedule to take multiple ruleset names instead of just one? With the semantics of this combined ruleset if you supply multiple names?

src/ast/desugar.rs
```diff
@@ -274,8 +274,10 @@ pub(crate) fn desugar_command(
             result
         }
         Command::Sort(sort, option) => vec![NCommand::Sort(sort, option)],
-        // TODO ignoring cost for now
         Command::AddRuleset(name) => vec![NCommand::AddRuleset(name)],
+        Command::CombinedRuleset(name, subrulesets) => {
+            vec![NCommand::CombinedRuleset(name, subrulesets)]
+        }
```
saulshanabrook, line 280: Would you be able to prefix this with unstable too? That way in the Python bindings the command will be prefixed with that as well.

saulshanabrook (commented):
Looks good overall! Just was wondering if you might be able to prefix the command as well, since those are exposed in the Python bindings and are public externally.

src/ast/desugar.rs
oflatt: Great idea

saulshanabrook (approved):
Looks good! Thanks Oliver!

PR #365, opened 2024-04-12
Easier access to logs for the web demo

yihozhang:
After this PR:
image
image

### DIFF
--- a/web-demo/static/worker.js
+++ b/web-demo/static/worker.js
@@ -8,16 +8,22 @@ async function work() {
     // Set callback to handle messages passed to the worker.
     self.onmessage = async event => {
         try {
+            logbuffer = [];
             let result = run_program(event.data);
             console.log("Got result from worker", result);
             // Can't send the result directly, since it contains a reference to the
             // wasm memory. Instead, we send the dot and text separately.
-            self.postMessage({dot: result.dot, text: result.text});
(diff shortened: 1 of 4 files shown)


web-demo/static/worker.js
```diff
@@ -8,16 +8,22 @@ async function work() {
...
-            self.postMessage({dot: "", text: "Something panicked! Check the console logs..."});
+            self.postMessage({ dot: "", text: "Something panicked! Check the console logs...", log: logbuffer });
         }
     };
 }
 
+logbuffer = [];
+function log(level, str) {
+    logbuffer.push([level, str]);
+}
```
saulshanabrook, line 27: Is this used anywhere else?

saulshanabrook (approved):
Nice! I just tried it and it seems to work well.

The frontend code is starting to get more complicated and it might make more sense to just send structured data from the backend instead of HTML and style it in the frontend, but it seems fine for now.

I imagine if the editor keeps getting more complicated it might be worth moving to something like React in the future to manage state/complexity.

web-demo/static/worker.js
yihozhang: This is used by the Rust WebDemoLogger!

yihozhang:
Yeah, I agree the frontend code is getting more complex. It makes sense to switch to a more established framework for managing states, etc.

PR #372, opened 2024-04-30
Fix web demo box shadowing problem

yihozhang:
Currently, in the web visualization, E-node may be shadowed by its surrounding E-classes after a series animated transitions. This PR fixes it.

Step to reproduce: step through the Fibonacci example in the web demo using the slideshow mode.

See https://github.com/magjac/d3-graphviz/issues/308 .


saulshanabrook:
Thank you @yihozhang for getting this fixed upstream!

Is the included source file a built version of the d3-graphviz library, since a new version hasn't been released yet with the included change?

yihozhang:
Yep! @saulshanabrook

saulshanabrook (approved):
Looks good for now! I imagine that when a new version of d3-graphviz is released we will want to revert inlining the library and go back to depending on the hosted version.

PR #374, opened 2024-05-03
Step through by block for slideshow mode

yihozhang:
Right now the step-throughs in the slideshow mode go command by command, which gets tedious if several commands should be introduced together (e.g., a set of algebraic rules)

### DIFF
--- a/web-demo/static/index.html
+++ b/web-demo/static/index.html
@@ -209,10 +209,12 @@
                 let start = currentIndex;
                 // find next open parenthesis which isn't inside a comment
                 let inComment = false;
+                let newlines = 0;
                 while (currentIndex < content.length) {
                     if (content[currentIndex] == ";") {
                         inComment = true;
                     } else if (content[currentIndex] == "\n") {
+                        if (!inComment) newlines++;
                         inComment = false;
                     } else if (content[currentIndex] == "(" && !inComment) {
                         break;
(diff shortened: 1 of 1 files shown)


oflatt:
What is the new behavior? I don't understand the changes, particularly the line counting if

yihozhang:
Let's use this program as an example:
```
;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
;; An Expr datatype
;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
(datatype Expr
  (Num i64)
  (Var String)
  (Add Expr Expr)
  (Mul Expr Expr))

;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
;; Some constants
;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
(let zero (Num 0))
(let one (Num 1))
(let two (Num 2))
(let one-plus-two (Add one two))
```
In the old behavior, You need to click `Next` 5 times. In the new behavior, you only need to click next twice (a command is in a new block if it is separated by more than once a newline character). @oflatt

chandrakananandi (approved):
I guess I maybe have the power to do this so here you go :) :shipit: 
good luck with demo!

hmmm maybe you need someone else...i am not sure i can approve :D

mwillsey (approved):

Issue #377, opened 2024-05-13
Potential bug in dot file output


hessammehr:
The following works seems to work fine for a small number of data points, i.e. `(ion ...)`s but larger inputs seem to produce dot files that make no sense (see attached test case).

```scheme
(datatype Obs
    (Ion f64)
    (Combination Obs Obs))

(rule (
        (= child1 (Ion childmz1))
        (= child2 (Ion childmz2))
        (>= childmz1 childmz2)
        (= lhs (Ion parentmz))
        (= delta (- parentmz (+ childmz1 childmz2)))
        (< delta 0.1) (> delta -0.1))
    (
        (union lhs (Combination child1 child2))
    ))

(let ion_0 (Ion 63.167))
; etc.
```

Egglog invocation

```sh
egglog test.egg --to-dot
```

# Expected behaviour
Each `Combination` node having two child nodes.

# Actual behaviour
Some `Combination` nodes have one or no child nodes. Not a rendering issue afaik, since the `.dot` file is also missing the necessary edges.

JSON output seems fine.

[test_case.txt](https://github.com/egraphs-good/egglog/files/15294327/test_case.txt)

yihozhang:
Yeah, egglog limits the number of nodes to be rendered: https://github.com/egraphs-good/egglog/blob/main/src/lib.rs#L1539 This prevents the visualization from blowing up, but seems confusing... The interface should probably be more explicit on that (e.g. a question-mark-in-bubble button).

oflatt:
Related to #368 and problems around the viz. I like the idea of a question mark and a warning

yihozhang:
I believe this will be fixed by #394

PR #379, opened 2024-05-16
SVG/dot output limits specified via command line arguments

hessammehr:
Partially addresses #377 by allowing the number of functions/calls per function to be specified via the command line. Preserved current behaviour if user does not specify limits.

Very happy to consider alternative approaches that may be more widely applicable.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1539,11 +1539,13 @@ impl EGraph {
     pub fn serialize_for_graphviz(
         &self,
         split_primitive_outputs: bool,
+        max_functions: usize,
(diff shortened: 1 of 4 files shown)


yihozhang (approved):
This seems to be a reasonable change. Thank you for the PR!

yihozhang:
The CI is failing. Could you also fix the web demo?

hessammehr:
Oh yes, missed that one. Should be fixed now.

yihozhang:
Thanks!

saulshanabrook:
Thank you for this!

PR #380, opened 2024-05-17
Prepare for 0.2.0

yihozhang:


yihozhang:
don't merge until we chat about future egglog release plans

PR #381, opened 2024-05-23
Add error message for ruleset that DNE.

cgyurgyik:
This PR adds an error message when the ruleset does not exist.

```
// Before
called `Option::unwrap()` on a `None` value 

// After
test panicked: ruleset does not exist: wrongnameruleset
```


saulshanabrook (approved):
Thank you for this! Seems like a good fix :)

PR #384, opened 2024-05-29
Accept f64 function arguments in `(input ...)`

hessammehr:
Being able to read in f64 fields would be quite useful. If this initial implementation seems reasonable enough I'd be happy to add a test or two ...


yihozhang:
LGTM, thanks!

yihozhang (approved):

PR #385, opened 2024-06-06
Fix support for nullary unstable functions

### LINKED ISSUE #382: Fix Nullary Unstable Functions
I noticed that nullary functions don't parse correctly with the unstable function sort (i.e. `(sort TestNullaryFunction (UnstableFn () Math))`)

That's because the second argument is now parsed as a `Unit` type instead of a `Call` arg.

I could fix it and keep the existing format, but I was thinking another option would be to just change the sort to take all its args as a list, so there is always at least one and put the return type first. This would make it parse the same as other collection types.

So instead of `(sort MathFn (UnstableFn (Arg) Ret))` you would have `(sort MathFn (UnstableFn (Ret Arg)))` or for this nullary case `(sort TestNullaryFunction (UnstableFn (Math))))`

It's a bit less intuitive for anyone writing `UnstableFn` manually in egglog, but makes the serialization code easier in Python and will make the parsing a bit easier in Egglog.  I don't think it's a big deal to make the ergonomics worse to use in pure egglog, b/c I think as is it's already pretty odd to use, and I assume if we want it in core we will add better ergonomics to it anyways, like there are in the Python bindings.

But I just wanted to check before making a PR for this change if anyone had objections, in which case I could keep the format the same as it is, but just fix the parsing for the empty args case.

saulshanabrook:
This resolves #382 by adding parsing support for defining sorts of 0 argument unstable functions.

It does this by handling that input `()` as the `Unit` constructor.

If we ever want to remove the Unit constructor, we could move the return type to the list of argument types, to make sure it always has at least one value.

This would also make parsing it more similar to other datatypes.

For now I just kept it the same to make the change minimal.

### DIFF
--- a/src/sort/fn.rs
+++ b/src/sort/fn.rs
@@ -62,27 +62,36 @@ impl FunctionSort {
         name: Symbol,
         args: &[Expr],
     ) -> Result<ArcSort, TypeError> {
-        if let [Expr::Call((), first, rest_args), Expr::Var((), output)] = args {
+        if let [inputs, Expr::Var((), output)] = args {
             let output_sort = typeinfo
                 .sorts
                 .get(output)
                 .ok_or(TypeError::UndefinedSort(*output))?;
-            let all_args = once(first).chain(rest_args.iter().map(|arg| {
-                if let Expr::Var((), arg) = arg {
-                    arg
-                } else {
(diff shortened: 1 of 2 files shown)


oflatt (approved):
looks good to me

PR #389, opened 2024-06-26
Adding span annotations to the internal representation

yihozhang:
What this PR does:
- [x] Add span annotations to the parser file
- [x] Add span annotations to internal representations (`CoreIR`, etc)

What this PR does not do:
* Actual span-annotated error reporting, which will be a future PR. However, the test in typechecking.rs checks that we are getting the correct span (at least for TypeError::Arity).


saulshanabrook:
So just to be clear, the goal for the PR after this is to not only have row/column for parsing but also when there is say an error running a rule or something? Or which types of errors do we hope to attach source information too?

yihozhang:
@saulshanabrook Since the ASTs are annotated, technically we should be able to locate the source program location for every (runtime/compile time) error. This PR leaves this problem open though, since it has not got to the actual error reporting part, but I think this is the goal.

src/ast/mod.rs
```diff
@@ -51,11 +52,38 @@ pub(crate) enum Ruleset {
...
-pub type NCommand = GenericNCommand<Symbol, Symbol, ()>;
+pub(crate) const DEFAULT_FILENAME: &str = "<unnamed.egg>";
+pub(crate) const DUMMY_FILENAME: &str = "<internal.egg>";
+
+/// A [`Span`] contains the file name and a pair of offsets representing the start and the end.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Span(pub Symbol, pub usize, pub usize);
+
+lazy_static! {
+    pub(crate) static ref DUMMY_SPAN: Span = Span(Symbol::from(DUMMY_FILENAME), 0, 0);
```
oflatt, line 63: we talked about this in the meeting, but we should either not have a dummy span or make the annotation not generic

oflatt (changes requested):

src/ast/mod.rs
ezrosent: +1; furthermore, I think having a single annotation type and not having type parameters may be the better option. Up to Yihong though.

src/core.rs
```diff
@@ -328,50 +373,65 @@ impl<Leaf: Clone> Query<ResolvedCall, Leaf> {
...
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
-pub enum GenericCoreAction<Head, Leaf> {
-    Let(Leaf, Head, Vec<GenericAtomTerm<Leaf>>),
-    LetAtomTerm(Leaf, GenericAtomTerm<Leaf>),
-    Extract(GenericAtomTerm<Leaf>, GenericAtomTerm<Leaf>),
-    Set(Head, Vec<GenericAtomTerm<Leaf>>, GenericAtomTerm<Leaf>),
-    Change(Change, Head, Vec<GenericAtomTerm<Leaf>>),
-    Union(GenericAtomTerm<Leaf>, GenericAtomTerm<Leaf>),
-    Panic(String),
+pub enum GenericCoreAction<Head, Leaf, Ann = Span> {
```
ezrosent, line 376: optional: I think having named fields here would make things a little more readable.

ezrosent (approved):

yihozhang:
@ezrosent @oflatt Thanks for the feedback! I have got rid of the `Ann` type parameter.

src/ast/mod.rs
```diff
@@ -55,16 +55,24 @@ pub(crate) enum Ruleset {
...
-pub struct Span(pub Symbol, pub usize, pub usize);
+pub struct Span(pub Arc<SrcFile>, pub usize, pub usize);
 
 lazy_static! {
-    pub(crate) static ref DUMMY_SPAN: Span = Span(Symbol::from(DUMMY_FILENAME), 0, 0);
+    pub(crate) static ref DUMMY_FILE: Arc<SrcFile> = Arc::new(SrcFile {
+        name: DUMMY_FILENAME.to_string(),
+        contents: None
+    });
+    pub(crate) static ref DUMMY_SPAN: Span = Span(DUMMY_FILE.clone(), 0, 0);
```
saulshanabrook, line 73: What if Span was an enum (or use an `Option<Span>`)?  I was just thinking if error reporting then we could differentiate between having no line number versus outputting a dummy one.

I don't think it's a big deal though either way.

oflatt (approved):
lgtm

PR #391, opened 2024-07-11
Fix `extract_variants` by comparing against leader `Value`

gussmith23:
The missing call to `find` meant that this code wasn't always figuring out when two Values were in the same eclass.

### DIFF
--- a/src/extract.rs
+++ b/src/extract.rs
@@ -84,7 +84,7 @@ impl EGraph {
 
                 func.nodes
                     .iter(false)
-                    .filter(|&(_, output)| (output.value == output_value))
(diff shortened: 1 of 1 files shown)


gussmith23:
@oflatt thinks this may not be needed; he thinks that perhaps I'm doing something that's causing the egraph to not be rebuilt, which is causing this issue.

gussmith23:
Note that doing a manual rebuild before calling `extract_variants` _without_ this PR doesn't work in my case.

oflatt:
@gussmith23 Do you have a branch where we can reproduce this issue? It seems like a rebuilding bug we'd like to look into.

Alex-Fischman:
Closing as unreproducible.

PR #394, opened 2024-07-15
Serialize edges to omitted nodes

saulshanabrook:
Previously, if certain nodes were omitted due to size constraints in the serialization format, than they would not appear at all and the edge to them would be missing as well.

This was confusing, because then some functions would have the wrong number of arguments.

Now, if an edge points to an e-class where all nodes in that e-class were omitted, a dummy `[...]` node is added to that e-class so that they have something to point to.

Previously, we discussed implementing this change by modifying the serialization format to allow pointing to `null` node IDs as children. I went with this approach here, by just creating omitted dummy children, so you can still know the sort of the omitted node and if two nodes point to the same omitted e-class, that will also be visible.

Example of omitted node (from `tests/combinators.svg` produced by `make graphs`):

Screenshot 2024-07-15 at 9 08 09 AM

### DIFF
--- a/src/serialize.rs
+++ b/src/serialize.rs
@@ -43,6 +43,7 @@ impl EGraph {
     /// - Primitives: 1.0
     /// - Function without costs: 1.0
     /// - Function with costs: the cost
+    /// - Omitted nodes: infinite
     ///
     /// For node IDs:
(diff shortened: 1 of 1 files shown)


yihozhang (approved):
This looks great!

Issue #388, opened 2024-06-20
Built In Collection Constructor Type Checking Prohibitively Slow


saulshanabrook:
Hey all,

I tried switching some of the Python numba code to use `Vec` and `vec-of` to construct lists and found the performance degraded by a ton. After doing some profiling, I saw all the time was in type checking, in particular typechecking `XOR` constraints. Which makes sense because I have a few different types of vecs and it has to infer what type it is based on the contents.

I also noticed that most of the time was spent cloning the assignments hashmap, which happens when handling the XOR:

https://github.com/egraphs-good/egglog/blob/92b3d9d04fb8ecde2194578db7aa480fd6217429/src/constraint.rs#L128-L131

I thought, hm, if cloning is so expensive, then maybe we should try a copy on write hashmap, so that cloning is O(1)? I [gave this a go](https://github.com/saulshanabrook/egg-smol/tree/immutable-assignments) and saw that the benchmark I was running (this is total time of Python + Rust) **went from 1143.05 seconds to 10.92 seconds**! A 100x speedup?

But then I noticed that on code that doesn't use a lot of vec collections, switching to the immutable data structure slows things down by a few %.

So for now, I'll probably just avoid `vec-of`, but I wanted to open this issue just to warn folks if they are using it when they have multiple types of vec's registered, that it maybe be really slow.

There is likely a way to solve this with some refactoring of the type checking logic.

oflatt:
Great find!
We should try to address this issue, imo the small % slowdown with switching to an immutable data structure is worth it if it fixes the problem.

In the long term, my opinion is still that the constraint-based typechecker is way too complex, and we should design a more principled type system that can be checked efficiently.

saulshanabrook:
I had a hunch that my issue was due to my large expression sizes. I would add a big (tens of thousands of expressions) expression tree to egglog, all as one expression. This is relatively inefficient, since there are a lot of duplicated nodes.

I recently added CSE to the Python egglog bindings transparently (https://github.com/egraphs-good/egglog-python/pull/179) creating let bindings automatically when a user inserts an expression to the e-graph, so that the type checking is less stressed out.

This was a huge performance improvement regardless and now with this change, I see no improvement with immutable data structures on the collection code. So I'll close this issue.

yihozhang:
I think it would still be useful to fix the original issue though- since there might be users who don't use the python binding and generate big S-expressions. Maybe we should reopen this issue?

saulshanabrook:
Sure

PR #395, opened 2024-07-15
Fix container type checking worst case

saulshanabrook:
Closes https://github.com/egraphs-good/egglog/issues/388 by using an immutable data type for type checking.

I also added a test case that reproduces the slowdown.

Running `time ./target/release/egglog tests/python_array_optimize.egg` before the immutable data types:

```
Executed in   31.46 secs    fish           external
   usr time   25.64 secs  135.00 micros   25.64 secs
   sys time    4.09 secs  993.00 micros    4.09 secs
```

And after:

```
Executed in    3.16 secs      fish           external
   usr time  926.80 millis    0.16 millis  926.64 millis
   sys time  190.23 millis    1.00 millis  189.23 milli
```

## Are there regressions?

I also observed previously when testing this against Python that this would make everything else a bit slower. We don't have performance testing set up in this repo, so it's a bit cumbersome to measure, but for sanity I checked the eggcc example to see. After this change:

```
Executed in   52.85 secs    fish           external
   usr time   51.94 secs    0.32 millis   51.93 secs
   sys time    0.66 secs    1.26 millis    0.66 secs
```

Before this change:

```
Executed in   51.37 secs    fish           external
   usr time   50.40 secs    0.17 millis   50.40 secs
   sys time    0.53 secs    1.09 millis    0.53 secs
```

I am not sure if the change here is significant, it would be hard to tell without repeated testing and some statistics. But overall, the baseline performance seems comparable.

<details><summary>Generating Example</summary>
<p>
To generate the example I ran this Python file to reproduce one of the test cases:

```python
import time
from pathlib import Path

from sklearn import config_context, datasets
from sklearn.discriminant_analysis import LinearDiscriminantAnalysis

from egglog.exp.array_api import *
from egglog.exp.array_api_numba import array_api_numba_schedule
from egglog.exp.array_api_program_gen import *

iris = datasets.load_iris()
X_np, y_np = (iris.data, iris.target)


def run_lda(x, y):
    with config_context(array_api_dispatch=True):
        lda = LinearDiscriminantAnalysis(n_components=2)
        return lda.fit(x, y).transform(x)


def trace_lda(egraph: EGraph):
    X_arr = NDArray.var("X")
    assume_dtype(X_arr, X_np.dtype)
    assume_shape(X_arr, X_np.shape)
    assume_isfinite(X_arr)

    y_arr = NDArray.var("y")
    assume_dtype(y_arr, y_np.dtype)
    assume_shape(y_arr, y_np.shape)
    assume_value_one_of(y_arr, tuple(map(int, np.unique(y_np))))  # type: ignore[arg-type]

    with egraph:
        return run_lda(X_arr, y_arr)


egraph = EGraph()
expr = trace_lda(egraph)
egraph = EGraph(save_egglog_string=True)
start_time = time.time()
egraph.simplify(expr, array_api_numba_schedule)
print(f"Elapsed time: {time.time() - start_time}")
(Path(__file__).parent / "tmp.egg").write_text(egraph.as_egglog_string)
```

I also had to patch the function naming code so that it could be properly parsed by egglog:

```diff
diff --git a/python/egglog/egraph.py b/python/egglog/egraph.py
index ec82c5d..bc59d21 100644
--- a/python/egglog/egraph.py
+++ b/python/egglog/egraph.py
@@ -672,6 +672,9 @@ class _FunctionConstructor:
         return decls, ref
 
 
+ALLOWED = "-+*/?!=<>&|^/%_"
+
+
 def _fn_body_name(fn: Callable, decls: Declarations, function_decl: FunctionDecl) -> str:
     """
     Creates a function name from the function body.
@@ -685,7 +688,9 @@ def _fn_body_name(fn: Callable, decls: Declarations, function_decl: FunctionDecl
     ]
     fn = f"lambda {', '.join(signature.arg_names)}: {fn(*args)}"
     tp = f"Callable[[{', '.join(str(tp) for tp in signature.arg_types)}], {signature.return_type}]"
-    return f"cast({tp}, {fn})"
+    res = f"cast({tp}, {fn})"
+    # replace all non alphanumeric characters with underscores
+    return "".join(c if c.isalnum() or c in ALLOWED else "_" for c in res)
 
 
 def d(x):
```

</p>
</details>


yihozhang (approved):
Thank you for the fix!

Nits: I noticed that the new test is 1.25MB in size, which is bigger than the current src and tests directory combined. Any chance we can make it smaller?

saulshanabrook:
> Nits: I noticed that the new test is 1.25MB in size, which is bigger than the current src and tests directory combined. Any chance we can make it smaller?

Yeah it is a bit big. I was trying to find a smaller reproducible version, but it was hard to get the worst-case performance to reproduce manually.

PR #396, opened 2024-07-15
Serialized Class ID <-> Value

### LINKED ISSUE #392: API-supported way for translating between `Value`s in an `egglog::EGraph` and `ClassID`s in an `egraph_serialize::EGraph`
For a number of reasons which are generally summed up by [this issue](https://github.com/egraphs-good/egglog/issues/361), I often need to work with both `egglog::EGraph`s and `egraph_serialize::EGraph`s. It would be great to have an API-supported, stable way of translating between Values in an EGraph and EClass IDs in a serialized EGraph. As @saulshanabrook pointed out to me, it seems like the process of serialization uses a Value's index into the hashcons to generate its corresponding EClass ID, so that's a hacky way of going back and forth:
https://github.com/egraphs-good/egglog/blob/0da32092999b558486b521b5a106a523d032ffd4/src/serialize.rs#L104-L105

However, it would give me peace of mind to have a way of doing this that's library-supported!

Thanks all!

saulshanabrook:
This PR fixes #392 to add a public API for going from serialized Class IDs to Values.

It does this by changing the serialized class IDs to be more uniform. It should preserve all behavior.

cc @gussmith23


gussmith23:
You're a lifesaver!!

Just to be clear: there's a lot of things I have to do over a serialized egraph, because I can't inspect the egglog::EGraph enough to do what I need. I realize that egglog is different from egg, and that this is one of the results. I'm just trying to say that my use of the serialized egraph is mostly just a hack around the core problem of not being able to inspect the egglog::EGraph. Either way, Saul's PR helps my hack be less hacky :)

saulshanabrook:
I think having the serialized e-graph as a public API maybe isn't the worst, it at least makes it easy to change the internals of egglog without disrupting the API at all.

src/serialize.rs
```diff
@@ -144,12 +142,35 @@ impl EGraph {
...
     }
 
+    /**
+     * Gets the serialized class ID for a value.
+     */
+    pub fn value_to_class_id(&self, value: &Value) -> egraph_serialize::ClassId {
+        // Canonicalize the value first so that we always use the canonical e-class ID
+        let sort = self.get_sort_from_value(value).unwrap();
+        let mut value = *value;
+        sort.canonicalize(&mut value, &self.unionfind);
```
saulshanabrook, line 158: add an assert that the tag has no dash

src/serialize.rs
```diff
@@ -144,12 +142,35 @@ impl EGraph {
...
+            .map(|v| self.value_to_class_id(v))
             .collect();
 
         egraph
     }
 
+    /**
+     * Gets the serialized class ID for a value.
+     */
+    pub fn value_to_class_id(&self, value: &Value) -> egraph_serialize::ClassId {
```
saulshanabrook, line 154: Switch to doc comment with `///`

oflatt (approved):
Looks good!

PR #398, opened 2024-07-24
Reporting source locations for errors

### LINKED ISSUE #210: We need better error messages
Currently, error messages refer to generated variables and don't give any source location. This is a big task that we will want to undertake at some time.

yihozhang:
image


src/ast/desugar.rs
```diff
@@ -339,6 +358,7 @@ pub(crate) fn desugar_command(
             };
 
             desugar.desugar_program(
+                // TODO: all spans should be that of the original query
                 desugar.parse_program(None, &desugaring).unwrap(),
```
oflatt, line 362: We could panic when this unwrap fails, and point to the span for the query extract. It would be an okay way to fix the error message here

src/ast/desugar.rs
```diff
@@ -460,16 +481,7 @@ impl Desugar {
     }
 
     pub fn desugar_function(&mut self, fdecl: &FunctionDecl) -> Vec<NCommand> {
```
oflatt, line 483: perhaps we don't need desugar_funciton anymore?

src/ast/mod.rs
```diff
@@ -66,10 +66,90 @@ pub struct SrcFile {
...
+        }
+        Location { line, col }
+    }
+}
+
 /// A [`Span`] contains the file name and a pair of offsets representing the start and the end.
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct Span(pub Arc<SrcFile>, pub usize, pub usize);
 
+impl Span {
```
oflatt, line 135: Why have `Quote` and `Span`? Why not just `Span` with a display that does the same thing as a `Quote`?
Looks like we don't build a quote manually anywhere

src/ast/mod.rs
```diff
@@ -119,16 +203,18 @@ where
...
-    PrintTable(Symbol, usize),
-    PrintSize(Option<Symbol>),
+    PrintTable(Span, Symbol, usize),
+    PrintSize(Span, Option<Symbol>),
     Output {
+        span: Span,
         file: String,
         exprs: Vec<GenericExpr<Head, Leaf>>,
     },
     Push(usize),
```
oflatt, line 213: How did you decide what gets a span? Looks like Push doesn't have one. Perhaps because there are no error messages with it?

oflatt (approved):
Great work! Super exciting to see these working! Had a couple minor changes, feel free to merge after fixing

src/ast/desugar.rs
yihozhang: This would not work in a lot of cases since the failure usually happens at a later stage (typechecking or run time). One way to do it right is to write a `subst_span : Command -> (Span -> Span) -> Command` function that substitutes the spans. However, since this is the only case where we need to give spans right now manually, I just build the AST directly. In the future we can implement this subst_span function if needed.

src/ast/mod.rs
yihozhang: You're right no one should build a quote manually. However, computing `Quote` from `Span` takes O(N) time so `fmt` might not be the best place to do it. However, the original implementation needs to clone the program texts which is costly, so I changed the implementation to take a reference instead.

src/ast/mod.rs
yihozhang: Yeah there's no error message associated with `push` right now. For now I just added spans to commands that may report a `TypeError`.

PR #402, opened 2024-07-30
In REPL, evaluates only when parens are closed

yihozhang:
Currently, the REPL mode evaluates programs line by line, which means `egglog < main.egg` almost certainly won't work because most expressions are multi-lines. Moreover, simply concatenating all lines in the source file (e.g., `tr -d "\n" < main.egg | egglog`) also does not work if the source file contains comments.

This PR fixes this by only evaluating a multi-line block when all the parentheses are closed. So that the program
```
(extract "1
2
3 
)
"
) (extract
";;;""))"
)

```
will only call `parse_and_run_program once` once at the end of the program

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -63,12 +63,12 @@ fn should_eval(curr_cmd: &str) -> bool {
             }
             ';' => {
                 // `any` moves the iterator forward until it finds a match
-                if indices.any(|ch| ch == '\n') {
+                if !indices.any(|ch| ch == '\n') {
                     return false;
                 }
             }
             '"' => {
-                if indices.any(|ch| ch == '"') {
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -45,6 +45,53 @@ struct Args {
     max_calls_per_function: usize,
 }
 
+// Takes a multi-line string, processes the first command, and returns the rest of the string
```
saulshanabrook, line 48: > " returns the rest of the string"

I think the docstring is out of date.

saulshanabrook (commented):
If it isn't too much work, might be nice to add a test or two for this?

yihozhang:
I added a small test

saulshanabrook (approved):

PR #405, opened 2024-08-02
Make dummy span and filenames public

saulshanabrook:
For custom sorts out of core it is useful to have these defined for `make_expr`.

Follow up on #389


yihozhang (approved):
Thanks!

Issue #35, opened 2022-09-05
Nondeterminism


mwillsey:
This is a tracking issue for sources of nondeterminism. @ezrosent ran into one in #33. I _think_ that can be attributed to using `HashMap` instead of `IndexMap` in a couple places, and that is somehow leaking the fact that `Symbols` might be different in the face of concurrency or multiple runs in the same process.

I would like to not have to use `IndexMap`_everywhere_, so I think we should track down why this matters.

PR #406, opened 2024-08-06
Fix determinism bug from new combined ruleset code

oflatt:


yihozhang (approved):
Good catch! Let's merge this once you fix the nits

Related: #35

PR #407, opened 2024-08-06
Move splitting primitives to serialize library

saulshanabrook:
This PR moves the ability to split primitive nodes into the serialize library from this library and also extends it to support splitting arbitrary nodes. See https://github.com/egraphs-good/egraph-serialize/pull/14 for the upstream PR that corresponds to this one, which includes an example using this new behavior.

By default, this should result in the same behavior as before, but just with the ability to split more e-classes besides just primitive nodes if desired.

In order to move the splitting logic upstream, I exposed a way to check if a serialized node ID is a primitive. In a way, this extends https://github.com/egraphs-good/egglog/pull/396 by also now supporting public deserialization/serialization for serialized node IDs as well as class IDs. This should also be useful if you do extraction externally and then want to convert back to egglog values.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -33,6 +33,7 @@ use hashbrown::hash_map::Entry;
 use index::ColumnIndex;
 use instant::{Duration, Instant};
 pub use serialize::SerializeConfig;
+pub use serialize::SerializedNode;
 use sort::*;
 pub use termdag::{Term, TermDag, TermId};
 use thiserror::Error;
@@ -1550,24 +1551,6 @@ impl EGraph {
     /// Gets the overall run report and returns it.
     pub fn get_overall_run_report(&self) -> &RunReport {
         &self.overall_run_report
-    }
-
-    /// Serializes the egraph for export to graphviz.
-    ///
-    /// This will limit the total number of nodes so that visualization does not blow up.
-    pub fn serialize_for_graphviz(
-        &self,
-        split_primitive_outputs: bool,
-        max_functions: usize,
-        max_calls_per_function: usize,
-    ) -> egraph_serialize::EGraph {
-        let config = SerializeConfig {
-            split_primitive_outputs,
-            max_functions: Some(max_functions),
-            max_calls_per_function: Some(max_calls_per_function),
-            ..Default::default()
-        };
-        self.serialize(config)
     }
 
     pub(crate) fn print_msg(&mut self, msg: String) {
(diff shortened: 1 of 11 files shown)


src/lib.rs
```diff
@@ -1552,24 +1553,6 @@ impl EGraph {
...
-        max_functions: usize,
-        max_calls_per_function: usize,
-    ) -> egraph_serialize::EGraph {
-        let config = SerializeConfig {
-            split_primitive_outputs,
-            max_functions: Some(max_functions),
-            max_calls_per_function: Some(max_calls_per_function),
-            ..Default::default()
-        };
-        self.serialize(config)
```
saulshanabrook, line 1570: I removed this help function since it wasn't really doing much

src/main.rs
```diff
@@ -43,6 +43,9 @@ struct Args {
     /// Maximum number of calls per function to render in dot/svg output
     #[clap(long, default_value = "40")]
     max_calls_per_function: usize,
+    /// Number of times to inline leaves
+    #[clap(long, default_value = "0")]
+    serialize_n_inline_leaves: usize,
```
saulshanabrook, line 48: I added the ability to inline leaves via the egglog CLI when making visualizations/json

src/main.rs
```diff
@@ -209,31 +212,25 @@ fn main() {
...
+
+            // if we are splitting primitive outputs, add `-split` to the end of the file name
+            let serialize_filename = if args.serialize_split_primitive_outputs {
+                input.with_file_name(format!(
+                    "{}-split",
+                    input.file_stem().unwrap().to_str().unwrap()
+                ))
+            } else {
+                input.clone()
+            };
```
saulshanabrook, line 233: I moved all this processing into one branch, so it would only be serialized once for JSON or for visualization.

src/sort/set.rs
```diff
@@ -190,6 +190,11 @@ impl Sort for SetSort {
         }
         Some((cost, expr))
     }
+
+    fn serialized_name(&self, _value: &Value) -> Symbol {
+        "set-of".into()
+    }
+
```
saulshanabrook, line 197: This is an unrelated change... It means that when set constructors are serialized they will use this as their name instead of the name of the sort.

src/main.rs
```diff
@@ -1,5 +1,5 @@
 use clap::Parser;
-use egglog::{EGraph, Error, RunMode, SerializeConfig};
+use egglog::{EGraph, Error, RunMode, SerializeConfig, SerializedNode};
```
saulshanabrook, line 2:
```suggestion
use egglog::{EGraph, Error, RunMode, SerializeConfig};
```

yihozhang (approved):

PR #412, opened 2024-08-13
Add power operator to i64 and f64 sorts

DeLectionnes:


saulshanabrook (approved):
Thank you for the tests as well!

DeLectionnes:
After review of the i64 interface in https://doc.rust-lang.org/std/primitive.i64.html, it seems that all power functions take a u32 as exponent instead of another i64.

Should I use the unsafe proposition that the test suite proposes, or should I only leave the f64 part ?

saulshanabrook:
Maybe just leave the f64 then, that seems easiest.

DeLectionnes:
Is there a way to run the test suite on my branch ? it did not transfer when I forked.

saulshanabrook:
@DeLectionnes If you go to the actions tab on your fork you should be able to re-enable github actions

PR #413, opened 2024-08-15
Fix typo in web demo example

tniessen:
Noticed while playing around with egg during Marktoberdorf :)


saulshanabrook (approved):
Thank you!

PR #414, opened 2024-08-15
Fix margin of <select>

tniessen:
Before:

Screenshot 2024-08-15 at 14-49-25 egglog demo

After:

Screenshot 2024-08-15 at 14-50-05 egglog demo


saulshanabrook (approved):
Thanks!

PR #418, opened 2024-08-20
Remove declare and calc

oflatt:
This PR removes declare and calc, since both were not being used widely and were buggy.


yihozhang (approved):
Looks good to me!

PR #426, opened 2024-09-20
Add support for the interactive visualizer

saulshanabrook:
This PR adds the ability to try the [new interactive visualizer](https://github.com/saulshanabrook/egraph-visualizer) in the web editor. I still kept the original graphviz visualizer as default, while this while let you switch between them:

https://github.com/user-attachments/assets/820529fe-d56c-46f7-a5bc-34b5f157b1a3

The new visualizer includes the following features compared to the existing editor:

- Hopefully, better performance when laying out and rendering many nodes
- The ability to focus on a single sub-graph, by clicking on a node and filtering to its descendants
- highlights the connected components to an edge/node on click (could add an on-hover option as well)
- incremental layout based on the existing positions, which the layout engine will take into account
- preserving colors of nodes across multiple consecutive renders, so they will stay consistent
- better edge routing and layout performed by [eclipse layout kernel](https://eclipse.dev/elk/)
- design closer to the examples at the [egg homepage](https://egraphs-good.github.io/), courtesy of [react flow](https://reactflow.dev/) and [tailwind css](https://tailwindcss.com/)

Moreover, it can serve as a base to build future interactive features that would be helpful.

### DIFF
--- a/web-demo/Cargo.toml
+++ b/web-demo/Cargo.toml
@@ -16,7 +16,7 @@ wee_alloc = "0.4.5"
 
 log = "0.4.19"
 wasm-logger = "0.2"
-
+serde_json = "*"
 console_error_panic_hook = "0.1.7"
 js-sys = "0.3"
 wasm-bindgen = "0.2"
--- a/web-demo/Cargo.toml
+++ b/web-demo/Cargo.toml
@@ -16,7 +16,7 @@ wee_alloc = "0.4.5"
 
 log = "0.4.19"
 wasm-logger = "0.2"
-serde_json = "*"
+serde_json = "1.0"
 console_error_panic_hook = "0.1.7"
 js-sys = "0.3"
 wasm-bindgen = "0.2"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1328,6 +1328,7 @@ dependencies = [
  "egglog",
  "js-sys",
  "log",
+ "serde_json",
  "wasm-bindgen",
  "wasm-logger",
  "web-sys",
--- a/web-demo/src/lib.rs
+++ b/web-demo/src/lib.rs
@@ -11,6 +11,7 @@ static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
 pub struct Result {
     pub text: String,
     pub dot: String,
+    pub json: String,
 }
 
 #[wasm_bindgen]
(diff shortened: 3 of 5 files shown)


saulshanabrook:
I have updated this PR to support transitioning between multiple e-graphs, the last thing I wanted to add before calling the visualizer good for now!

https://github.com/user-attachments/assets/561cd4a0-7f49-46ba-a10f-8c24381d24ed

yihozhang (approved):
This looks awesome!! Thank you for making the visualizer better!

web-demo/Cargo.toml
```diff
@@ -16,7 +16,7 @@ wee_alloc = "0.4.5"
 
 log = "0.4.19"
 wasm-logger = "0.2"
-
+serde_json = "*"
```
Alex-Fischman, line 19: nit: specify major version number

saulshanabrook: Thanks!

yihozhang:
This PR seems to make some unintended style changes to the entire web demo (font, font size, alignments, button styles, etc). 

Before:

image

After:

image

saulshanabrook:
Yes, it does! I thought they mainly were fine, except it's just not great when the buttons wrap. That's based on your screen size and the addition of the extra toggle.

I think the changes are from using tailwindcss to style the visualizer and it pulls in some ["pre-flight" styles](https://tailwindcss.com/docs/preflight):

> Built on top of [modern-normalize](https://github.com/sindresorhus/modern-normalize), Preflight is a set of base styles for Tailwind projects that are designed to smooth over cross-browser inconsistencies and make it easier for you to work within the constraints of your design system.

yihozhang:
@saulshanabrook  I see! Thanks for the explanation. Some differences I noticed

1. Font size increases a bit in general
2. The font weight used in the code editor is lighter
3. Spacing between lines in the code editor is wider
4. The buttons and dropdown menus are less recognizable (e.g., no background, no border)
5. The button wraps at the end (although this goes away if I further zoom out).

1 and 2 seem fine, I'm not sure about 3 (the original spacing looks slightly more natural), and I think we should fix 4 and 5. What do you think?

saulshanabrook:
Thanks @yihozhang!

I published a new version that limited the preflight styles to just my component ([using this](https://github.com/Roman86/tailwindcss-scoped-preflight)) and I think it fixed it?

Screenshot 2024-10-16 at 9 09 36 PM


For reference I had to [purge the JS cache](https://www.jsdelivr.com/tools/purge) of these two URLs after publishing to load the new one:

```
https://cdn.jsdelivr.net/npm/egraph-visualizer@2/dist/style.css
https://cdn.jsdelivr.net/npm/egraph-visualizer@2/+esm
```

yihozhang:
Yes, thank you!! @saulshanabrook

PR #427, opened 2024-09-25
Display build info when running egglog

yihozhang:
This is useful when you install egglog to the system via `cargo install --path .` and later want to find out what the exact version you installed is.

After this PR:
```
> cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.16s
     Running `target/debug/egglog`
[INFO ] Welcome to Egglog! (build: 0.2.0_2024-09-25_5bf9dee)
```

and

```
> cargo run -- --version
    Finished dev [unoptimized + debuginfo] target(s) in 0.22s
     Running `target/debug/egglog --version`
egglog 0.2.0_2024-09-25_5bf9dee
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,6 +62,7 @@ im-rc = "15.1.0"
 
 [build-dependencies]
 lalrpop = "0.20"
+chrono = "0.4"
 
 [dev-dependencies]
 glob = "0.3.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,7 +62,7 @@ im-rc = "15.1.0"
 
 [build-dependencies]
 lalrpop = "0.20"
-chrono = "0.4"
+chrono = { version = "0.4", default-features = false, features = ["now"] }
 
 [dev-dependencies]
 glob = "0.3.1"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -62,6 +62,7 @@ im-rc = "15.1.0"
 
 [build-dependencies]
 lalrpop = "0.20"
+chrono = { version = "0.4", default-features = false, features = ["now"] }
```
yihozhang, line 65: I disabled the default feature set because it will otherwise import a complex set of dependencies due to the `locale` feature.

build.rs
```diff
@@ -1,3 +1,15 @@
+use std::{env, process::Command};
+
+#[allow(clippy::disallowed_macros)] // for println!
 fn main() {
     lalrpop::process_root().unwrap();
+    let output = Command::new("git")
```
ezrosent, line 6: Can we make this optional info? (along with the `CARGO_PKG_VERSION` query).

That way we can still run / build this on machines that don't have git, or similar.

ezrosent (changes requested):

yihozhang:
@ezrosent Thanks, I've made the dependency on git optional:

With this PR, if we change `git` in the command to a non-existent command:
```
> cargo run
   Compiling egglog v0.2.0 (/Users/yihongzhang/egglog)
    Finished dev [unoptimized + debuginfo] target(s) in 2.77s
     Running `target/debug/egglog`
[INFO ] Welcome to Egglog! (build: 0.2.0_2024-09-26)
```

ezrosent (approved):

PR #428, opened 2024-10-03
Remove dead code

### LINKED ISSUE #318: Unnecessary arguments to eval_expr and run_actions
Both eval_expr and run_actions take flag `make_defaults`. However, it seems you always want make_defaults to be true, so we can probably remove this flag to make the codebase less confusing. Something to revisit after the ongoing refactoring.

yihozhang:
This PR

* removes the following options: `proofs_enabled`, `terms_enabled`, `match_limit`, and `node_limit`, since they are currently dead code and does nothing.
* removes the `make_defaults` in the `run_actions` function since it is always set to true.

Closes #318


saulshanabrook (approved):
hooray, removing code!

PR #429, opened 2024-10-04
Remove useless type_info getters

Alex-Fischman:
This PR removes the `type_info` and `type_info_mut` from the `EGraph` struct, since having a mutable getter that doesn't do anything else doesn't make sense.

Happy to close this PR if there is a reason for these functions, but only if that reason is documented.


yihozhang (approved):
Thanks! AFAIR, the getter is here because type_info used to live in Desugar, not directly under EGraph

saulshanabrook (approved):
Thank you!

PR #430, opened 2024-10-04
Remove unused options

Alex-Fischman:
This PR removes options that don't do anything. We can add them back later if we need them.


Alex-Fischman:
Closing as a duplicate of #428

PR #432, opened 2024-10-04
Support recursive datatypes in egglog

yihozhang:
I actually accidentally pushed this to `main`... which is bad (now reverted). I have updated repo settings so that no one can directly push to `main`


src/ast/mod.rs
```diff
@@ -878,6 +880,16 @@ where
...
             } => list!("simplify", schedule, expr),
+            GenericCommand::Datatypes { span: _, datatypes } => {
+                let datatypes: Vec<_> = datatypes
+                    .iter()
+                    .map(|(_, name, variants)| match variants {
+                        Ok(variants) => list!(name, ++ variants),
+                        Err((head, args)) => list!("sort", name, list!(head, ++ args)),
+                    })
+                    .collect();
+                list!("datatypes", ++ datatypes)
```
saulshanabrook, line 891: Should this be `datatype*`? Hm, I wonder if there are no tests for these?

yihozhang: Good catch! For `NCommand`, cases like this would be caught by resugared tests (which re-run same tests over programs converted from ASTs), but we don't have tests for Command (which is immediately desugared into `NCommand`)

src/ast/mod.rs
```diff
@@ -450,6 +450,8 @@ pub type Command = GenericCommand<Symbol, Symbol>;
 
 pub type Subsume = bool;
 
+pub type Subdatatypes = Result<Vec<Variant>, (Symbol, Vec<Expr>)>;
```
saulshanabrook, line 453: Are you using `Result` to separate sorts vs datatypes? I think it might be clearer if you have a custom enum instead? Unless I'm missing something, just when trying to follow the code.

yihozhang: Yeah, I changed it to use a custom enum.

src/ast/desugar.rs
```diff
@@ -44,17 +44,59 @@ pub(crate) fn desugar_command(
...
-            &rewrite,
-            subsume,
-        ),
-        Command::BiRewrite(ruleset, rewrite) => desugar_birewrite(
-            ruleset,
-            rewrite.to_string().replace('\"', "'").into(),
-            &rewrite,
-        ),
+        Command::Datatypes { span: _, datatypes } => {
+            let mut res = vec![];
```
saulshanabrook, line 48:
```suggestion
            // first declare all the datatypes as sorts, then add all explicit sorts which could refer to the datatypes, and finally add all the variants as functions
            let mut res = vec![];
```

saulshanabrook: Thought a little comment here might be useful to help follow the desugaring logic

yihozhang: Thank you!

saulshanabrook (approved):

PR #433, opened 2024-10-04
Remove HIGH_COST

Alex-Fischman:
This PR removes `HIGH_COST` for a few reasons:

1. It's unused.
2. It's buggy. On 32-bit systems (such as WASM?) `i64::MAX` will overflow.
3. It's undocumented, so I don't believe it's used outside of this crate. If it is used outside of this crate, I'll edit the PR to use `isize` instead of `i64` to fix the bug, and add documentation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -292,8 +292,6 @@ impl RunReport {
     }
 }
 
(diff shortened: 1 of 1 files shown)


yihozhang (approved):
good catch!

PR #434, opened 2024-10-05
More dead code removal

Alex-Fischman:
This PR removes code related to proofs, on the basis that we don't have proofs yet.


oflatt (approved):
Thanks!

yihozhang (approved):

PR #435, opened 2024-10-06
Static parser

Alex-Fischman:
This PR moves the `ProgramParser` struct out of `Desugar` and into a `lazy_static` block. This has a couple of benefits.

1. Separation of concerns
2. We won't clone the parser (which is apparently slow) when we `push`


oflatt (approved):
Cool

yihozhang (approved):

PR #436, opened 2024-10-07
SymbolGen cleanup

Alex-Fischman:
This PR:

1. Deletes the unused `_special` methods on `SymbolGen`.
2. Deletes the `ResolvedGen` struct and just `impl FreshGen for SymbolGen` twice.
3. Creates one `SymbolGen` in `EGraph` that is used everywhere, instead of constructing multiple `SymbolGen`s which is potentially buggy.

### DIFF
--- a/src/ast/desugar.rs
+++ b/src/ast/desugar.rs
@@ -96,7 +96,7 @@ fn desugar_birewrite(ruleset: Symbol, name: Symbol, rewrite: &Rewrite) -> Vec<NC
 // TODO(yz): we can delete this code once we enforce that all rule bodies cannot read the database (except EqSort).
 fn add_semi_naive_rule(
     desugar: &mut Desugar,
-    fresh_gen: &mut SymbolGen,
+    symbol_gen: &mut SymbolGen,
     rule: Rule,
 ) -> Option<Rule> {
     let mut new_rule = rule;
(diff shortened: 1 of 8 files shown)


oflatt (approved):
Thank you so much for this change! We've had multiple iterations on this and it's great to see it unified.

src/util.rs
```diff
@@ -64,46 +64,30 @@ where
...
-    fn fresh(&mut self, name_hint: &Head) -> Leaf;
-    fn has_been_used(&self) -> bool;
-}
-
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub(crate) struct SymbolGen {
+pub struct SymbolGen {
     gen: usize,
     reserved_string: String,
     special_reserved: HashMap<Symbol, String>,
```
yihozhang, line 71: This field is no longer necessary?

src/ast/desugar.rs
```diff
@@ -11,8 +10,6 @@ pub struct Desugar {
 impl Default for Desugar {
     fn default() -> Self {
         Self {
-            // the default reserved string in egglog is "_"
-            fresh_gen: SymbolGen::new("_".repeat(2)),
```
yihozhang, line 15: The current codebase uses a different default symbol for desugar ("__") and typechecking ("$"). Is it okay to merge them into a single fresh generator? I think while this has no soundness issue, we (either users or developers of egglog) no longer know the provenance of internally generated variables.

src/util.rs
Alex-Fischman: Yep, good catch.

src/ast/desugar.rs
oflatt: We should use the same distinguishing symbol ("$") for both
If we want to know the provenance, we can change the name hint

PR #437, opened 2024-10-07
Delete desugar struct

Alex-Fischman:
This PR cleans up `ast/desugar.rs`, now that we've moved both the parser and the symbol generation out of the `Desguar` struct.
More specifically, this PR:

1. Deletes the `Desugar` struct. (We also discussed doing something like `struct Desugar(&mut SymbolGen)`, but I don't want to do this unless we have at least two things inside the struct.)
2. Deletes the `desugar::get_fresh` function, as I think that we should instead give name hints to `SymbolGen` that indicate where the variable was generated from.
3. Deletes `desugar_commands`, as without the `Desugar` struct this is exactly the same as `desugar_program` (which I've moved to the top of the file as the entry point).

### DIFF
--- a/src/ast/desugar.rs
+++ b/src/ast/desugar.rs
@@ -1,8 +1,18 @@
 use super::{Rewrite, Rule};
 use crate::*;
 
-#[derive(Default)]
-pub struct Desugar {}
+pub(crate) fn desugar_program(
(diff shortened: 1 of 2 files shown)


oflatt (approved):
Much nicer!

PR #438, opened 2024-10-07
Fix extractor nondeterminism

oflatt:
The extractor iterates over the functions map, causing nondeterminism


oflatt:
Also changed rulesets for less brittle future code

Alex-Fischman (approved):
LGTM

PR #441, opened 2024-10-10
Fix `eval_lit` performance bug

Alex-Fischman:
This PR fixes the performance bug in `eval_lit` as noticed by Yihong in [439](https://github.com/egraphs-good/egglog/pull/439#issuecomment-2404072221).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -677,12 +677,13 @@ impl EGraph {
     }
 
     pub fn eval_lit(&self, lit: &Literal) -> Value {
+        let any_sort = self.type_info.infer_literal(lit).as_arc_any();
         match lit {
-            Literal::Int(i) => i.store(&self.type_info.get_sort_nofail()).unwrap(),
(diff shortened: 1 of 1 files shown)


Alex-Fischman:
#439

yihozhang:
Nice! What does the flamegraph now looks like for `eggcc-extraction`?

Alex-Fischman:
I think something's wrong with my flamegraphs, but I'm getting ~10s now compared to ~25s before.

main: before

this branch: after

saulshanabrook:
We didn't get there in the meeting, but would there be interest in adding [codspeed](https://codspeed.io/) tracking? They support flamegraphs in CI and will also show the diff in the flamegraph between a PR and main, showing which parts improved and which parts were degraded.

Alex-Fischman:
I'm keeping this PR open to test with codspeed, but I will close it afterwards because it's subsumed by #442 .

PR #442, opened 2024-10-12
Sort declaration cleanup

Alex-Fischman:
1. Removes `name` from basic sorts.
2. Removes unconditional `get_sort`.
3. Adds `Presort` trait.
4. Removes basic sorts from being stored in primitives.
5. Removes unused and undocumented macros.
6. Makes `Rational` set lazy static.

### DIFF
--- a/src/ast/expr.rs
+++ b/src/ast/expr.rs
@@ -123,9 +123,9 @@ pub enum GenericExpr<Head, Leaf> {
 }
 
 impl ResolvedExpr {
-    pub fn output_type(&self, type_info: &TypeInfo) -> ArcSort {
+    pub fn output_type(&self) -> ArcSort {
         match self {
-            ResolvedExpr::Lit(_, lit) => type_info.infer_literal(lit),
+            ResolvedExpr::Lit(_, lit) => sort::literal_sort(lit),
             ResolvedExpr::Var(_, resolved_var) => resolved_var.sort.clone(),
             ResolvedExpr::Call(_, resolved_call, _) => resolved_call.output().clone(),
         }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,7 +40,7 @@ use thiserror::Error;
 use generic_symbolic_expressions::Sexp;
 
 use ast::*;
-pub use typechecking::{TypeInfo, UNIT_SYM};
+pub use typechecking::TypeInfo;
 
 use crate::core::{AtomTerm, ResolvedCall};
 use actions::Program;
@@ -739,7 +739,7 @@ impl EGraph {
             .get(&sym)
             // function_to_dag should have checked this
             .unwrap();
-        let out_is_unit = f.schema.output.name() == UNIT_SYM.into();
+        let out_is_unit = f.schema.output.name() == UnitSort.name();
 
         let mut buf = String::new();
         let s = &mut buf;
@@ -1367,7 +1367,7 @@ impl EGraph {
             let mut exprs: Vec<Expr> = str_buf.iter().map(|&s| parse(s)).collect();
 
             actions.push(
-                if function_type.is_datatype || function_type.output.name() == UNIT_SYM.into() {
+                if function_type.is_datatype || function_type.output.name() == UnitSort.name() {
                     Action::Expr(span.clone(), Expr::Call(span.clone(), func_name, exprs))
                 } else {
                     let out = exprs.pop().unwrap();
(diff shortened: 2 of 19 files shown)


saulshanabrook (approved):
This generally seems more sensible to me than the current implementation but I also am not that familiar with the original design decisions around how sorts are implemented. I would be curious to hear from whoever implemented it the way it is if this change seems appropriate, is that maybe @mwillsey? But looks good to me!

src/typechecking.rs
```diff
@@ -36,57 +36,47 @@ impl Default for TypeInfo {
...
-        }
-        .unwrap()
-        .clone()
+    pub fn add_sort<S: Sort + 'static>(&mut self, sort: S, span: Span) -> Result<(), TypeError> {
+        self.add_arcsort(Arc::new(sort), span)
     }
 
-    pub fn add_sort<S: Sort + 'static>(&mut self, sort: S, span: Span) {
-        self.add_arcsort(Arc::new(sort), span).unwrap()
+    pub fn add_presort<S: Presort>(&mut self, span: Span) -> Result<(), TypeError> {
```
yihozhang, line 62: I’m less sure about this interface. This means presort (unlike sort) has to be static, and we can’t define the behavior of a presort dynamically. There might be cases where we may want not only sort but presort to be dynamically defined in the future. Examples include presorts `(Database a)` or `(Network a)` that read values of type `a` from an external source, where the presort need to own a database/network object

src/sort/mod.rs
```diff
@@ -102,6 +104,16 @@ pub trait Sort: Any + Send + Sync + Debug {
     }
 }
 
+pub trait Presort {
+    fn name() -> Symbol;
```
yihozhang, line 108: Optional: both Presort and Sort have the `name()` methods. This can be confusing as they are used for different purposes.

src/ast/expr.rs
```diff
@@ -123,9 +123,9 @@ pub enum GenericExpr<Head, Leaf> {
 }
 
 impl ResolvedExpr {
-    pub fn output_type(&self, type_info: &TypeInfo) -> ArcSort {
+    pub fn output_type(&self) -> ArcSort {
```
yihozhang, line 126: Great catch!

src/typechecking.rs
Alex-Fischman: I don't believe that this trait prevents any of that stuff though? All I was trying to do was collect the methods that presorts already depend on in one place.

src/sort/mod.rs
Alex-Fischman: I agree. I was going to use `Presort::presort_name` instead of `Presort::name`, but that conflicts with `Presort::presort_names`. Maybe the problem is `TypeInfo.presort_names`? Is there a better name for that?

src/typechecking.rs
Alex-Fischman: Conclusion: add a comment to Presort explaining that it's currently intended to be used on the same struct as Sort but that can change if e.g. we want partial application in the future.

src/sort/mod.rs
Alex-Fischman: Conclusion: change `presort_names` to `reserved_primitives` and add a check to make sure that primitives don't get reserved twice.

src/typechecking.rs
```diff
@@ -60,12 +60,12 @@ impl TypeInfo {
     }
 
     pub fn add_presort<S: Presort>(&mut self, span: Span) -> Result<(), TypeError> {
```
yihozhang, line 62: could you add a comment to the API?

yihozhang (approved):

Alex-Fischman:
I had to cache the sort names to fix a performance regression with `cykjson`, where the lookup in the symbol table was noticeable. This PR can be merged with the cached names, and I'm going to investigate whether the `Symbol`s are actually helping with performance.

Alex-Fischman:
@yihozhang The ~20% speedup on `eggcc-extraction` is what we expect to see; it matches the speedup in #441. (It's lower than you might be expecting because Saul changed `run 5` to `run 2` to speed up CI in #443.)

saulshanabrook:
Wow nice, that's a huge win!

src/lib.rs
```diff
@@ -1476,11 +1476,6 @@ impl EGraph {
         self.type_info.get_sort_by(pred)
     }
 
-    /// Returns a sort based on the type
```
yihozhang, line 1479: @Alex-Fischman Actually, is there a reason for removing this convenient method (`get_sort`)? Found myself using it today but it went missing.

Alex-Fischman: I thought it was unused, you can add it back

PR #444, opened 2024-10-17
Update Python example 

saulshanabrook:
~~In this PR we change the benchmarks to only run long-running examples instead of all of them.  This is to reduce variance caused by indeterminacy in the memory allocator, which is overly prominent in short benchmarks (https://github.com/oxc-project/backlog/issues/89).~~

It also updates the Python example to more accurately reflect the current workload, including optimizing, and turning into a final Python string, all with CSE (so the size of the file is smaller than before). It should output a final string with something like this:

```python
def __fn(X, y):
    assert X.dtype == np.dtype(np.float64)
    assert X.shape == (150, 4,)
    assert np.all(np.isfinite(X))
    assert y.dtype == np.dtype(np.int64)
    assert y.shape == (150,)
    assert set(np.unique(y)) == set((0, 1, 2,))
    _0 = y == np.array(0)
    _1 = np.sum(_0)
    _2 = y == np.array(1)
    _3 = np.sum(_2)
    _4 = y == np.array(2)
    _5 = np.sum(_4)
    _6 = np.array((_1, _3, _5,)).astype(np.dtype(np.float64))
    _7 = _6 / np.array(float(150))
    _8 = np.zeros((3, 4,), dtype=np.dtype(np.float64))
    _9 = np.sum(X[_0], axis=0)
    _10 = _9 / np.array(X[_0].shape[0])
    _8[0, :,] = _10
    _11 = np.sum(X[_2], axis=0)
    _12 = _11 / np.array(X[_2].shape[0])
    _8[1, :,] = _12
    _13 = np.sum(X[_4], axis=0)
    _14 = _13 / np.array(X[_4].shape[0])
    _8[2, :,] = _14
    _15 = _7 @ _8
    _16 = X - _15
    _17 = np.sqrt(np.array(float(1 / 147)))
    _18 = X[_0] - _8[0, :,]
    _19 = X[_2] - _8[1, :,]
    _20 = X[_4] - _8[2, :,]
    _21 = np.concatenate((_18, _19, _20,), axis=0)
    _22 = np.sum(_21, axis=0)
    _23 = _22 / np.array(_21.shape[0])
    _24 = np.expand_dims(_23, 0)
    _25 = _21 - _24
    _26 = np.square(_25)
    _27 = np.sum(_26, axis=0)
    _28 = _27 / np.array(_26.shape[0])
    _29 = np.sqrt(_28)
    _30 = _29 == np.array(0)
    _29[_30] = np.array(float(1))
    _31 = _21 / _29
    _32 = _17 * _31
    _33 = np.linalg.svd(_32, full_matrices=False)
    _34 = _33[1] > np.array(0.0001)
    _35 = _34.astype(np.dtype(np.int32))
    _36 = np.sum(_35)
    _37 = _33[2][:_36, :,] / _29
    _38 = _37.T / _33[1][:_36]
    _39 = np.array(150) * _7
    _40 = _39 * np.array(float(1 / 2))
    _41 = np.sqrt(_40)
    _42 = _8 - _15
    _43 = _41 * _42.T
    _44 = _43.T @ _38
    _45 = np.linalg.svd(_44, full_matrices=False)
    _46 = np.array(0.0001) * _45[1][0]
    _47 = _45[1] > _46
    _48 = _47.astype(np.dtype(np.int32))
    _49 = np.sum(_48)
    _50 = _38 @ _45[2].T[:, :_49,]
    _51 = _16 @ _50
    return _51[:, :2,]
```


Alex-Fischman:
This seems wrong to me. We want to be able to catch regressions in our whole codebase, not just in 6 slow tests, right?

saulshanabrook:
I think the issue is that the shorter tests have too much variability to be deterministic enough to benchmark. Alternatively we could try a deterministic memory allocator for our benchmarks. There's a thread I link to above which outlines the issue. 

My sense is that if a change doesn't effect our larger benchmarks it's probably fine to pull in... Like as long as they are representative enough of all the code.

Alex-Fischman:
If the large tests are representative then this is a good change; I just didn't think that they were.

Alex-Fischman:
Looking at the codspeed regressions on a different PR (#448) I'm seeing that a lot of the variance comes from the parser.

yihozhang:
Aren't most of the regressions from short-running programs uninteresting because many are one-off costs like parsing, compilation, and bookkeeping? We care more about performance-critical components like query/apply/rebuild, which only become prominent in longer-running programs.

yihozhang:
I think the old python example is a good stress test for the type checker. Maybe we can still keep it but under a different name?

Alex-Fischman:
I think that we should increase the regression threshold back to 10%: look at [this report](https://codspeed.io/egraphs-good/egglog/branches/Alex-Fischman:fast-parser). Even with a faster, less-variance parser, the `rebuild` method still gets significantly slower, and this is not one of the tiny benchmarks (takes >10ms).

saulshanabrook:
Another option is I could "ignore" all benchmarks that aren't long running: https://docs.codspeed.io/features/ignoring-benchmarks/

That way they still get recorded and can be viewed manually but won't influence the reports. 

The only downside here is that if we add additional short examples, they will be have to manually added to the ignore list in the UI.

Alex-Fischman:
That last option seems strictly better than the solution in this PR: opt-out vs. opt-in. Hopefully it would motivate us to create better benchmarks in the future.

saulshanabrook:
> That last option seems strictly better than the solution in this PR to me: opt-out vs. opt-in. Hopefully it would motivate us to create better benchmarks in the future.

OK sounds good! I will update this PR to just add the Python benchmark (and keep the old one) and then just ignore the smaller benchmarks for now :)

tests/python_array_optimize_big.egg
Alex-Fischman, line 1: Why is this one called `big` when it's smaller lol

README.md
```diff
@@ -124,3 +124,17 @@ These primitives are only defined when the result itself is a pure rational.
...
+
+## Benchmarks
+
+We run all of our "examples" [as benchmarks in codspeed](https://codspeed.io/egraphs-good/egglog). These are in CI
+for every commit in main and for all PRs. It will run the examples with extra instrumentation added so that it can
+capture a single trace of the CPU interactions ([src](https://docs.codspeed.io/features/understanding-the-metrics/)):
+
+> CodSpeed instruments your benchmarks to measure the performance of your code. A benchmark will be run only once and the CPU behavior will be simulated. This ensures that the measurement is as accurate as possible, taking into account not only the instructions executed but also the cache and memory access patterns. The simulation gives us an equivalent of the CPU cycles that includes cache and memory access.
+
+Since many of the shorter running benchmarks have unstable timings due to non deterministic behavior, we ["ignore"](https://docs.codspeed.io/features/ignoring-benchmarks/)
```
Alex-Fischman, line 136: Can you clarify "non deterministic behavior"? This makes it sound like egglog will give different results (which I don't believe is true?).

README.md
```diff
@@ -124,3 +124,17 @@ These primitives are only defined when the result itself is a pure rational.
...
+them in codspeed. That way, we still capture their performance, but their timings don't show up in our reports by default.
+
+We use 50ms as our cutoff currently, any benchmarks shorter than that are ignored. This number was selected to try to ignore
```
Alex-Fischman, line 139: Is this automatic or manual? This description makes it sound automatic but the CodSpeed docs above make it sound manual. (If it's automatic no changes are necessary.)

README.md
```diff
@@ -124,3 +124,17 @@ These primitives are only defined when the result itself is a pure rational.
 
 Use double quotes to get a quote: `"Foo "" Bar"` is `Foo " Bar`.
 No primitives defined.
+
+## Benchmarks
```
Alex-Fischman, line 128: nit: this should be above the documentation

Alex-Fischman (approved):
I have quite a few nits but no need for re-review

tests/python_array_optimize_big.egg
saulshanabrook: I believe the file is actually larger, but yeah it's shorter. I can change the name to `large_expr_stresstest` which is the really the purpose it serves now, to test our performance when emitting parsing and type checking a large expression. I avoid this bottleneck now in Python by doing CSE with let bindings.

README.md
saulshanabrook: Manual, I added a clarification that's the case

saulshanabrook:
OK I think this is ready to merge

PR #446, opened 2024-10-18
Add multisets

saulshanabrook:
This PR adds a multiset sort. It is based on a data structure that implements functional sharing. Using that sort, an example is added to show how you can use it to express associative & commutative operations like addition in multiplication with multisets, so that their canonical forms don't need to re-encoded for every ordering. See [these](https://egraphs.zulipchat.com/#narrow/channel/328972-general/topic/Reducing.20node.20explosion.20through.20algebraic.20representations.3F) [threads](https://egraphs.zulipchat.com/#narrow/channel/328972-general/topic/Linear.20and.20Polynomial.20Equations) on zulip for some more background.


src/multiset.rs
```diff
@@ -0,0 +1,86 @@
+use rpds::RedBlackTreeMapSync;
+use std::hash::Hash;
+
+/// Immutable multiset implementation, which is threadsafe and hash stable, regardless of insertion order.
```
saulshanabrook, line 4: I had a go at implementing this data structure in a separate file, keeping its implementation private (following some broad advice from @oflatt and @Alex-Fischman). Hopefully, that should make the code cleaner and make it easier to replace the implementation at a later date. I only exposed the operations needed in the sort so far.

Alex-Fischman: I think it would be better to keep the data structure in the same file as the sort, as that would let you keep everything private instead of `pub(crate)`.

saulshanabrook: The only reason I wanted to put it in a separate file was so that the sort couldn't reach in and use any private methods. Or is that already the case even if I have it in the same file?

src/sort/multiset.rs
```diff
@@ -0,0 +1,440 @@
...
+                && (s.inputs[0].name() == inner_name)
+        });
+        // Only include map function if we already declared a function sort with the correct signature
+        if let Some(fn_sort) = fn_sort {
+            typeinfo.add_primitive(Map {
+                name: "unstable-multiset-map".into(),
+                multiset: self.clone(),
+                fn_: fn_sort,
+            });
+        }
```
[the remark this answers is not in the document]
saulshanabrook: It's strictly necessary, if you look at the example.  It's just as scary as the other unstable higher order function sorts which is why I marked it as unstable.

oflatt:
I'm a little hesitant to add new containers given the churn on the API and the new backend coming soon.

Can this be done in egglog already by using a `Map` and mapping keys to the number of that element?

yihozhang:
I'm in favor of merging this after some cleanups (e.g., merge the two `multiset.rs` files), since this is no worse than the containers and higher-order functions we currently have, and we have not had a clear idea about what will and won't be supported for containers in the future. I especially like the implementation of `map` of this PR. Let's talk about this in the dev meeting this week!

src/multiset.rs
```diff
@@ -0,0 +1,86 @@
...
+        MultiSet(RedBlackTreeMapSync::new_sync())
+    }
+
+    /// Check if the multiset contains a key.
+    pub(crate) fn contains(&self, value: &T) -> bool {
+        self.0.contains_key(value)
+    }
+
+    /// Return the total number of elements in the multiset.
+    pub(crate) fn len(&self) -> usize {
```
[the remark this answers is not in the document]
saulshanabrook: I don't think this would work because we don't just want to count how many entries are in the map, but sum all the values. i.e. this length is will include duplicates.

src/multiset.rs
saulshanabrook: I moved it into the same file, but in its own module to keep its implementation private from the outer module.

Cargo.toml
```diff
@@ -62,6 +62,7 @@ ordered-float = { version = "3.7" }
 getrandom = { version = "0.2.10", features = ["js"], optional = true }
 
 im-rc = "15.1.0"
+rpds = "1.1.0"
```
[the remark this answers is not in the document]
saulshanabrook: I switched to using `im`. Originally I used `rpds` because it seemed like `im` hasn't been maintained much (last commit 2 years ago). But don't feel strongly about it.

I had to actually install `im`, we only had `im-rc` installed, which is the non threadsafe version.

src/multiset.rs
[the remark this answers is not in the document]
Alex-Fischman: Oh, I see. Can we maintain the count as a struct field and assert that this code and that count are the same in debug mode? (Making len non-constant time is evil, worth slightly more complexity imo.)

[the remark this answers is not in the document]
saulshanabrook: Sounds good, resolved.

Cargo.toml
```diff
@@ -62,6 +62,7 @@ ordered-float = { version = "3.7" }
 getrandom = { version = "0.2.10", features = ["js"], optional = true }
 
 im-rc = "15.1.0"
```
yihozhang, line 64: `im-rc` is not used anywhere in the codebase?

yihozhang (approved):
Looks good to me!

Cargo.toml
saulshanabrook: It is, it's just imported as `im_rc` so might be missed on a quick find... It's used in the constraint solver.

PR #450, opened 2024-10-21
Reimplement parser

Alex-Fischman:
This PR implements a fully-Rust parser, which adds better error messages. This has the additional benefits of reducing dependencies and reducing variance in small benchmarks.


yihozhang:
Thanks for this heroic effort! This is amazing! Two thoughts:
* Parser combinator is known to be asymptotically slower than bottom-up parsers like what lalrpop provides. However here we see speedups for several small to medium programs. Have you compared the two parsers with larger programs and see their performance?
* The parser lalrpop generates is very, very large, something like 50k lines of code, so I imagine this PR can greatly reduce the compilation time (especially when one just tweaks the parser a little bit). We can maybe try to get some concrete numbers on this once your PR is ready.

Alex-Fischman:
> * Parser combinator is known to be asymptotically slower than bottom-up parsers like what lalrpop provides. However here we see speedups for several small to medium programs. Have you compared the two parsers with larger programs and see their performance?

These [two](https://github.com/egraphs-good/egglog/pull/450/commits/41db44f3e36eab7f7bc360db38f0cf2c7cd192e7) [commits](https://github.com/egraphs-good/egglog/pull/450/commits/d6de0bc0a26078daef591fff91349e18246a0537) speed up the slowest part of the combinator approach, which are the large multi-way choices. It fixes this by branching on the name of the command/action/schedule, which lets us skip attempting to parse them.

Alex-Fischman:
This parser still doesn't have great error messages, but they are better in some cases. I'm marking as ready for review to get more eyes.

thaliaarchi:
It seems to me that this would be vastly simplified by parsing to a general S-expression structure, then resolving commands according to their identifiers. Parsing and semantic analysis would be decoupled and both would become more consistent.

For parsing: The grammar for a single S-expression would be something like `"(" WS* IDENT (WS+ (":" IDENT WS+)? (sexp | literal))* ")"` (but with proper handling of whitespace; e.g., required between ident and integer literal, but not ident and paren). The code duplication for similar patterns is then eliminated and combinators wouldn't be necessary.

For semantic analysis: Each command would register its signature (e.g., arity, types, optional arguments, or whatever needed), then they'd be resolved by identifier. Since they'd be parsed to a single structure with spans, shared across all commands, then error reporting has far fewer cases. This would allow more invalid programs to be parsed, where, for example, the only problem is in arities or names of optional arguments, and the correct signature could be reported in the error.

If there is no lookahead, each top-level command could be parsed independently with allocations recycled, so the AST for the entire program does not need to be generated.

Alex-Fischman:
I agree that parsing to S-expressions first does almost certainly improve the error reporting somewhat, although I'm not sure how much of a win it actually is; currently error reporting is pretty simple.

I'm not motivated enough to write two parsers for the same language in one week, but if someone does want to take a shot at it you can use my [toy parser from a year ago](https://github.com/Alex-Fischman/egglog-speedrun/blob/main/src/syntax.rs) as a reference/baseline/starting point. It parses a subset of egglog but the basic ideas are there. (`Slice` is just `Span`.)

I do think that merging this PR makes that job slightly easier, since this PR consolidates all of the parsing in one place and does some related cleanup (like removing `Quote`).

Alex-Fischman:
An additional related Zach thought to put on people's radar: an egglog fuzzer would be nice, but it's the sort of thing that's perpetually low priority until it gets done.

src/ast/parse.rs
```diff
@@ -0,0 +1,895 @@
...
+
+    fn span(&self) -> Span {
+        Span(self.source.clone(), self.index, self.index)
+    }
+}
+
+type Res<T> = Result<(T, Span, Context), ParseError>;
+
+trait Parser<T>: Fn(&Context) -> Res<T> + Clone {}
+impl<T, F: Fn(&Context) -> Res<T> + Clone> Parser<T> for F {}
```
ezrosent, line 157: a thought: add `map` as a method here? Might make some of the stuff a bit more idiomatic rust.

Similar for other combinators that take a parser has their first arg and then some other arg, then return another parser. 

I think this is possible to do right now that RPITIT is in.

src/ast/parse.rs
```diff
@@ -0,0 +1,895 @@
...
+struct Location {
+    line: usize,
+    col: usize,
+}
+
+impl SrcFile {
+    pub fn get_location(&self, offset: usize) -> Location {
+        let mut line = 1;
+        let mut col = 1;
+        for (i, c) in self.contents.char_indices() {
```
ezrosent, line 48: optional: 

Should we build an index when constructing `SrcFile` that keeps a btree map mapping from position of a newline to an increasing count of newlines?

That way, we could use the `upper_bound` method to essentially seek to the line in question and compute this all in log(n) time?.

Maybe it's not worth it though. I don't think it'd be that much code though.

ezrosent (approved):
I like this change!

I agree that if we wanted to make the fastest parser possible here we'd probably avoid something like parser combinators, but I think in this case they're reasonable. And to your point we can probably make them reasonably quick for S-expressions. I'm also excited for faster compile times!

I also see the merits of Thalia's suggested architecture. While I personally think this code is fairly easy to digest (the structure of the s-expression kinda shows up when you chain a bunch of `paren`s and `list`s together, even if it's not in the code), I could see having an explicit registration structure making for better error messages.

It's up to you, but I think we could build in some of that more explicit structure incrementally on top of the parser here. I'm good to merge and left some optional suggestions.

src/ast/parse.rs
Alex-Fischman: `get_location` is allowed to be slow because it only ever gets called twice-ish, and only on the error path.

src/ast/parse.rs
Alex-Fischman: Turns out that this breaks compile times, see #468. I will fix this.

Alex-Fischman:
New S-expression based parser at #489.

PR #454, opened 2024-10-28
Fix multiset after `Value.tag`

Alex-Fischman:
The `Value.tag` PR passed CI when it shouldn't have. This PR fixes the compiler error currently on `main`.


Alex-Fischman:
(CodSpeed is just showing the same speedup as on `Value.tag`.)

saulshanabrook (approved):

PR #456, opened 2024-10-28
Update symbol_table

thaliaarchi:
Brings symbol_table to the same version of hashbrown as egglog and has multi-threaded performance improvements. Follows up on [PR#445 Update hashbrown](https://github.com/egraphs-good/egglog/pull/445).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,17 +3,6 @@
 version = 3
 
 [[package]]
-name = "ahash"
-version = "0.7.8"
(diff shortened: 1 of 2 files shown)


thaliaarchi:
Huh. cykjson seems to be very sensitive to hashing.

saulshanabrook (approved):

yihozhang:
We saw regression on cykjson, but we think it is fine because this is just a dependency change and seems unavoidable. Also for context we are also considering removing symbol_table.

PR #457, opened 2024-10-29
`BigInt` and `BigRat`

Alex-Fischman:
This PR adds `BigInt` and `BigRat` sorts by external demand. This PR does not add significant complexity to the codebase, as both `BigInt` and `BigRat` are implemented the same way as we've implemented `Rational`.

As a side effect, this PR also:
- consolidates our dependency on `num` for `BigInt`
- fixes a bug in the `add_primitives` macro to allow for arguments of different types
- adds a very very small test

### DIFF
--- a/tests/bignum.egg
+++ b/tests/bignum.egg
@@ -1,0 +1,5 @@
+
+(let x (bigint -1234))
+(let y (from-string "2"))
+(let z (bigrat x y))
+(check (= (to-string (numer z)) "-617"))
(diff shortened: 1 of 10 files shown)


tests/bignum.egg
```diff
@@ -0,0 +1,5 @@
+
```
yihozhang, line 1: Could you add an example where bigint and bigrat are stored as columns of some tables? The test here does not show what the name of the types are (`BigInt`, `BigRat`)

yihozhang:
Thanks for the quick implementation! cc @rtzam  

# 🐀

yihozhang (approved):

PR #458, opened 2024-10-30
Speed up local `make test` runs

Alex-Fischman:
This PR enables incremental builds for release mode to speed up hot runs of `make test`. It also turns off incremental builds in CI, following the advice of [this blog post](https://fasterthanli.me/articles/why-is-my-rust-build-so-slow#:~:text=Generally%20I%20would%20advise%20setting%20incremental%20%3D%20true%20in%20profile.release%2C%20and%20just%20disabling%20it%20in%20CI%20by%20setting%20the%20environment%20variable%20CARGO_INCREMENTAL%20to%200.).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,7 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
+    - run: echo "CARGO_INCREMENTAL=0" >> "$GITHUB_ENV"
     - uses: actions/checkout@v3
     - uses: taiki-e/install-action@v2
       with:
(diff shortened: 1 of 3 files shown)


saulshanabrook (approved):
Thank you for this!

I did some testing and on main when I run `make test` and then add a newline somewhere, and re-run it, the time to compile is 1m 56s Then with this PR the incremental time is 7s! Compared with the time to run tests and doctests which is only 8s on my machine.



It looks like [our cache action will set this env variable already](https://github.com/marketplace/actions/rust-cache#cache-details:~:text=For%20this%20reason%2C%20this%20action%20automatically%20sets%20CARGO_INCREMENTAL%3D0%20to%20disable%20incremental%20compilation%2C%20so%20that%20the%20Rust%20compiler%20doesn%27t%20waste%20time%20creating%20the%20additional%20artifacts%20required%20for%20incremental%20builds.) but it seems like no harm in making it explicit.

PR #459, opened 2024-10-31
Remove symbol table

saulshanabrook:
This PR removes the symbol_table dependency. Previously it was used to intern strings, so that we could represent them as integers and refer to a global hasmap to find their values.

The current crate to do that, [`symbol_table`](https://github.com/mwillsey/symbol_table/), was created by @mwillsey and his suggestion is that we moved off of it, possibly to a different string interning library.

Instead this PR chooses to not intern any strings, besides those which are values in the e-graph. For those, it uses a similar strategy as other


Alex-Fischman:
Cloning `ArcSort`s is (probably) faster than cloning `String`s, and lets us avoid a lookup in the typechecker (for those symbols which represent sorts). This is sort of an orthogonal change though.

saulshanabrook:
Closing for now, because although this makes things simpler the slowdown is just too large. There are probably smarter ways to deal with it.

If anyone else wants to try something, go for it.

Issue #420, opened 2024-09-04
Disallowing looking up non-constructor functions


yihozhang:
(assigning @FTRobbin)
Currently, we allow looking up non-constructor functions (i.e., functions whose output is not an EqSort or have a `:merge` function) in actions. This can cause actions to fail and make the database end up in an intermediate state.

```
(function f () i64)
(function g () i64 :merge (min old new))
(datatype E)
(function h () E :merge old)
(rule () 
      ((let x (f))))
(rule () 
      ((let x (g))))
(rule () 
      ((let x (h))))
(run 1)
```

Moreover, in principle, we don't want actions in a rule to observe the database, because such an implicit dependency implies that if the database changes, the rule needs to be fired again on the database. This is why our desugaring phase needs special handling for our semi-naive evaluation to be sound.

A fix to this is that we should disallow looking up values of functions in actions. By doing so, we can also remove the special handling for our semi-naive evaluation. Note that we should still allow global actions to look up functions, so the following program should still be allowed:

```
(function f () i64)
(let x (f)) ;; typechecks and throws a run-time error
```

FTRobbin:
This overrides #228. 

This test would be obsolete:
https://github.com/egraphs-good/egglog/blob/757c52f6dcf8fe5b7b5eefb4bdc872ef70a5433a/tests/semi_naive_set_function.egg


Related code: 
Extra desugar that ensures the soundness of the semi-naive evaluation:
https://github.com/egraphs-good/egglog/blob/757c52f6dcf8fe5b7b5eefb4bdc872ef70a5433a/src/ast/desugar.rs#L110

yihozhang:
Fixed in #461

Issue #422, opened 2024-09-06
Renaming `function` whose output is an E-class to `constructor`


yihozhang:
To emphasize the differences between functions whose output is an E-class ("constructor" function), which has unique semantics*, and other functions, we decided to change the syntax for declaring functions whose output is an E-class to use keyword constructor. So what used to be

```
(datatype E)
(function Add (E E) E)
```

should now be written as

```
(datatype E)
(constructor Add (E E) E)
```

Some non-constructors:

```
(datatype E)
(function best (E E) E :merge new) ;; because there is a merge expression
(function add1 (i64) i64) ;; because output is a primitive (and merge expression is an implicit `panic`)
(function upper-bound (E) i64 :merge (max old new)) ;; because it has a merge function
```

We should try to support this new constructor without breaking existing programs. So we should still support existing forms like `(function Add (E E) E)` but issue a warning to prompt the user.

*: constructor functions can be looked up in actions (cf. #420 for non-constructors) where it will be created on the fly (cf. #421 for non-constructors)

oflatt:
I was under the impression that we were making a breaking change- existing programs must re-name to constructor.
`(function Add (E E) E)` would not be supported. I believe our rational was that egglog is still quite young, so we want to make these changes while we can.

yihozhang:
Sure, I think there are three paths we can take

1. Simplify disallowing `(function Add (E E) E)` in our parser.
2. Emitting a nicer error for `(function Add (E E) E)`, rejects the program, and suggests the user use `constructor`.
3. Emitting a warning for `(function Add (E E) E)`, still accepts the program, and suggests the user use `constructor`.

Although my original issue suggested (3), I'm fine with either one. Doing (1) is all good, although (2) and (3) are bonus points that don't take a lot of effort but can improve user experience (and can potentially avoid us some Zulip questions about unexpected errors).

oflatt:
2 would be awesome, it's what I was imagining! A little blurb explaining the difference between the two in an error message would go a long way. Same for the documentation

yihozhang:
fixed by #461

Issue #421, opened 2024-09-06
Removing :default keyword


yihozhang:
We have decided that functions no longer accepts a `:default` argument. In other words, no "non-constructor" functions can no longer be implicitly created. Instead, the user is responsible for initializing functions using `set` explicitly.

This means the following function (from `merge-saturates.egg`) is no longer expressible in egglog, because (1) baz should not have a :default and (2) looking up (baz new) in bar's merge expression is not allowed (related: #420).

```
(function baz (i64) i64 :default 0)

(function bar () i64 :merge (min (baz new) 0))

(set (bar) 1)
(set (bar) 2)

(rule ((= f (baz x)) (< x 50))
      ((set (bar) (+ x 1))))
```

yihozhang:
fixed by #461

PR #461, opened 2024-11-06
Split `function` into `constructor`/`relation`/(custom)`function`; Remove `default`; Disallow `function` lookup in the RHS of a rule

FTRobbin:
This PR fixes Issue #420. Lookup actions in rules will now cause a type error `LookupInRuleDisallowed`.

Move specifically, this PR:

1. Removes `-naive` flag and related desugaring code due to being replaced by this change.

2. Fixes 'fail' failing due to not being identified as global in the `remove_global` rewrite pass.

3. Adds new positive and negative tests for this type error.

4. Rewrites the existing tests for compatibility with the new type error.


src/gj.rs
```diff
@@ -748,27 +748,22 @@ impl EGraph {
                 }
             }
 
-            let do_seminaive = self.seminaive;
             // for the later atoms, we consider everything
             let mut timestamp_ranges =
                 vec![0..u32::MAX; cq.query.funcs().collect::<Vec<_>>().len()];
-            if do_seminaive {
```
yihozhang, line 755: I believe we still want to keep the `-naive` flag as well as the code here, so the user can still do naive evaluation (useful for debugging, also have a different semantics than semi-naive for "unsafe" egglog).

saulshanabrook:
I'm a little worried about the time improvements, especially for lambda... That one is so dramatic I worry that maybe the semantics of the example changed?

Seeing all the changes, I also worry about the degradation for UX, it seems just more unwieldy with this change. 

I know you said that automated desuguring had some issues, but I am wondering if that could be used to at least addressost of these cases? Where there particular issues with it for some cases or just in general?

src/typechecking.rs
```diff
@@ -367,6 +367,19 @@ impl TypeInfo {
         let (query, mapped_query) = Facts(body.clone()).to_query(self, symbol_gen);
         constraints.extend(query.get_constraints(self)?);
 
+        //Disallowing Let/Set actions to look up non-constructor functions in rules
+        for action in head.iter() {
+            match action {
+                GenericAction::Let(_, _, Expr::Call(_, symbol, _)) => {
```
oflatt, line 373: Don't you need to check if this is a function vs a constructor call here?

tests/eggcc-extraction.egg
```diff
@@ -1494,18 +1520,21 @@
...
-      ((set (ival lhs) (BoolI b b))))
+(rule ((= lhs (Node (PureOp (Const (BoolT) (const) (Bool b)))))
+       (= nval (BoolI b b)))
+      ((set (ival lhs) nval)))
 
-(rule ((= lhs (Node (PureOp (Const (IntT) (const) (Num n))))))
-      ((set (ival lhs) (IntI n n))))
+(rule ((= lhs (Node (PureOp (Const (IntT) (const) (Num n)))))
+       (= nval (IntI n n)))
+      ((set (ival lhs) nval)))
```
oflatt, line 1529: `IntI` is a constructor, not a function
So this isn't a lookup and doesn't need to be changed

FTRobbin:
Bumping the PR for review. Now, it properly deals with Issue #420 by enforcing a write-only RHS for rules through a new type-checking pass that checks every subexpression.

This restriction forbids looking up functions even with a default value. This breaks existing tests primarily due to the pervasive use of `relation`. I have manually rewritten the tests to use `set` instead.

I also added a few more test cases for this new pass.

I'm working on #421 to remove default values and #422 to distinguish constructors and functions. Due to the dependencies, I recommend merging after those two are done in this branch.

Known issues:
- merge functions can cause implicit read
- primitives can have side effects (uncertain)

tests/array.egg
```diff
@@ -20,21 +20,21 @@
 (relation neq (Math Math))
 
 (rule ((neq x y))
-      ((neq y x)))
+      ((set (neq y x) ())))
```
oflatt, line 23: Oh I see, now relations have to be set. My opinion is that we shouldn't make users use `set` for relations.

I think the long-term plan is for us to desugar relations to be constructors so people don't have to do this. You could either implement that plan in this PR, or desugar relations to a `set` for now and do that in another PR.

oflatt (changes requested):
Great job on this PR! I had some concerns that need to be addressed before we merge this- the main one has to do with `relation` desugaring

src/gj.rs
FTRobbin: I don't think we should keep it. It is unhelpful and adds complexity to the later passes as they need to support the naive semantics correctly. I am also against keeping it as a use-at-your-own-risk feature.

For Egglog users, if you don't use delete, semi-naive and naive are indistinguishable, so it is unhelpful for debugging. If you use delete, then you care much about performance, and there's no point in using naive. Even when you debug with unsafe features, you should probably debug the semi-naive case instead because that's what you want.

For Egglog developers, I see some value in being a sanity check for ensuring semi-naive is implemented correctly. But we are not doing this now, and it can also be done through stronger end-to-end test cases.

tests/array.egg
FTRobbin: I don't think I'll implement the desugaring in this PR. My impression is that it is still up to discussion what it should be desugared into. Performance overhead aside, desugaring into a constructor might not be ideal because that makes it more expressive than needed. For instance, now you can union two edges in the relation or have a relation as the codomain of a function.

oflatt: That's a good point- we would have to disallow unioning two relation terms.
You can't have a relation as the codomain of a function because the sort would be fresh (no way to get a handle on it)

Relation desugaring aside, I don't think we want to make this breaking change when we plan to fix it right away.

src/gj.rs
oflatt: That was convincing to me. I've never personally used it before
@yihozhang what do you think?

yihozhang: What complexity does this add to later passes? I thought the only difference between seminaive and naive is that in the seminaive case, we split the original query into many small queries depending on the timestamps (i.e., what this code snippet does).

I strongly recommend that we keep the naive evaluation. We can view semi-naive as an optimization of the naive evaluation, and this optimization is not always semantic-preserving, when given bizarre programs that violate certain assumptions. Examples include

* rules that use `extract` / user-defined primitives
* rules where the merge function is not associative or idempotent

I'm also not confident that our semi-naive is implemented correctly- do we really update timestamp every time we update the table? I just looked at table.rs and it seems we don't update the timestamp for at least `get_mut`. The naive evaluation serves as a ground truth for this purpose. Personally, when I am debugging a primitive I wrote, the first thing I do is to disable semi-naive evaluation.

tests/array.egg
yihozhang: Let's make functions whose output is `Unit` a special case: Such functions have default value `()` and an implicit merge function.

yihozhang (changes requested):
Great PR!

tests/array.egg
FTRobbin: I don't think that would work. I think the plan is to get rid of default values altogether. Desugared or not.

yihozhang: I meant getting rid of the `default` field/keyword, but handle unit specially (similar to how EqSort is handled specially).

src/gj.rs
FTRobbin: If we keep the naive flag, either we need to split the latter passes into two, which is unlikely, or each piece of downstream code must support both naive and semi-naive. I am skeptical about the claim that semi-naive code would just work for naive. For one thing, I don't see how semi-naive can be implemented as pure syntactic rewrites. As you pointed out, something more needs to happen to the timestamps in the semi-naive case. And yet, the naive flag is not used anywhere else in the codebase.

There are cases where the two give different semantics. Still, the naive semantics is not more helpful to the users in those cases.

For your last point: Firstly, you still need to debug your new primitive work with semi-naive. Secondly, I will only trust naive evaluation as a ground truth if it is well supported with a clear separation between the two semantics. Relying on your program to be tested to produce test output is a terrible idea to me.

However, I do think this discussion raised a significant concern about the correctness of Egglog. We should investigate the issue.

tests/array.egg
FTRobbin: Conclusions:
- Conceptually: 3 function subtypes
  - Constructor | Relation | Custom
  - Operates differently
- Syntax
  - Keeping the status quo
    - Because there is not enough reason to change it
    - The decision is final for the June vision
  - Adding constructor keyword
    - Only relation if declared as relation
    - Only constructor if declared as constructor
- Semantics/implementation
  - Everything still uses the same function data structure under the hood
  - Only relation and constructor are allowed on RHS
     - It adds the relation/creates the term
  - Union requires both arguments to be the same EqSort
  - Set requires the head to be a custom function

src/gj.rs
FTRobbin: Conclusion: Keep
- Not comprising the comfort of `-naive` for a smaller core
- Too much effort to actually implement `-naive`, we settle for the timestamp hack
- Reconsider when merging the new backend

FTRobbin:
Bumping up this PR again for review:

- Removed `default` keyword, resolving #421.
- Reverted the `-naive` flag change as discussed.
- Implemented splitting `function` into three subtypes: `constructor`/`relation`/(custom)`function`, resolving #420 & #422.
  - `function` is not allowed in the RHS of a rule (`merge` functions are unchecked).
    - `function` can have eqSort as an output.
      - It does not have union as the default merge function
  - `constructor` and `relation` are allowed
    - A `constructor` expression inserts a new enode
      - It has union as the default merge function
    - A `relation` expression inserts a new edge
  - A `constructor`’s output type must be sort
- Reverted the previous changes to tests and then fixed all the tests again
- Added new negative and positive tests

src/actions.rs
```diff
@@ -283,41 +277,25 @@ impl EGraph {
...
+                        } else if function.decl.subtype == FunctionSubtype::Constructor {
+                            let value = Value {
+                                #[cfg(debug_assertions)]
+                                tag: function.schema.output.name(),
+                                bits: self.unionfind.make_set(),
+                            };
+                            function.insert(values, value, ts);
+                            value
+                        } else {
+                            return Err(Error::NotFoundError(NotFoundError(format!(
```
yihozhang, line 294: Nit: this should probably provide a different error message given this PR, since the only case this is possible is when there is a bug in our checker.

src/ast/mod.rs
```diff
@@ -321,7 +345,37 @@ where
...
+    /// For example, here's how to make a sort for vectors
+    /// of some user-defined sort `Math`:
+    /// ```text
+    /// (sort MathVec (Vec Math))
+    /// ```
+    ///
+    /// Now `MathVec` can be used as an input or output sort.
+    Sort(Span, Symbol, Option<(Symbol, Vec<Expr>)>),
+
+    /// Egglog supports three types of functions
```
yihozhang, line 366: nice documentation!

src/ast/mod.rs
```diff
@@ -321,7 +345,37 @@ where
...
+    ///
+    /// A constructor models an egg-style user-defined datatype
+    /// It can only be defined through the `datatype`/`datatype*` command
+    /// or the `constructor` command
+    ///
+    /// A relation models a datalog-style mathematical relation
+    /// It can only be defined through the `relation` command
+    ///
+    /// A custom function is a map
```
yihozhang, line 375: The map part of the definition is a bit weird to me, but it's fine for now since we will need a big documentation refactor anyway.

src/ast/mod.rs
```diff
@@ -356,33 +410,42 @@ where
...
-    /// It can also be used to create
-    /// a container sort.
-    /// For example, here's how to make a sort for vectors
-    /// of some user-defined sort `Math`:
+
+    /// The `constructor` command defines a new constructor for a user-defined datatype
+    /// Example:
     /// ```text
-    /// (sort MathVec (Vec Math))
+    /// (Constructor Add (i64 i64) Math)
```
yihozhang, line 417: nit Constructor -> constructor

src/ast/mod.rs
```diff
@@ -397,44 +460,23 @@ where
...
-    /// - No merge function is provided
-    /// - No default is provided
-    ///
-    /// For example, the following is a datatype:
-    /// ```text
-    /// (function Add (i64 i64) Math)
-    /// ```
-    ///
-    /// However, this function is not:
+    /// Specifically, a custom function can also have an EqSort output type:
```
yihozhang, line 463: nit: this is not an example where a custom function has an EqSort output

yihozhang (approved):
Nice job! Code is clean, with detailed documentation and good tests.

Let us make a release after this PR is merged.

tests/cykjson.egg
```diff
@@ -13,7 +13,7 @@
 (rule ((End a s)
        (= s (getString pos)))
       ((P 1 pos a)
-       (union (B 1 pos a) (T a s))))
```
yihozhang, line 16: Should B be a constructor so that `union` would still work?

tests/intersection.egg
```diff
@@ -27,8 +27,8 @@
 (let t2p (f (f b2)))
 (union t2 t2p)
 
-(union (intersect a1 a2) a3)
-(union (intersect b1 b2) b3)
+(set (intersect a1 a2) a3)
```
yihozhang, line 30: this could just be `union`?

tests/lambda.egg
```diff
@@ -62,20 +62,20 @@
 (function evals-to (Term) Value)
 
 (rule ((= e (Val val)))
-      ((union (evals-to e) val)))
```
yihozhang, line 65: Same here. It's an interesting choice to make evals-to a custom function instead of a constructor. This is a new pattern to me, but it seems to work.

src/actions.rs
FTRobbin: I think it can still be triggered by merge functions reading a table, e.g.:

```
(function foo () i64)

(function bar () i64 :merge (foo))

(set (bar) 0)

(fail (set (bar) 1))
```

tests/lambda.egg
FTRobbin: This is indeed a new pattern. I'll explain more during the meeting.

src/actions.rs
oflatt: Yeah- could add a TODO since this is a known issue
In the future it should be checked and disallowed

tests/intersection.egg
oflatt: Are we disallowing `set` instead of `union` in a follow up PR?

tests/unification-points-to.egg
```diff
@@ -136,23 +136,23 @@
...
+(constructor expr-points-to (ExprT) AllocT)
+(constructor ptr-points-to (AllocT) AllocT)
 
 ;; If `v = malloc(...)`, then `v -> alloc[v]`.
 (rule (
     (assign s t1 v c)
     (malloc c t2)
 )(
-    (union (expr-points-to v) (AllocVar v))
+    (set (expr-points-to v) (AllocVar v))
```
oflatt, line 147: Should these be `union`?

src/ast/parse.rs
```diff
@@ -370,21 +370,33 @@ fn command(ctx: &Context) -> Res<Command> {
                 map(option(text(":unextractable")), |x, _| x.is_some()),
                 map(option(sequence(text(":on_merge"), list(action))), snd),
                 map(option(sequence(text(":merge"), expr)), snd),
```
oflatt, line 372: This doesn't seem right- shouldn't `:merge` be required for functions?
Otherwise, you can use a constructor (or panic for primitives).
Perhaps this should be a follow-up PR.

src/ast/mod.rs
```diff
@@ -99,7 +100,30 @@ where
             GenericNCommand::Sort(span, name, params) => {
                 GenericCommand::Sort(span.clone(), *name, params.clone())
             }
-            GenericNCommand::Function(f) => GenericCommand::Function(f.clone()),
+            // This is awkward for the three subtypes change
```
oflatt, line 103: is this "change" a todo?

src/ast/mod.rs
```diff
@@ -99,7 +100,30 @@ where
...
+                    schema: f.schema.clone(),
+                    cost: f.cost,
+                    unextractable: f.unextractable,
+                },
+                FunctionSubtype::Relation => GenericCommand::Relation {
+                    span: f.span.clone(),
+                    constructor: f.name,
+                    inputs: f.schema.input.clone(),
+                },
+                FunctionSubtype::Custom => GenericCommand::Function {
```
oflatt, line 117: Why call this "Custom"? Why not name it "Function"? Are there other non-function things this could be?

src/ast/mod.rs
```diff
@@ -321,7 +345,37 @@ where
...
+    Sort(Span, Symbol, Option<(Symbol, Vec<Expr>)>),
+
+    /// Egglog supports three types of functions
+    ///
+    /// A constructor models an egg-style user-defined datatype
+    /// It can only be defined through the `datatype`/`datatype*` command
+    /// or the `constructor` command
+    ///
+    /// A relation models a datalog-style mathematical relation
+    /// It can only be defined through the `relation` command
```
oflatt, line 373: Nice- could add "relations are like constructors, but can't be `union`-ed

src/ast/mod.rs
```diff
@@ -960,9 +1145,10 @@ where
     Head: Clone + Display + ToSexp,
     Leaf: Clone + PartialEq + Eq + Display + Hash + ToSexp,
 {
+    // Not used
```
oflatt, line 1148: why does this say not used? Shouldn't clippy catch that problem?

src/typechecking.rs
```diff
@@ -388,13 +388,74 @@ impl TypeInfo {
...
+        Self::check_lookup_actions(&actions)?;
+
         Ok(ResolvedRule {
             span: span.clone(),
             body,
             head: actions,
         })
     }
 
+    fn check_lookup_expr(expr: &GenericExpr<ResolvedCall, ResolvedVar>) -> Result<(), TypeError> {
```
oflatt, line 400: In the future, let's do this check on merge expressions (:

oflatt (changes requested):
Awesome work Haobin! None of my comments are blocking, but they probably require follow-up PR(s)
Will approve once you respond to them

src/ast/parse.rs
FTRobbin: This is not enforced right now. Functions without a merge function will throw an error when set with different values.

src/ast/mod.rs
FTRobbin: The case here is that this change broke the one-to-one correspondence between `GenericCommand` and `GenericNCommand`. I'm not sure what to do with that.

src/ast/mod.rs
FTRobbin: We decided to call it "custom function" to distinguish it from "function," which contains all three subtypes.

src/ast/mod.rs
FTRobbin: I think it is never used now. The old code used this as a catch-all for non-relation functions. Now, the three subtypes are handled separately. However, it is still there to implement `Display`.

oflatt: If it's there to implement display then it's used
Perhaps change the comment to "used in implementation of `Display`"

src/ast/mod.rs
oflatt: ah ok

src/ast/mod.rs
oflatt: I think it's okay, just wondering what you meant here

src/ast/parse.rs
oflatt: Ah okay- so if we don't default to `union` even for eq-able types? That was the old behavior that was confusing

tests/unification-points-to.egg
oflatt: Shouldn't we disallow `set` on constructors?

oflatt (approved):
🎉

saulshanabrook:
I am finally implementing this in Python land and have noticed some weird runtime behavior where certain things are allowed that probably shouldn't be? Like you can union a function that outputs a primtive with a primtive and it can silently pass but just not do anything: https://github.com/egraphs-good/egglog/issues/483#issuecomment-2661487974

I was trying to make the Python types more narrow so that statically you would know like you can't union two primitives. But then I was trying to understand what the rules were and found that there doesn't seem to be many currently...

PR #465, opened 2024-11-10
Remove unused 'serde-1' attribute

DaniPopes:
The feature "serde-1" doesn't exist.


saulshanabrook (approved):
I had noticed this as well. I guess maybe originally there was some hope to make the union find serializable. I think at this point that has been superseded by the e-graph serialization format.

Thank you!

PR #466, opened 2024-11-10
Extract egraph-serialize features

DaniPopes:
`egraph-serialize`'s "serde" and "graphviz" features are not used in the library but are always enabled. Disable them by default and extract them into our own features that just delegate to `egraph-serialize` to preserve the functionality.

This is a breaking change for library users that disable default features.


saulshanabrook (approved):
That makes sense to me, thanks for the warning on the possible downstream breakage.

PR #467, opened 2024-11-10
Disable build script for library builds

DaniPopes:
The build script is only used for emitting the `FULL_VERSION` environment variable, which is only used in the binary. Make it a noop if "bin" is not enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,14 @@ required-features = ["bin"]
 [features]
 default = ["bin"]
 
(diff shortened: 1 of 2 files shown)


DaniPopes:
Yeah and that the build script won't do anything. Unfortunately I don't think there is a way to completely disable the build script, but we can make it to the least amount of work possible.

saulshanabrook (approved):

Alex-Fischman (approved):

PR #469, opened 2024-11-11
Store str instead of String in ParseError

DaniPopes:
Minor performance improvement for the parser. In `repeat_until_end_paren` the `Err` is constructed and discarded immediately after, which allocated a string unnecessarily.


saulshanabrook (approved):

Alex-Fischman (approved):
LGTM!

Issue #468, opened 2024-11-10
Compile time regression


DaniPopes:
Building a simple program takes significantly longer on the latest git versions than on crates.io

```rust
fn main() {
    for path in std::env::args().skip(1) {
        let path = std::path::Path::new(&path);
        let contents = std::fs::read_to_string(&path).unwrap();
        let filename = path.file_stem().unwrap().to_str().unwrap();
        egglog::EGraph::default()
            .parse_and_run_program(Some(filename.to_owned()), &contents)
            .unwrap();
    }
}
```

Clean builds:
- with crates.io v0.3.0: 15s, most of it in dependencies
- with git main 2c8d9472ce510d4f031cc64accd7659c08d13da5: 40s, most of it on egglog

I've bisected it to https://github.com/egraphs-good/egglog/commit/4c280616db0b1bf2bb850aab0075c8f83acd5f33: `cargo build --bin egglog` in this repository goes from 6s to 45s on that commit.

I could reproduce this on 1.79, 1.82 and 1.84.0-nightly (fbab78289 2024-11-04)

Output of `-Ztime-passes`:
```text
time:   0.008; rss:  106MB ->  113MB (   +7MB)  parse_crate
time:   0.000; rss:  113MB ->  114MB (   +1MB)  setup_global_ctxt
time:   0.088; rss:  114MB ->  184MB (  +70MB)  expand_crate
time:   0.088; rss:  114MB ->  184MB (  +70MB)  macro_expand_crate
time:   0.026; rss:  184MB ->  207MB (  +23MB)  late_resolve_crate
time:   0.035; rss:  184MB ->  207MB (  +23MB)  resolve_crate
time:   0.057; rss:  208MB ->  215MB (   +8MB)  unused_lib_feature_checking
time:   0.057; rss:  208MB ->  215MB (   +8MB)  looking_for_derive_registrar
time:   0.063; rss:  208MB ->  218MB (  +10MB)  misc_checking_1
time:   0.073; rss:  218MB ->  275MB (  +57MB)  coherence_checking
time:   0.195; rss:  218MB ->  337MB ( +118MB)  type_check_crate
time:   0.096; rss:  337MB ->  401MB (  +64MB)  MIR_borrow_checking
time:  38.326; rss:  401MB ->  401MB (   +1MB)  MIR_effect_checking
time:   0.027; rss:  401MB ->  401MB (   +0MB)  privacy_checking_modules
[...]

PR #470, opened 2024-11-11
New-type return type of Parser::map

DaniPopes:
Fixes https://github.com/egraphs-good/egglog/issues/468. I'm not sure why this works but it probably has something to do with return-position `impl Trait` in traits having some exponential behavior in rustc.

Compile times for `cargo build --bin egglog` went from 40s to 8s on my machine, with `egglog` (library) now taking about 3s.

### DIFF
--- a/src/ast/parse.rs
+++ b/src/ast/parse.rs
@@ -2,6 +2,7 @@
 
 use crate::*;
 use ordered_float::OrderedFloat;
+use std::marker::PhantomData;
 
 pub fn parse_program(filename: Option<String>, input: &str) -> Result<Vec<Command>, ParseError> {
     let (out, _span, rest) = program(&Context::new(filename, input))?;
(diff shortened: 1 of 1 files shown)


saulshanabrook (commented):
Wow, that's a huge improvement! Thanks for tracing this down.

Could you add a comment to the code maybe linking the issue and explaining the fix so that people who come across this code know why it exists?

Alex-Fischman (changes requested):
My immediate thought is that I would rather simply revert [this commit](https://github.com/egraphs-good/egglog/commit/4c280616db0b1bf2bb850aab0075c8f83acd5f33). It was only changed for [ergonomics](https://github.com/egraphs-good/egglog/pull/450#discussion_r1818445219), and I don't think it's worth either the additional `PhantomData` complexity or the compile time hit.

@ezrosent Do you agree?

ezrosent:
Thanks for the debugging around slow compile times , and for following up in an issue on rustc.

Sorry for the delayed response here. I agree with Alex. I think the added `parse` methods probably outweigh the benefit of having `map` being a method. I think it may be better to revert to map being a function.

PR #471, opened 2024-11-11
Add multiset-sum primitive

RiscInside:
`multiset-sum` combines values from both multisets together, allowing for rules such as
```
(rewrite (Add (Sum  xs) (Sum ys)) (Add (Sum (multiset-sum xs ys))))
```
In my understanding, this primitive is crucial for modelling associative + commutative operations with multisets, but I might be missing something.

### DIFF
--- a/src/sort/multiset.rs
+++ b/src/sort/multiset.rs
@@ -95,6 +95,14 @@ mod inner {
             }
             multiset
         }
+
+        /// Compute the sum of two multisets.
+        pub fn sum(self, MultiSet(other_map, other_count): Self) -> Self {
+            Self(
+                self.0.union_with(other_map, std::ops::Add::add),
+                self.1 + other_count,
+            )
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


saulshanabrook (approved):
Thank you for this!

Yeah it was not intentional... I just surfaces the primitives necessary to recreate an example to start and figured they could grow based on user need. 

So happy for more PRs to expose more things, including more holistically.

RiscInside:
I did also consider adding other operations (`union`, `intersection`, and `difference`), but I don't immediately see a use-case for any of them  (`intersection`/`union` also have a problem of not being fragile to introducing equalities, e.g. consider `{e1, e1, e2} union {e1, e2, e2}` when we later discover that `e1 = e2`)

saulshanabrook:
Sounds good.

Multiset should work with rebuilding canonicalization, but it would be good to add test for this (not related to the work in this PR necessarily). Meaning that if you have {a: 1, b: 1} and then a and b are unioned, the resulting multiset should be {(a | b): 2}. That is handled in `canonicalize`.

PR #472, opened 2024-11-18
Make "declared function" printouts debug mode

oflatt:
The eggcc project relies on info printing, and the declared function print outs are becoming super verbose (one per variable, which we generate tens of thousands of)
This PR makes these prints debug instead of info

It also slightly cleans up a confusing error message in the extractor

### DIFF
--- a/src/extract.rs
+++ b/src/extract.rs
@@ -66,7 +66,10 @@ impl EGraph {
                     }
                 }
 
-                panic!("No cost for {:?}", value)
+                panic!(
+                    "Failed to extract value {:?}. Did you use :unextractable constructors or the delete keyword?",
+                    value
+                )
             })
     }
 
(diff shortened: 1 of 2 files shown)


yihozhang:
I'm not sure if this should be in the `main`. Right now `info` is mostly used for responding to user commands (i.e., if it succeed), while `debug` is for recording internal processing information. We should generate at least one `INFO` message per user command. If you want a less overwhelming log, you should probably make egglog print desired log info to `stdout` via some commands, and then redirect egglog's `stdout` to e.g., a log file

oflatt:
@yihozhang is this for the web demo? Or for sidecar egglog?
Sidecar egglog shouldn't use info prints, it should use the normal stdout right?

yihozhang:
Sidecar egglog only uses stdout for communication. `log::info` is for CLI and web demo. I believe one of the original intentions of `log::info` is so that the user can see when each command has finished running (otherwise how would a REPL user know?), while stdout should be silent until the user expects a result (e.g., an extractant, run statistics). This is not documented.

oflatt:
Racket's REPL doesn't print out something on each command, it just runs it for example. It's pretty verbose to print on each command unless you need that interactivity for sidecar egglog.
Perhaps we should have a new interactive mode that is verbose and parse-able?

PR #474, opened 2024-12-02
Fix the missing `include_subsume` flag so that subsumed tuples does not show up for rules

yihozhang:
To see how this bug can cause failures, comment out the lines where `from_column_index` is used (which is hiding the bug) and use `LazyTrie::default` instead and run the tests again.


oflatt (approved):
Oof, good catch
Was this making eggcc slower?

yihozhang:
I'm not sure. We can bump eggcc's egglog version and see if it's faster.

PR #475, opened 2024-12-02
Remove Rational

Alex-Fischman:
Removes the `Rational` type, as the semantics are unclear and we have `BigRat` now. I'm working on the `egglog-experimental` repo at the moment and this was the first feature we decided to move. The tests that this PR changes/removes all still exist in that repo.


Alex-Fischman:
I don't think the `herbie` tests make sense without `Rational`, but the python-generated tests can presumably be regenerated from something.

oflatt:
Are you moving these tests to the egglog experimental repo? We should still run them, and run them with the same checks

Alex-Fischman:
> Are you moving these tests to the egglog experimental repo? We should still run them, and run them with the same checks

Yes, see top of PR and current main of `egglog-experimental`.

oflatt:
Oh I see
I'd like to see a PR I can review in the experimental repo
For example, shouldn't we share the test harness code between the two repositories? Later when we add stronger testing it'll have to be duplicated otherwise

Alex-Fischman:
There's no PR because it would just be the entire repo. I can draft one if you want?

@yihozhang's opinion is that duplicating the tests between the two PRs is more trouble than it's worth, and that we can just split the tests between the PRs. We don't expect much overlap anyway.

oflatt:
An initial PR I can look over real quick would be nice
I don't want to duplicate the tests themselves, I just want to share the `files.rs` test harness

Another thing we should decide about is if we run all the tests in the CI for both repos or not

Alex-Fischman:
Here's a diff: https://github.com/egraphs-good/egglog-experimental/compare/cf9ed6bec2ebfd3bbfed2d1b37a8c217688a8b8d...main

I don't know what draft PR you want me to make, sorry for being unclear.

oflatt:
I know this is going to be annoying, but my suggestion is:
- clear the egglog-experimental repo
- make a PR of the current code to egglog-experimental
I feel that if you commit directly to egglog-experimental you circumvent the review process.

As for files.rs, I suggest that we have egglog provide the function that produces tests based on an egglog file, and have both crates share that code.

Alex-Fischman:
Okay, I've reset the repo and made the PR, assigning you as a reviewer.

I've put factoring out `files.rs` on my todo list.

saulshanabrook:
Should I try to regenerate `python_array_optimize` based on this PR removing rat? I would have to make a PR to egglog python removing the rat sort and then rewrite the code to use a different sort.

Alex-Fischman:
I've changed the existing tests to just use `BigRat` instead of `Rational`.

oflatt (approved):
Looks good

PR #478, opened 2024-12-05
Subsume should succeed even when the tuple is not present in the database

### LINKED ISSUE #462: Subsumption confusing semantics
Subsumption currently works kind of like a query and kind of like an action. For example, this code:

`(Subsume (Add (Add 2 3) 4))`

Adds `(Add 2 3)` to the database. However, if `(Add (Add 2 3) 4)` isn't present, it errors out because it couldn't find an existing entry.

Is this the behavior we want? Perhaps we wanted subsumption to first add the tuple, then subsume it?

yihozhang:
Fixes #462

### DIFF
--- a/src/function/table.rs
+++ b/src/function/table.rs
@@ -124,13 +124,6 @@ impl Table {
         Some(&self.vals[off].1)
     }
 
-    pub(crate) fn get_mut(&mut self, inputs: &[Value]) -> Option<&mut TupleOutput> {
-        let hash: u64 = hash_values(inputs);
-        let &TableOffset { off, .. } = self.table.find(hash, search_for!(self, hash, inputs))?;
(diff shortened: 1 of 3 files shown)


src/function/table.rs
```diff
@@ -124,13 +124,6 @@ impl Table {
         Some(&self.vals[off].1)
     }
 
-    pub(crate) fn get_mut(&mut self, inputs: &[Value]) -> Option<&mut TupleOutput> {
```
yihozhang, line 127: `get_mut()` is unsafe because it does not require the caller to update the timestamp. Use `insert_and_merge` instead.

yihozhang:
Actually this fix is bad. Will re-do it after #461 is merged since the right fix will have merge conflict with #461

PR #484, opened 2024-12-06
Fix display implementation for DUMMY_SPAN

RiscInside:
Calling `Display` on `DUMMY_SPAN` used to crash the program in debug mode due to underflow. No longer.


yihozhang (approved):
Great!

PR #485, opened 2024-12-07
Remove :on_merge

Alex-Fischman:
This PR removes `:on_merge`. Oliver says we don't need it.

It also removes `:cost` and `:unextractable` from `function`s, since they can no longer ever be extracted (right?).

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -119,9 +119,6 @@ where
                     schema: f.schema.clone(),
                     name: f.name,
                     merge: f.merge.clone(),
-                    merge_action: f.merge_action.clone(),
-                    cost: f.cost,
-                    unextractable: f.unextractable,
                 },
             },
             GenericNCommand::AddRuleset(name) => GenericCommand::AddRuleset(*name),
@@ -481,9 +478,6 @@ where
         name: Symbol,
         schema: Schema,
         merge: Option<GenericExpr<Head, Leaf>>,
-        merge_action: GenericActions<Head, Leaf>,
-        cost: Option<usize>,
-        unextractable: bool,
     },
 
     /// Using the `ruleset` command, defines a new
@@ -745,9 +739,6 @@ where
                 name,
                 schema,
                 merge,
-                merge_action,
-                cost,
-                unextractable,
             } => {
                 let mut res = vec![
                     Sexp::Symbol("function".into()),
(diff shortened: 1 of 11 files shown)


Alex-Fischman:
Whoa, that's a suspiciously large speedup.

yihozhang:
@Alex-Fischman yeah, I think all the functions in python_array_optimize except for `Program_parent` should actually be constructors.

yihozhang:
I think this means we need to ban functions without merge expressions ASAP, since this causes programs to silently fail under the current main branch.

Alex-Fischman:
Okay, I've required functions to either contain `:merge` or `:no-merge`.

saulshanabrook:
So functions can never be extracted because... Their outputs always have something else in their e-class? I.e. they have to be set to a real value?

Like if I have a function that returns an eqsort, couldn't it not have anything else equal to it?

I guess I'm confused why they should always be in extractable. 

I think in the Python code and all my use cases this if fine though, so I have nothing against it in practice.

Alex-Fischman:
The argument is that `function`s should never be extracted because semantically, that's not what they're for. `function`s are for analysis. If you wanted it to be extractable, you would've used a `constructor`.

saulshanabrook:
What's the  difference again between functions and constructors whose return type is an eqsort?

Like you can never union a function?

Alex-Fischman:
`union` operates on values so that has to still work. The difference is that you can only write `(name a b)` on the right-hand side of a rule if `name` is a constructor or relation. This was part of Haobin's 3-in-1 PR.

The actual semantic difference between functions that return an eq-able sort and constructors is still somewhat up in the air. If this PR gets merged then it codifies some of these differences. We briefly talked about this on Thursday (see the last two hollow bullets under the first agenda item).

Hopefully that helps.

src/ast/mod.rs
```diff
@@ -1062,18 +1036,14 @@ impl FunctionDecl {
...
         Self {
             name,
             subtype: FunctionSubtype::Custom,
             schema,
             merge,
-            merge_action,
-            cost,
-            unextractable,
+            cost: None,
+            unextractable: false,
```
yihozhang, line 1046: should this be true?

src/ast/mod.rs
```diff
@@ -1108,7 +1077,6 @@ impl FunctionDecl {
                 output: Symbol::from("Unit"),
             },
             merge: None,
-            merge_action: Actions::default(),
```
yihozhang, line 1111: similarly below, should unextractable be true?

yihozhang (approved):
Nice! This PR simplifies things a lot, thanks!

PR #486, opened 2024-12-09
Add `egraph_interact()` debugging tool

gussmith23:
I've been using something like this while developing Churchroad, and I figured it wouldn't hurt to see if you all want it in egglog itself. Let me know -- happy to just keep it local, and happy to change it however you want.

### DIFF
--- a/src/debugging.rs
+++ b/src/debugging.rs
@@ -1,0 +1,79 @@
+//! egglog debugging tools.
+
+use std::io::{stdin, stdout, BufRead, Write};
+
+use crate::EGraph;
+
+/// Allows user to directly interact with the egraph via commands on the command
+/// line, for debugging purposes. egglog commands are taken over stdin, and
+/// results are printed to stdout. Ctrl+D exits the debugging session.
+///
+/// The following example uses [`egraph_interact_via`] for testing purposes, but
+/// the same code will work with [`egraph_interact`].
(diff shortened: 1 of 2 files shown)


Alex-Fischman:
I'm not sure I understand; how is this different from the REPL? (The REPL is started by running egglog with no input files.)

gussmith23:
Yes, sorry, I should have given an example use case.

There are two primary uses I can think of for this tool. First, when you're
building a standalone binary that *uses* egglog internally, and second,
when you're using any "advanced" feature of egglog that can't be accessed
from the text format (and thus you must use Rust).

In both cases, you're using egglog as a Rust library. As soon as you need
to use your own custom Rust, I'm not sure how to enter the REPL without
something like this -- but maybe there's a way to do it! If there is,
something like this isn't needed.

For example, Churchroad is egglog-based, but is its own standalone Rust
binary that uses egglog as a library. Within main.rs, I do things to the
egraph that currently can't be done in text format alone (eg custom
primitives). But at various points I still want a REPL to debug. This tool
provides that!

Again, if there's a way to drop into the egglog REPL itself from Rust, let
me know. This is essentially just a tiny imitation of the REPL exposed for
debugging.

Alex-Fischman:
Closing because #507 subsumes this PR.

PR #489, opened 2024-12-10
S-Expression based parser

### LINKED ISSUE #488: Unintuitive parser errors
```
❯ echo "(datatype E) (function add (E E) E :unextractable)" | egglog
[INFO ] Welcome to Egglog! (build: 0.3.0_2024-12-10_9684208)
[ERROR] In 1:36-44: :unextrac
    expected ":no-merge"
```
I think we should make our parser token-based rather than character-based

Alex-Fischman:
Fixes #488.

Implements an S-expression based parser, greatly improving error messages in the process. For example, we now print usage strings if the parser is unable to recognize a specific command.

### DIFF
--- a/src/ast/expr.rs
+++ b/src/ast/expr.rs
@@ -6,7 +6,7 @@ use std::{fmt::Display, hash::Hasher};
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
 pub enum Literal {
     Int(i64),
-    F64(OrderedFloat<f64>),
+    Float(OrderedFloat<f64>),
     String(Symbol),
     Bool(bool),
     Unit,
@@ -33,14 +33,14 @@ macro_rules! impl_from {
 }
 
 impl_from!(Int(i64));
-impl_from!(F64(OrderedFloat<f64>));
+impl_from!(Float(OrderedFloat<f64>));
 impl_from!(String(Symbol));
 
 impl Display for Literal {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match &self {
             Literal::Int(i) => Display::fmt(i, f),
-            Literal::F64(n) => {
+            Literal::Float(n) => {
                 // need to display with decimal if there is none
                 let str = n.to_string();
                 if let Ok(_num) = str.parse::<i64>() {
--- a/src/ast/parse.rs
+++ b/src/ast/parse.rs
@@ -846,19 +846,27 @@ fn sexp(ctx: &mut Context) -> Result<Sexp, ParseError> {
             Token::String(s) => Sexp::Literal(Literal::String(s), span),
             Token::Other => {
                 let s = span.string();
-                let int = s.parse::<i64>();
-                let float = s.parse::<f64>();
-                match s {
-                    "true" => Sexp::Literal(Literal::Bool(true), span),
-                    "false" => Sexp::Literal(Literal::Bool(false), span),
-                    _ if int.is_ok() => Sexp::Literal(Literal::Int(int.unwrap()), span),
-                    "NaN" => Sexp::Literal(Literal::Float(OrderedFloat(f64::NAN)), span),
-                    "inf" => Sexp::Literal(Literal::Float(OrderedFloat(f64::INFINITY)), span),
-                    "-inf" => Sexp::Literal(Literal::Float(OrderedFloat(f64::NEG_INFINITY)), span),
-                    _ if float.is_ok() && float.as_ref().unwrap().is_finite() => {
-                        Sexp::Literal(Literal::Float(OrderedFloat(float.unwrap())), span)
+
+                if s == "true" {
+                    Sexp::Literal(Literal::Bool(true), span)
+                } else if s == "false" {
+                    Sexp::Literal(Literal::Bool(false), span)
+                } else if let Ok(int) = s.parse::<i64>() {
+                    Sexp::Literal(Literal::Int(int), span)
+                } else if s == "NaN" {
+                    Sexp::Literal(Literal::Float(OrderedFloat(f64::NAN)), span)
+                } else if s == "inf" {
+                    Sexp::Literal(Literal::Float(OrderedFloat(f64::INFINITY)), span)
+                } else if s == "-inf" {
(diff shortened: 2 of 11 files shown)


Alex-Fischman:
Note: since this parser isn't character based I had to fix `cyk.egg` to change
`(query-extract:variants 10  test2)`
to
`(query-extract :variants 10 test2)`.

I think this is obviously good but wanted to put it on the record.

saulshanabrook:
I don't think the performance slowdown here is a deal breaker since the stress test is a bit of an edge case and the slowdowns isn't that bad.

Alex-Fischman:
Okay, regression fixed! Now waiting for reviews.

src/ast/expr.rs
```diff
@@ -6,7 +6,7 @@ use std::{fmt::Display, hash::Hasher};
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
 pub enum Literal {
     Int(i64),
-    F64(OrderedFloat<f64>),
+    Float(OrderedFloat<f64>),
```
saulshanabrook, line 9: 👍

src/ast/parse.rs
```diff
@@ -4,16 +4,14 @@ use crate::*;
 use ordered_float::OrderedFloat;
 
 pub fn parse_program(filename: Option<String>, input: &str) -> Result<Vec<Command>, ParseError> {
-    let (out, _span, rest) = program(&Context::new(filename, input))?;
-    assert!(rest.is_at_end(), "did not parse entire program");
-    Ok(out)
+    let sexps = all_sexps(Context::new(filename, input))?;
```
oflatt, line 7: Do you mind moving the sexp parser into it's own file?

oflatt (approved):
Did a quick skim of the changes, they look pretty good to me!
negative diff is a great sign
Recommended splitting parse.rs into two

src/ast/parse.rs
Alex-Fischman: Hmm, a little bit. Where should `Span` live? It's under 1000 lines and we have way worse offenders, whereas everything here (hopefully) is only in service of implementing `parse_program` and `parse_expr`.

oflatt: Yeah idk where Span should live
ok

src/ast/parse.rs
```diff
@@ -94,801 +92,785 @@ impl Display for Span {
...
+                let mut string = String::new();
+
+                loop {
+                    match self.current_char() {
+                        None => {
+                            span.2 = self.index;
+                            return error!(span, "string is missing end quote");
+                        }
+                        Some('"') if !in_escape => break,
+                        Some('\\') if !in_escape => in_escape = true,
```
yihozhang, line 782: Nit: This means the only escaped characters are `\"` and `\\`. Maybe we can reject other escaped characters like `\n`, `\t`? This can help prevent user's misunderstanding.

src/ast/parse.rs
```diff
@@ -94,801 +92,785 @@ impl Display for Span {
...
+                    return error!(span, "unexpected `)`");
+                }
+                let (mut list_span, list) = stack.pop().unwrap();
+                list_span.2 = span.2;
+                Sexp::List(list, list_span)
+            }
+            Token::String(s) => Sexp::Literal(Literal::String(s), span),
+            Token::Other => {
+                let s = span.string();
+                let int = s.parse::<i64>();
```
yihozhang, line 843: nit: this logic is not wrong but looks a bit weird

src/ast/parse.rs
```diff
@@ -94,801 +92,785 @@ impl Display for Span {
...
+    slice: &[Sexp],
+    func: impl Fn(&Sexp) -> Result<T, ParseError>,
+) -> Result<Vec<T>, ParseError> {
+    slice.iter().map(func).collect::<Result<_, _>>()
 }
 
-fn snd<T, U>(x: Option<(T, U)>, _span: Span) -> Option<U> {
-    x.map(|(_, x)| x)
-}
+fn command(sexp: &Sexp) -> Result<Command, ParseError> {
```
yihozhang, line 183: Make this function an implementation of some trait, so that we can support user-defined keywords?

yihozhang (commented):
This looks great! Thanks for the hard work!

Besides my comments, I also have a question: Do you think if the current implementation of `command` is too coarse-grained? 

The user may want to say, "use my desugaring for `rewrite` and `rule` instead of the standard one". Some use cases include, say, when someone implements a macro-based proof support or tree automata refinement, they want their macros to work for existing egglog programs (without inventing a new keyword). So they may want to directly hack the parser and use their own desugaring logic for standard keywords like `rule`.

src/ast/parse.rs
Alex-Fischman: This is future work

src/ast/parse.rs
```diff
@@ -94,801 +92,785 @@ impl Display for Span {
...
+}
 
-    fn advance_char(&mut self) {
-        assert!(self.index < self.source.contents.len());
-        loop {
-            self.index += 1;
-            if self.source.contents.is_char_boundary(self.index) {
-                break;
-            }
+enum Sexp {
```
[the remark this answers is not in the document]
Alex-Fischman: This is future work

Alex-Fischman:
@yihozhang I've implemented the comments that I felt were appropriate for this PR. Note that this PR is scoped to be a strict improvement over the existing parser, without adding features, so I've avoided making things public and generalizing to traits prematurely.

To respond to your question, I am still not sure whether the syntax plugin system will be `Sexp -> Sexp` or `Sexp -> Command`. However, I think that for either case, `command` as it is currently will work fine:
- For `Sexp -> Sexp`, it doesn't matter what `command` does, as the plugin will be able to rewrite keywords before `command` sees them.
- For `Sexp -> Command`, I was thinking that immediately after the `expect_call("command")`, we would look up the `head` in the plugin list to see if it was registered. Then we would either run the plugin or proceed to the big `match`. This would allow the user to override the default behaviour.

Does this address your concerns?

yihozhang (approved):
Either is reasonable design IMO!

ezrosent (approved):
Big diff but pretty mechanical overall; fairly easy to review. Nice improvement!

PR #490, opened 2024-12-11
Use SmallVec::from_slice instead of .into()

DaniPopes:
Small performance gain since the generic From/Extend implementations do not (cannot) specialize on Copy vs Clone types.


saulshanabrook (approved):
I don't see any real change in the benchmarks from this (https://codspeed.io/egraphs-good/egglog/branches/DaniPopes%3Asmallvec-from-slice) but seems small enough to be fine to add.

PR #491, opened 2024-12-11
Implement search_for! using functions

DaniPopes:
Avoid unnecessary monomorphization due to macro inlining of closures. This can't always be a method because of double borrows when searching mutably.

### DIFF
--- a/src/function/table.rs
+++ b/src/function/table.rs
@@ -55,22 +55,6 @@ pub(crate) struct Table {
     pub(crate) vals: Vec<(Input, TupleOutput)>,
 }
 
-/// Used for the HashTable probe sequence.
-macro_rules! search_for {
-    ($slf:expr, $hash:expr, $inp:expr) => {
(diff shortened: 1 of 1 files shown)


saulshanabrook (approved):

src/function/table.rs
```diff
@@ -235,7 +220,10 @@ impl Table {
     /// removed.
     pub(crate) fn remove(&mut self, inp: &[Value], ts: u32) -> bool {
         let hash = hash_values(inp);
-        let Ok(entry) = self.table.find_entry(hash, search_for!(self, hash, inp)) else {
+        let Ok(entry) = self
+            .table
+            .find_entry(hash, search_for(&self.vals, hash, inp))
```
yihozhang, line 225: this can just use `self.search_for`? (similarly above)

DaniPopes: It would cause a double borrow error

PR #492, opened 2024-12-11
Test extraction of vecs & allow disabling messages for performance

saulshanabrook:
This PR adds a test so that we can benchmark extracting an expression with vecs in it. This was recently discovered as a performance bottleneck. Hopefully adding the benchmark should let us resolve the extraction issue.

EDIT: This also lets us disable saving/printing messages (for use cases like Python where you use the ExtractReport instead of the messages).

EDIT 2: It also changes the benchmarks to run the Python generated examples with messages disabled to better track this performance.


RiscInside:
> It looks like this test fails before it finishes, maybe due to resource constraints of the runner?
> 
> ```
> SIGTERM [  81.248s] egglog::files extract_vec_bench
> ```

I had a similar issue on my laptop. I believe this can be caused by quadratic blowup in `TermDag::to_string`. I made a fix - https://github.com/egraphs-good/egglog/pull/496

saulshanabrook:
This is ready for review!

tests/extract-vec-bench-no-messages.egg
Alex-Fischman, line 1: Can you move this into a folder called no-messages? I don't expect the test name to matter, while I do expect the folder name to matter, so only having no-messages in the test name is a footgun for me.

src/actions.rs
```diff
@@ -339,9 +339,11 @@ impl EGraph {
                     let variants = values[1].bits as i64;
                     if variants == 0 {
                         let (cost, term) = self.extract(values[0], &mut termdag, sort);
-                        let extracted = termdag.to_string(&term);
-                        log::info!("extracted with cost {cost}: {extracted}");
-                        self.print_msg(extracted);
+                        if self.messages_enabled() {
```
Alex-Fischman, line 342: Why do you have these guards? `print_msg` won't do anything if messages are disabled, and the `to_string` PR is in already, right?

src/main.rs
```diff
@@ -37,6 +37,8 @@ struct Args {
     /// Number of times to inline leaves
     #[clap(long, default_value = "0")]
     serialize_n_inline_leaves: usize,
+    #[clap(long)]
+    dont_print_messages: bool,
```
Alex-Fischman, line 41: Optional: should this be called `quiet` or `silent`?

Alex-Fischman (approved):
I had a couple of nits but it looks fine overall. Can you explain more (in this PR discussion) why we need to disable messages though?

src/actions.rs
RiscInside: I think it depends on the current goal for adding this benchmark. 

If the goal is to benchmark extraction itself (which still takes 400 ms - not as pathologically bad as it used to be, but arguably still a lot for a problem of this side), it might be helpful to disable stratification. Converting dag to a string still dominates runtime - it takes around a second and generates 889mb long string.

If the goal is to verify that pathologically bad behaviour doesn't occur anymore, it's probably okay to re-enable message printing.

In any case, having the ability to disable message printing is quite valuable in itself.

saulshanabrook: Yeah, this PR mixes two things:

1. Allowing the e-graph to be run without collecting any messages and omits turning a term dag into a string
2. Run some benchmarks with this mode

The first is useful for situations where you are embedding egglog, like in the Python bindings, and you can get the extraction report instead of relying on the string output to know what's been extracted.

For the second, yeah maybe it's not necessary now that the performance is more reasonable for stringifying. On the other hand, maybe it does make sense to run our all our Python created benchmarks under this mode, if we want to closer track that performance...

src/main.rs
saulshanabrook: I renamed it to `--no-messages` to make it a bit shorter but also to keep it explicit about what part of the output it's disabling.

src/actions.rs
saulshanabrook: I updated the PR to run all tests with messages enabled, since now it's not too slow, but also switched the benchmarks to run the Python files with them disabled, to better track that performance and the slowdowns relevant to the upstream Python code.

src/main.rs
Alex-Fischman: Looks good, thanks!

PR #494, opened 2024-12-12
Small: clean up option (`:`) parsing

Alex-Fischman:
This PR cleans up the parsing of options such as `:ruleset` in `ast::parse::command`. This ensures that all the commands parse options in the same way.

Note: there are still some places where options are parsed manually; these are due to variadic constructs, such as `run`, where if we don't parse the second-to-last child as an option then we need to parse all the children as sorts.


yihozhang (approved):
lgtm

PR #495, opened 2024-12-13
Refactor sort extraction API

RiscInside:
[Zulip thread](https://egraphs.zulipchat.com/#narrow/channel/375765-egg.2Fegglog/topic/bottleneck.20in.20run_program.20function)

This fixes a part of the issue in #492 not concerning stringification. Original test still doesn't run on my computer due to a crash in `TermDag::to_string` (I suspect I don't have enough memory?)

The main contribution here is making extraction for `Sorts` return a `Term` instead of `Expr`. This cuts down the extraction time significantly. I was able to verify this by replacing extraction command with `(extract 1)` in the original test case (this is a meaningful test case since extract command still computes optimal extraction for *all* e-classes).

Ideally this would also get rid of `is_eq_sort` checks by moving that logic to the `extract_expr` for `EqSort`. This is currently impossible since `EGraph::find` accepts `ArcSort`, but I can change it to be `Sort` to make this simplification in this PR as well.

### DIFF
--- a/src/serialize.rs
+++ b/src/serialize.rs
@@ -1,7 +1,10 @@
 use ordered_float::NotNan;
 use std::collections::VecDeque;
 
-use crate::{util::HashMap, ArcSort, EGraph, Function, Symbol, TupleOutput, Value};
+use crate::{
+    extract::Extractor, util::HashMap, ArcSort, EGraph, Function, Symbol, TermDag, TupleOutput,
+    Value,
+};
 
 pub struct SerializeConfig {
     // Maximumum number of functions to include in the serialized graph, any after this will be discarded
@@ -312,7 +315,13 @@ impl EGraph {
                 let op = if sort.is_container_sort() {
                     sort.serialized_name(value).to_string()
                 } else {
-                    sort.make_expr(self, *value).1.to_string()
+                    let mut termdag = TermDag::default();
+                    let extractor = Extractor::new(self, &mut termdag);
+                    let (_, term) = sort
+                            .extract_term(self, *value, &extractor, &mut termdag)
+                            .expect("Extraction should be successful since extractor has been fully initialized");
+
+                    termdag.term_to_expr(&term).to_string()
                 };
                 egraph.nodes.insert(
                     node_id.clone(),
--- a/src/sort/fn.rs
+++ b/src/sort/fn.rs
@@ -185,34 +185,24 @@ impl Sort for FunctionSort {
         });
     }
 
-    fn make_expr(&self, egraph: &EGraph, value: Value) -> (Cost, Expr) {
-        let mut termdag = TermDag::default();
-        let extractor = Extractor::new(egraph, &mut termdag);
-        self.extract_expr(egraph, value, &extractor, &mut termdag)
-            .expect("Extraction should be successful since extractor has been fully initialized")
-    }
-
-    fn extract_expr(
+    fn extract_term(
         &self,
         _egraph: &EGraph,
         value: Value,
         extractor: &Extractor,
         termdag: &mut TermDag,
-    ) -> Option<(Cost, Expr)> {
+    ) -> Option<(Cost, Term)> {
         let ValueFunction(name, inputs) = ValueFunction::load(self, &value);
         let (cost, args) = inputs.into_iter().try_fold(
-            (
-                1usize,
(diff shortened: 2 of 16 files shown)


RiscInside:
For reference, here is the scale of improvement (`egglog` is what I installed semi-recently before my changes)
```
grass:[iurii]:~/Dev/egglog$ tail -n 1 tensor_example.egg 
(extract 1)
grass:[iurii]:~/Dev/egglog$ time target/release/egglog tensor_example.egg 2> /dev/null
1

real    0m0.066s
user    0m0.051s
sys     0m0.016s
grass:[iurii]:~/Dev/egglog$ time egglog tensor_example.egg 2> /dev/null
1

real    0m51.701s
user    0m50.087s
sys     0m1.557s
```

saulshanabrook:
I was able to test these changes against the example added in #492. They improve things a ton! From `77.60 secs` to `24.69 secs`

src/sort/mod.rs
```diff
@@ -97,22 +97,25 @@ pub trait Sort: Any + Send + Sync + Debug {
...
         value: Value,
         _extractor: &Extractor,
         _termdag: &mut TermDag,
-    ) -> Option<(Cost, Expr)> {
-        Some(self.make_expr(egraph, value))
+    ) -> Option<(Cost, Term)>;
+
+    /// Extracting an expression (with smallest cost) out of a primitive value.
+    /// The default implementation uses [`Sort::extract_term`].
+    fn make_expr(&self, egraph: &EGraph, value: Value) -> (Cost, Expr) {
```
saulshanabrook, line 111: Could we just remove this function? I don't believe it's used anymore.

src/sort/fn.rs
```diff
@@ -185,34 +185,24 @@ impl Sort for FunctionSort {
...
             |(cost, mut args), (sort, value)| {
                 let (new_cost, term) = extractor.find_best(value, termdag, &sort)?;
-                args.push(termdag.term_to_expr(&term));
+                args.push(term);
                 Some((cost.saturating_add(new_cost), args))
             },
         )?;
 
-        Some((cost, Expr::call_no_span("unstable-fn", args)))
+        Some((cost, termdag.app(name, args)))
```
saulshanabrook, line 205: Shouldn't this be `"unstable-fn"` instead of `name`? `name` is the first add above in the fold?

saulshanabrook (commented):
This looks great overall! Thank you for working on it.

Left a few small comments.

src/sort/mod.rs
RiscInside: It's used here https://github.com/egraphs-good/egglog/blob/main/src/serialize.rs#L315
I can probably inline the implementation however

src/serialize.rs
```diff
@@ -312,7 +315,13 @@ impl EGraph {
...
                     sort.serialized_name(value).to_string()
                 } else {
-                    sort.make_expr(self, *value).1.to_string()
+                    let mut termdag = TermDag::default();
+                    let extractor = Extractor::new(self, &mut termdag);
+                    let (_, term) = sort
+                            .extract_term(self, *value, &extractor, &mut termdag)
+                            .expect("Extraction should be successful since extractor has been fully initialized");
+
+                    termdag.term_to_expr(&term).to_string()
```
saulshanabrook, line 324: Thanks! I think this makes sense inlined here for now. It sets us up in case we want to refactor the extraction to batch extractions or something in the future if we want.

saulshanabrook (approved):
Looks good! Thank you!

I'll leave this open for a little to see if anyone else has thoughts.

src/serialize.rs
yihozhang: Could you elaborate a little more about the change here? the make_expr is changed to extract_term, which returns an additional Optiion wrapper. The only case it returns None is when the extractor is not initialized. Is my understanding correct? When would extract_term return None?

RiscInside: > Could you elaborate a little more about the change here? the make_expr is changed to extract_term, which returns an additional Optiion wrapper. The only case it returns None is when the extractor is not initialized. Is my understanding correct? When would extract_term return None?

`extract_term` would return None if `find_best` returns `None`, i.e. when one of the operands lowering hasn't been computed yet

PR #496, opened 2024-12-13
Avoid quadratic memory blowup in TermDag::to_string

RiscInside:
[Zulip](https://egraphs.zulipchat.com/#narrow/channel/375765-egg.2Fegglog/topic/bottleneck.20in.20run_program.20function)

Current version of `TermDag::to_string` caches a string for each DAG node. This results in quadratic memory blowup (these strings contain each other) which prevents vector benchmark from running on my machine.

This PR fixes the issue by building one result string and using string ranges to refer to already built substrings. Memory consumption should now be proportional to the length of the string. This also avoids quite a few allocations.

With these changes I am finally able to run the benchmark (it used to OOM for me). In conjunction with the other PR, full running time is now under a second.

### DIFF
--- a/src/termdag.rs
+++ b/src/termdag.rs
@@ -1,8 +1,9 @@
 use crate::{
     ast::Literal,
-    util::{HashMap, HashSet, IndexSet},
+    util::{HashMap, IndexSet},
     Expr, GenericExpr, Symbol,
 };
+use std::io::Write;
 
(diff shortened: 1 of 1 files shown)


RiscInside:
For reference, generated string is 889MB! We most definitely don't actually want to compute it!

yihozhang (approved):
Thanks!

PR #497, opened 2024-12-16
Make parser extensible

Alex-Fischman:
This PR adds a `Macros` struct that can be used to extend the parser from external crates. This introduces a name conflict with the existing public `Sexp` dependency. Since I don't believe that we need this dependency, I have removed it, which breaks all of the code that depends on the `ToSexp` trait, which is basically just resugaring. Since I do not see the value in resugaring I have removed it. However, I am open to having my mind changed on resugaring, in which case I will reimplement `Display` for the affected structs without using `ToSexp`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -376,7 +376,6 @@ dependencies = [
  "codspeed-criterion-compat",
  "egraph-serialize",
  "env_logger",
- "generic_symbolic_expressions",
  "getrandom",
  "glob",
  "hashbrown",
@@ -466,12 +465,6 @@ dependencies = [
  "typenum",
  "version_check",
 ]
-
-[[package]]
-name = "generic_symbolic_expressions"
-version = "5.0.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "597eb584fb7cfd1935294fc3608a453fc35a58dfa9da4299c8fd3bc75a4c0b4b"
 
 [[package]]
 name = "getrandom"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,8 +53,6 @@ lazy_static = "1.4"
 num = "0.4.3"
 smallvec = "1.11"
 
-generic_symbolic_expressions = "5.0.4"
-
 egraph-serialize = { version = "0.2.0", default-features = false }
 
 # binary dependencies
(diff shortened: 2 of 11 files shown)


oflatt:
From slack
The resugaring tests are still valuable! In fact they just caught a bug with haobin's branch.
They test that after our desugaring pass, you still get valid egglog code that runs correctly

Alex-Fischman:
@oflatt I don't understand how they test anything other than resugaring though. The desugaring pass is AST to AST, so it's guaranteed to typecheck anyway, and the resugaring tests don't check for e-graph equality or anything similar. Was the bug that it caught more complicated than forgetting to update the resugaring after introducing `constructor` and `relation`?

@RiscInside Yes. Would it still work for your use case if the `impl Display for Command` printed a span instead of directly resugaring? Would it work for you if it printed a span + source code?

I'm going to try and make Display use spans instead of resugaring today and see how that feels.

RiscInside:
> @RiscInside Yes. Would it still work for your use case if the `impl Display for Command` printed a span instead of directly resugaring? Would it work for you if it printed a span + source code?

I am relying on it being code that egglog binary or playground can run directly. I am guessing both of these options will break this assumption.

(To be clear, I am not saying that I need `Display` implementation specifically. I just want to have some way to go from `egglog::ast::Command` to a string in a way that produces identical code modulo span positions)

saulshanabrook:
Fwiw same as @RiscInside for the python bindings... We save the egglog ASTs we create and then print those out as strings if need be for debugging or to copy and paste it for a non python use case. 

I don't think we need or use desugaring in the Python bindings though, just the ability to turn AST nodes back into strings that will parse the same, modulo spans (which we don't use).

oflatt:
> I don't understand how they test anything other than resugaring though. The desugaring pass is AST to AST, so it's guaranteed to typecheck anyway, and the resugaring tests don't check for e-graph equality or anything similar. Was the bug that it caught more complicated than forgetting to update the resugaring after introducing constructor and relation?

I think the "resugaring" tests are more a test that our string output and parsing are work correctly round-trip. For example, if you get string conversion wrong for our AST the tests hopefully catch it. I think haobin's bug was related to this bug not sure

oflatt (approved):
Looks good! Thanks for extra effort to preserve our testing

PR #498, opened 2024-12-17
Replace `DUMMY_SPAN` with `Span::Panic` and `Span::Rust`

Alex-Fischman:
This PR improves error messages by introducing two new types of spans. The first is `Span::Rust`, which stores a location in a Rust file using [the `file!` macro family](https://doc.rust-lang.org/std/macro.file.html). The second is `Span::Panic`, which can be used when you know that the span should not be introspected; for an example, see the changes in `serialize.rs`.

### DIFF
--- a/src/ast/parse.rs
+++ b/src/ast/parse.rs
@@ -52,27 +52,33 @@ pub enum Span {
     Panic,
     /// A span from a `.egg` file.
     /// Constructed by `parse_program` and `parse_expr`.
-    Egglog {
-        file: Arc<SrcFile>,
-        i: usize,
-        j: usize,
-    },
+    Egglog(Arc<EgglogSpan>),
     /// A span from a `.rs` file. Constructed by the `span!` macro.
-    Rust {
-        file: &'static str,
-        line: u32,
-        column: u32,
-    },
+    Rust(Arc<RustSpan>),
+}
+
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub struct EgglogSpan {
+    file: Arc<SrcFile>,
+    i: usize,
+    j: usize,
+}
+
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub struct RustSpan {
+    pub file: &'static str,
+    pub line: u32,
+    pub column: u32,
 }
 
 #[macro_export]
 macro_rules! span {
     () => {
-        $crate::ast::Span::Rust {
+        $crate::ast::Span::Rust(std::sync::Arc::new($crate::ast::RustSpan {
             file: file!(),
             line: line!(),
             column: column!(),
-        }
+        }))
     };
 }
 
@@ -80,8 +86,8 @@ impl Span {
     pub fn string(&self) -> &str {
         match self {
             Span::Panic => panic!("Span::Panic in Span::String"),
-            Span::Rust { .. } => todo!(),
-            Span::Egglog { file, i, j } => &file.contents[*i..*j],
+            Span::Rust(_) => todo!(),
+            Span::Egglog(span) => &span.file.contents[span.i..span.j],
         }
     }
 }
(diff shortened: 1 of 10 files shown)


saulshanabrook:
For Python code I have been using a dummy span for all nodes. The thought would be eventually I could add tracebacks to the Python source lines but I haven't done that yet. 

For now, I guess I could just make up a dummy rust span that doesn't exist? Like it won't crash if the line/file  aren't real right?

Alex-Fischman:
Oh, I wasn't thinking about Python. Yeah, `Span::Rust` shouldn't try to read the file or anything. Can you think of a better name for it?

Alex-Fischman:
I think the performance regression is coming from spans getting larger from 3 words to 4, since the compiler can't distinguish between `Egglog` and `Rust` without an extra word for the tag. This would explain the slowdown because `to_core_actions` does a lot of `span.clone()`.

The only potential fix I can think to try would be to wrap spans in an `Rc`/`Arc`. If that doesn't make things faster then I'm not sure how to improve it.

Alex-Fischman:
Okay, that seems to have improved things. Opening for review.

saulshanabrook:
> Yeah, Span::Rust shouldn't try to read the file or anything. Can you think of a better name for it?

What about just File? Or SourceFile?

RiscInside:
If this gets merged, it would be cool to have `var!`, `call!` and `primitive!` macro replacements for deleted `*_no_span` functions.

Alex-Fischman:
@RiscInside Done, assuming that by `primitive!` you meant `lit!`?

RiscInside:
Great, thank you!

> assuming that by `primitive!` you meant `lit!`?

Indeed, my bad.

src/ast/parse.rs
```diff
@@ -46,20 +46,54 @@ pub fn parse_expr(
...
+        }))
+    };
 }
 
 impl Span {
     pub fn string(&self) -> &str {
-        &self.0.contents[self.1..self.2]
+        match self {
+            Span::Panic => panic!("Span::Panic in Span::String"),
+            Span::Rust(_) => todo!(),
```
saulshanabrook, line 89: This would just load the file from disk and load that line number? Is the question about the column?

FWIW I know Python just went through a refactor process so that they now store starting line, staring column, as well as ending line and ending column for bytecdoe ops and tracebacks: https://peps.python.org/pep-0657/

Do you think that might make sense in this case or is there a reason that just starting line and column is more appropriate?

src/serialize.rs
```diff
@@ -321,7 +317,7 @@ impl EGraph {
                             .extract_term(self, *value, &extractor, &mut termdag)
                             .expect("Extraction should be successful since extractor has been fully initialized");
 
-                    termdag.term_to_expr(&term).to_string()
+                    termdag.term_to_expr(&term, Span::Panic).to_string()
```
saulshanabrook, line 320: Why not use the `span!()` macro here? Just trying to understand when you want panic vs that one.

saulshanabrook (commented):
This seems good, better than what we have now, thank you @Alex-Fischman!

saulshanabrook (approved):

src/ast/parse.rs
Alex-Fischman: Yes, the problem is that Rust only exposes `file!`, `line!`, and `column!`.

Docs: https://doc.rust-lang.org/std/macro.file.html

src/serialize.rs
Alex-Fischman: The `to_string` method ignores spans, so using `Span::Panic` is just a runtime check that the spans are in fact ignored.

In general, I think `Span::Panic` is useful when you know that the spans are not observable by outside code; i.e. I would avoid it if the function returned an `Expr`.

PR #499, opened 2024-12-22
Impls BigRat `pow` for negative powers

rtzam:
This also changes the behavior of 0^0 to be 1
as opposed to undefined

### DIFF
--- a/src/sort/bigrat.rs
+++ b/src/sort/bigrat.rs
@@ -49,22 +49,30 @@ impl Sort for BigRatSort {
 
         add_primitives!(eg, "pow" = |a: Q, b: Q| -> Option<Q> {
             if a.is_zero() {
-                if b.is_positive() {
+                if b.is_zero() {
+                    // 0^0 = 1 by common convention
+                    Some(Q::one())
+                } else if b.is_positive() {
+                    // 0^n = 0 where (n > 0)
                     Some(Q::zero())
                 } else {
-                    None
-                }
-            } else if b.is_zero() {
-                Some(Q::one())
-            } else if let Some(b) = b.to_i64() {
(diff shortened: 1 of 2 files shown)


rtzam:
Hi Eli! Been a while since we last spoke. When I went to use `pow` for `BigRat`, I noticed that these features were missing. Also, I changed the behavior of `0^0` to being 1 from being undefined which may be controversial 🤷. Would love any guidance on how to add proper test cases any other egglog-dev conventions as this is my first PR. Happy Holidays!

saulshanabrook:
Thanks for the PR! I think for tests the easiest way to add them for this kind of change is to add to an existing .egg test file or create a new one that runs some check commands to verify that the output of the primitives matches what you expect.

src/sort/bigrat.rs
```diff
@@ -48,23 +48,36 @@ impl Sort for BigRatSort {
...
-                    None
-                }
-            } else if b.is_zero() {
-                Some(Q::one())
-            } else if let Some(b) = b.to_i64() {
-                if let Ok(b) = usize::try_from(b) {
-                    num::traits::checked_pow(a, b)
-                } else {
-                    // TODO handle negative powers
+                    // 0^n => (1/0)^(abs n) where (n < 0)
```
saulshanabrook, line 65: Is this comment still relevent?

saulshanabrook (approved):
Looks good overall! Thank you for all the tests.

src/sort/bigrat.rs
rtzam: Which part? The change removes the `TODO` but I add in a comment for which mathematical case the branch is handling

PR #502, opened 2024-12-28
Fix the behaviour of i64 primitives on overflow

RiscInside:
i64 primitives handle overflow with wraparound in Release mode and panic in Debug mode. This PR makes wraparound behavior consistent across compilation modes.

(Side note: I didn't know shift operations can fail. Would it be a good idea to add a wrapping version for them?)

PS: took me some time to sort out git situation after unsuccessful rebase, sorry about that.

### DIFF
--- a/src/sort/i64.rs
+++ b/src/sort/i64.rs
@@ -19,9 +19,6 @@ impl Sort for I64Sort {
     }
 
     #[rustfmt::skip]
-    // We need the closure for division and mod operations, as they can panic.
-    // cf https://github.com/rust-lang/rust-clippy/issues/9422
-    #[allow(clippy::unnecessary_lazy_evaluations)]
     fn register_primitives(self: Arc<Self>, typeinfo: &mut TypeInfo) {
         typeinfo.add_primitive(TermOrderingMin {
            });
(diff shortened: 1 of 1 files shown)


saulshanabrook:
Thanks for this! 

I agree it makes sense to have consistent overflow behavior across builds. 

Personally, I would prefer checked_add so that the operation will fail on an overflow instead id succeeding and wrapping.  

At least in all my use cases, this would be preferred behavior, I wouldn't want it to silently wrap. 

What do you think?

RiscInside:
This seems like a good idea. It would also be in line with how shl and shr behave. I will change my PR to this.

Some kind of wrapping arithmetic would be useful too. Perhaps adding separate wrapping equivalents would a good idea?

saulshanabrook:
> Some kind of wrapping arithmetic would be nice to have though. Perhaps a different set of wrapping primitives (like "+%") is in order?


Yeah I think if you want a wrapping version for a certain use case, then adding a new op for it would make sense.

RiscInside:
I will add wrapping ones in a separate PR.

saulshanabrook (approved):
Thank you for this fix!

PR #503, opened 2024-12-28
Expose extraction module publicly

RiscInside:
It's currently impossible to define custom sorts outside of `egglog` due to the fact that types like `Extractor` and `Cost` aren't exposed. This PR fixes this by exposing `extractor` module.

The module is already using `pub` and `pub(crate)` in all the right places. I do wonder however if `costs` field of the `Extractor` needs to be public

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,7 +15,7 @@ mod actions;
 pub mod ast;
 pub mod constraint;
 mod core;
-mod extract;
+pub mod extract;
 mod function;
 mod gj;
 mod serialize;


RiscInside:
> Is there some way to make Rust to tell us about issues like this? I thought it would make a warning if something had lower visibility than required. Is the `Sort` trait not public enough?

`Sort` is public and can be referred to outside of the crate. However, one of the methods in `Sort` uses `Extractor` type that is not accessible outside of the crate. This type itself is public, but the module it is defined in isn't, so egglog consumers can't use the type and hence can't use this trait.

Rust compiler has a lint called `unreachable_pub` that complains if pub item isn't reachable from outside of the crate. It is allow by default due to false positives, but perhaps enabling it wouldn't be a big deal for egglog. I will have a look.

RiscInside:
Okay, the lint works, and it finds *a lot* of errors, but surprisingly it thinks `Extractor` is reachable somehow? I can't figure out how though - it doesn't even show up in `rustdoc` output!

Alex-Fischman (approved):
Weird, it does find `Cost` though when I run it.

Anyways, this change LGTM.

PR #506, opened 2025-01-05
Polish the macro interface

yihozhang:
This PR does a few small things:

* Make parsing-related functions public and make them member functions of Parser
* Make the macro type a trait- the current macro type is defined as a `fn`, which has to be static (e.g., no captured values from the environment nor dynamic behavior)
* Add a few convenience methods so it's usable from outside the crate

### DIFF
--- a/src/ast/parse.rs
+++ b/src/ast/parse.rs
@@ -3,47 +3,6 @@
 use crate::*;
 use ordered_float::OrderedFloat;
 
-pub type Macro<T> = fn(&[Sexp], Span, &Parser) -> Result<T, ParseError>;
-
-#[derive(Clone)]
-pub struct Parser {
-    pub commands: HashMap<Symbol, Macro<Vec<Command>>>,
-    pub actions: HashMap<Symbol, Macro<Vec<Action>>>,
-    pub exprs: HashMap<Symbol, Macro<Expr>>,
-    pub symbol_gen: SymbolGen,
-}
-
-impl Default for Parser {
-    fn default() -> Self {
-        Self {
-            commands: Default::default(),
-            actions: Default::default(),
-            exprs: Default::default(),
-            symbol_gen: SymbolGen::new("$".to_string()),
-        }
-    }
-}
-
-pub fn parse_program(
-    filename: Option<String>,
-    input: &str,
-    parser: &Parser,
-) -> Result<Vec<Command>, ParseError> {
-    let sexps = all_sexps(Context::new(filename, input))?;
-    let nested = map_fallible(&sexps, parser, commands)?;
-    Ok(nested.into_iter().flatten().collect())
-}
-
-// currently only used for testing, but no reason it couldn't be used elsewhere later
-pub fn parse_expr(
-    filename: Option<String>,
-    input: &str,
-    parser: &Parser,
-) -> Result<Expr, ParseError> {
-    let sexp = sexp(&mut Context::new(filename, input))?;
-    expr(&sexp, parser)
-}
-
 /// A [`Span`] contains the file name and a pair of offsets representing the start and the end.
 #[derive(Clone, PartialEq, Eq, Hash)]
 pub enum Span {
@@ -250,11 +209,11 @@ impl Sexp {
 fn map_fallible<T>(
     slice: &[Sexp],
     parser: &Parser,
-    func: impl Fn(&Sexp, &Parser) -> Result<T, ParseError>,
+    func: impl Fn(&Parser, &Sexp) -> Result<T, ParseError>,
 ) -> Result<Vec<T>, ParseError> {
     slice
         .iter()
-        .map(|sexp| func(sexp, parser))
+        .map(|sexp| func(parser, sexp))
         .collect::<Result<_, _>>()
 }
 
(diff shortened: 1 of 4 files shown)


src/ast/parse.rs
```diff
@@ -174,6 +133,12 @@ impl Display for Span {
 #[derive(Debug, Error)]
 pub struct ParseError(Span, String);
```
Alex-Fischman, line 134: Nit: the `ParseError` fields should just be `pub`, as this will be nicer than using `ParseError::new` when the message comes from `format!`.

src/ast/parse.rs
```diff
@@ -287,548 +252,659 @@ fn options(sexps: &[Sexp]) -> Result<Vec<(&str, &[Sexp])>, ParseError> {
...
+            .expect_list("input sorts")?
+            .iter()
+            .map(|sexp| sexp.expect_atom("input sort"))
+            .collect::<Result<_, _>>()?,
+        output: output.expect_atom("output sort")?,
+    })
+}
+
+pub trait Macro<T> {
+    fn get_head(&self) -> String;
```
Alex-Fischman, line 267: I don't think this should be a trait method. I don't see anything wrong with just passing it in to `add_*_macro`, and making it a trait method makes it seem like it could change over time (when it will only be called once).

src/ast/parse.rs
```diff
@@ -249,12 +214,12 @@ impl Sexp {
...
-    func: impl Fn(&Sexp, &Parser) -> Result<T, ParseError>,
+    parser: &mut Parser,
+    func: impl Fn(&mut Parser, &Sexp) -> Result<T, ParseError>,
 ) -> Result<Vec<T>, ParseError> {
     slice
         .iter()
-        .map(|sexp| func(sexp, parser))
+        .map(|sexp| func(parser, sexp))
         .collect::<Result<_, _>>()
 }
```
Alex-Fischman, line 225: The `options` helper should probably be `pub`, and maybe get renamed.

src/ast/parse.rs
```diff
@@ -287,548 +252,659 @@ fn options(sexps: &[Sexp]) -> Result<Vec<(&str, &[Sexp])>, ParseError> {
     Ok(out)
 }
 
-fn commands(sexp: &Sexp, parser: &Parser) -> Result<Vec<Command>, ParseError> {
-    let (head, tail, span) = sexp.expect_call("command")?;
+fn schema(input: &Sexp, output: &Sexp) -> Result<Schema, ParseError> {
```
Alex-Fischman, line 255: The `schema` helper can be `pub`

Alex-Fischman (changes requested):

src/ast/parse.rs
```diff
@@ -287,548 +252,659 @@ fn options(sexps: &[Sexp]) -> Result<Vec<(&str, &[Sexp])>, ParseError> {
...
+            .iter()
+            .map(|sexp| sexp.expect_atom("input sort"))
+            .collect::<Result<_, _>>()?,
+        output: output.expect_atom("output sort")?,
+    })
+}
+
+pub trait Macro<T> {
+    fn get_head(&self) -> String;
+    fn parse(&self, args: &[Sexp], span: Span, parser: &mut Parser) -> Result<T, ParseError>;
```
Alex-Fischman, line 268: (This comment got deleted from my first review for some reason)

I think that the `Macro` trait should be a "trait alias":
```rust
pub trait Macro<T>: Fn(&[Sexp], Span, &mut Parser) -> Result<T, ParseError> {}
impl<T, F: Fn(&[Sexp], Span, &mut Parser) -> Result<T, ParseError>> Macro<T> for F {}
```
This lets you get rid of `SimpleMacro`.

Alex-Fischman (changes requested):

src/ast/parse.rs
Alex-Fischman: Although I'm not sure that using a trait instead of a `fn` is actually useful in this case. Because it has to end up in an `Arc`, the lifetime has to be `'static`, so the only references it can make are static and immutable (`Fn` vs. `FnMut` is the same as using `&self` vs `&mut self` in the trait method). I don't see why this would be necessary for a macro; is there some ergonomic benefit that I'm missing?

src/ast/parse.rs
yihozhang: I think of `head` more as a property of Macro than an argument to `add_*_macro`, which may change for different `add_*_macro` calls. The sort/primitive interface has a similar `name` getter.

src/ast/parse.rs
yihozhang: Re your second comment, `fn` is for static functions (different from `Fn` which is the closure type), so it used to just store a function pointer and that's why it does not need `dyn` keyword (and `Arc`). Your SimpleMacro test works because the closure does not touch any local variable, so it is treated as a function. It would break if you lift "xxxx" out of the closure definition.

Also as in the SimpleMacro example, that macro is not 'static lifetime, and will be destructed when rc drops to 0.

yihozhang: Re your first comment, my worry about not making Macro a proper trait is that it requires the user to either write global functions (which are static) or write a local anonymous function (which does not scale), so in my opinion a proper trait is ergonomically better

Alex-Fischman: > Re your first comment, my worry about not making Macro a proper trait is that it requires the user to either write global functions (which are static) or write a local anonymous function (which does not scale), so in my opinion a proper trait is ergonomically better

Can you elaborate on what you mean by "do not scale"?

> Re your second comment, `fn` is for static functions (different from `Fn` which is the closure type), so it used to just store a function pointer and that's why it does not need `dyn` keyword (and `Arc`). Your SimpleMacro test works because the closure does not touch any local variable, so it is treated as a function. It would break if you lift "xxxx" out of the closure definition.

Yes, for some reason I thought that using `move` would let the compiler constant fold the data but it does not.

src/ast/parse.rs
Alex-Fischman: Can you rename `get_head` to `name` then, they do the same thing right? Also maybe it should return a `Symbol` instead?

src/ast/parse.rs
yihozhang: Yeah, an example of an advanced macro I want to write is to add attributes to rules (similar to https://github.com/egraphs-good/egglog/pull/431 ). For each rule, macro will read the attributes, keep records of them in the state of the `Macro`, and nuke the attributes before sending it to egglog's rule parsing. I think this would be hard to implement as an anonymous function.

src/ast/parse.rs
yihozhang: Done!

src/ast/parse.rs
Alex-Fischman: 1. I think this would require the `parse` method to take `&mut self`.
2. How would the rest of the system be able to read the macros? This seems like something that would have to be/should be implemented in core egglog to me.

yihozhang: Re 1: I tried to make parse take `&mut self` but since the same macro may be called nestedly, (e.g., when parsing `(macro-1 (macro-1))`, the borrow checker does not allow this. My current thinking is to do interior mutability within the macro with e.g., `RefCell`.
Re 2: The parser needs to then pass the read attributes to the E-graph or other plugins after the parsing is done.

I think both are doable and I'll play around with this idea more later. Back to the point here, I think having Macro as a proper trait is slightly more advantageous here and there, at the cost of needing to write `SimpleMacro::new` (which I didn't use in https://github.com/egraphs-good/egglog-experimental/pull/2). The trait alias also does not work with other member methods like `name()` I suppose.

Alex-Fischman (approved):
LGTM

PR #507, opened 2025-01-06
Expose egglog REPL as a function

yihozhang:
This PR should subsume #486 by exposing the repl functionality of egglog as a public function. Compared to #486, this handles for multi-line commands and does not introduce duplicate code.

Other minor changes:

* Fix a bug with the parser where input consisting of newline characters cause the parser to fail (see [this program](https://egraphs-good.github.io/egglog/?program=XQAAgAADAAAAAAAAAAAFaD3_____4AAAAA%253D%253D) ).
* Move implementation for `interactive_mode` to `run_commands` for correct handling, since the previous logic for `interactive_mode` assumes one command per line and just prints `(done)` for each line of the input.

### DIFF
--- a/src/repl.rs
+++ b/src/repl.rs
@@ -7,6 +7,7 @@ impl EGraph {
     pub fn repl(&mut self) -> io::Result<()> {
         self.repl_with(io::stdin(), io::stdout())
     }
+
     pub fn repl_with<R, W>(&mut self, input: R, mut output: W) -> io::Result<()>
     where
         R: Read,
@@ -27,8 +28,6 @@ impl EGraph {
 
         if !cmd_buffer.is_empty() {
             run_command_in_scripting(self, &cmd_buffer, &mut output)?;
-
-            log::logger().flush();
         }
 
         Ok(())
--- a/src/repl.rs
+++ b/src/repl.rs
@@ -48,9 +48,7 @@ where
                 writeln!(output, "{msg}")?;
             }
         }
-        Err(err) => {
-            log::error!("{err}");
-        }
+        Err(err) => log::error!("{err}"),
     }
     Ok(())
 }
--- a/src/repl.rs
+++ b/src/repl.rs
@@ -1,0 +1,152 @@
+use std::io::{self, Read};
+use std::io::{BufRead, BufReader, Write};
+
+use crate::EGraph;
+
+impl EGraph {
+    pub fn repl(&mut self) -> io::Result<()> {
+        self.repl_with(io::stdin(), io::stdout())
+    }
+    pub fn repl_with<R, W>(&mut self, input: R, mut output: W) -> io::Result<()>
+    where
+        R: Read,
+        W: Write,
+    {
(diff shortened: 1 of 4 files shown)


yihozhang:
cc @gussmith23

src/repl.rs
```diff
@@ -0,0 +1,150 @@
...
+            if should_eval(&cmd_buffer) {
+                run_command_in_scripting(self, &cmd_buffer, &mut output)?;
+                cmd_buffer = String::new();
+            }
+        }
+
+        if !cmd_buffer.is_empty() {
+            run_command_in_scripting(self, &cmd_buffer, &mut output)?;
+
+            log::logger().flush();
```
Alex-Fischman, line 31: Are there other points where we need to flush the buffer? It seems wrong to only do it here...

src/repl.rs
```diff
@@ -0,0 +1,150 @@
...
+                }
+            }
+            ';' => {
+                // `any` moves the iterator forward until it finds a match
+                if !indices.any(|ch| ch == '\n') {
+                    return false;
+                }
+            }
+            '"' => {
+                if !indices.any(|ch| ch == '"') {
```
Alex-Fischman, line 62: I think this is wrong (it was wrong before) due to string escapes. Can we reuse the tokenizer here?

src/repl.rs
```diff
@@ -0,0 +1,150 @@
+use std::io::{self, Read};
+use std::io::{BufRead, BufReader, Write};
+
+use crate::EGraph;
+
+impl EGraph {
+    pub fn repl(&mut self) -> io::Result<()> {
+        self.repl_with(io::stdin(), io::stdout())
+    }
+    pub fn repl_with<R, W>(&mut self, input: R, mut output: W) -> io::Result<()>
```
Alex-Fischman, line 10: nit: missing newline between functions

src/repl.rs
```diff
@@ -0,0 +1,150 @@
...
+{
+    match egraph.parse_and_run_program(None, command) {
+        Ok(msgs) => {
+            for msg in msgs {
+                writeln!(output, "{msg}")?;
+            }
+        }
+        Err(err) => {
+            log::error!("{err}");
+        }
```
Alex-Fischman, line 85: nit: the Err case can be 1 line

Alex-Fischman (changes requested):
Some nits, the important note is about `should_eval`

src/repl.rs
yihozhang: Interesting. It seems `log::logger().flush();` is called at the end of `EGraph::run_program`, so I just removed this line

src/repl.rs
yihozhang: good point. I suppose this also does not handle square brackets. Will address

yihozhang:
`should_eval` now uses the Sexp tokenizer. The code is much more concise. This is a great example of "parse, don't validate"!

Alex-Fischman (approved):
LGTM

PR #509, opened 2025-01-06
Make the type constraint system user-extensible

yihozhang:
This PR allows users to BYO type constraints by making constraints trait-oriented rather than ADT-oriented. With this PR, we can implement #452 in `egglog-experimental` instead of in core egglog.

assigning @saulshanabrook and @oflatt for review


oflatt (approved):
Looks good to me! Very cool

saulshanabrook (approved):
I agree looks cool! I don't have a concrete sense of how to move forward with the function stuff after this but happy to trust you if you do.

PR #510, opened 2025-01-07
Expose CLI for `egglog-experimental`

Alex-Fischman:
This PR exposes the command-line interface so that the `egglog-experimental` crate can reuse it, as well as significantly improving the existing source code.


Alex-Fischman:
I would review this PR commit-by-commit, they are all pretty small and well separated, and the last one is code motion (necessary to expose the `cli` function to `egglog-experimental`).

src/main.rs
```diff
@@ -40,8 +40,12 @@ struct Args {
     no_messages: bool,
 }
 
-#[allow(clippy::disallowed_macros)]
 fn main() {
+    cli(EGraph::default())
+}
+
+#[allow(clippy::disallowed_macros)]
+pub fn cli(mut egraph: EGraph) {
```
yihozhang: This should not be in `main.rs?

src/cli.rs
```diff
@@ -0,0 +1,252 @@
...
+    #[clap(long, default_value = "0")]
+    serialize_n_inline_leaves: usize,
+    /// Prevents egglog from printing messages
+    #[clap(long)]
+    no_messages: bool,
+}
+
+#[cfg(not(target_arch = "wasm32"))]
+#[allow(clippy::disallowed_macros)]
+pub fn cli(mut egraph: EGraph) {
```
yihozhang, line 56: We may want a wasm repl? I wonder if we can have a struct that contains the build information and make it an argument to `cli` to avoid the dependence on `env!`. This also has other benefits: For example, egglog-experimental may want to display `Welcome to Egglog (experimental) REPL!` instead of the default message.

yihozhang (changes requested):
I have a small question and otherwise it looks good

src/cli.rs
Alex-Fischman: `clap` is the main problem, not the welcome message. See the failing build log: https://github.com/egraphs-good/egglog/actions/runs/12657497954/job/35272487220. However, note that the `EGraph::repl` function still works on WASM, so we can add it in later. The issue on WASM is that there's no command line.

Alex-Fischman:
After talking to people offline I believe that the `bin` feature is adding unnecessary complexity, so I've removed it (and cleaned up `Cargo.toml` in the process). I will add it back if someone can give a reason to keep it, but no one that I've talked to on the core team has one.

src/cli.rs
Alex-Fischman: We resolved this offline, see my comment below.

yihozhang:
@DaniPopes we are thinking about simplifying our build dependencies and removing the feature flags that used to keep the build thin. Please take a look at this PR. Since you made #467, do you have any usage for having these feature flags?

DaniPopes:
Yes, it's for excluding clap, graphviz, chrono... dependencies that are unused when importing egglog as a library. This brings the total transitive dependencies imported from 133 down to 56.

Alex-Fischman:
@DaniPopes Do you have a use case where having fewer dependencies matters?

For context, the egglog team's current goal is to stabilize this crate by June, and it doesn't seem like having Cargo features is worth the smaller build to me. _That said_, if someone depends on this, we can support it as a library feature.

Alex-Fischman:
I think the ping failed (I edited it in), my bad. @DaniPopes

DaniPopes:
Well it's not really a use case, but I would say it's pretty significant the fact that it's 3 times faster to build basically for free?

Having feature gates for the library, required to build the binary is a pretty common pattern for this reason; the features are enabled by default to allow "cargo install" to work without extra flags, and library users can opt-in to lower dependencies with default-features=false

Alex-Fischman:
I've added back the features. I've cut the `serde_json` dependency as it is not used in the main crate.

yihozhang (approved):
Thanks!!

Issue #512, opened 2025-01-17
[Potential Bug] set.rs: Inconsistent cost calculation between Sets and MultiSets


julianjumper:
## Description
```
(sort imultiset (MultiSet i64))
(sort iset (Set i64))

(query-extract (set-of 1 2 3))        ; Output: (set-insert (set-insert (set-insert (set-empty) 3) 2) 1)
(query-extract (multiset-of 1 2 3))   ; Output: (multiset-of 1 2 3)
```
There is an inconsistent cost calculation between Set and MultiSet operations, making `(set-insert (set-insert (set-insert (set-empty) 3) 2) 1)` less expensive than `(set-of 1 2 3)` but only for sets, not for MultiSets. <br>
In my understanding, the shorter `set-of` expression should have lower cost than multiple `set-insert` operations.

#### Code Location
- Sets: [src/sort/set.rs](https://github.com/egraphs-good/egglog/tree/main/src/sort/set.rs) (line 174)
- Multisets: [src/sort/multiset.rs](https://github.com/egraphs-good/egglog/tree/main/src/sort/multiset.rs) (line 256)

If this it not intended behaviour, I'd be happy to submit a PR!

saulshanabrook:
Thanks for this issue!

I added the multiset but the set was already there. Does anyone know why we wouldn't use the more concise extracted form for containers like this? Like set-of instead of the cons list version? 

That would be my preference for brevity and to march more closely with the backing data structure.

yihozhang:
> Does anyone know why we wouldn't use the more concise extracted form for containers like this?

I think this is accidental. Probably because we didn't have the type system that allows us to express `set-of` at the time. Changing to `set-of` makes sense.

julianjumper:
I submitted a PR (#514), this works for my needs, not sure if this takes everything into account

PR #514, opened 2025-01-19
[#512] set-of instead of set-insert

julianjumper:
This PR addresses my issue #512, which was about the inconsistency between Sets and MultiSets. <br>

I changed the `extract_term` logic of Sets to the same of MultiSets. This works for me, but I am not sure if this takes everything into account.


saulshanabrook (approved):
Looks good, thank you!

Issue #516, opened 2025-01-27
Can EGraph be Sync + Send?


cksac:
Currently EGraph not Sync+Send due to `Arc<dyn Macro<T>>` not Sync+Send. Can we make `Parser` Sync+Send?

```rust
#[derive(Clone)]
pub struct Parser {
    commands: HashMap<Symbol, Arc<dyn Macro<Vec<Command>>>>,
    actions: HashMap<Symbol, Arc<dyn Macro<Vec<Action>>>>,
    exprs: HashMap<Symbol, Arc<dyn Macro<Expr>>>,
    pub symbol_gen: SymbolGen,
}
```

Alex-Fischman:
I'm not super familiar with Rust concurrency, but the following code gives me many error messages that are not related to `Parser`. Why do you believe that the `Parser` is the problem?

```rust
const _: () = {
    fn assert_send<T: Send>() {}
    fn assert_sync<T: Sync>() {}
    fn assert_all() {
        assert_send::<EGraph>();
        assert_sync::<EGraph>();
    }
};
```

Alex-Fischman:
Also, what is your use case? We can (maybe?) make it `Send` and `Sync` if we have a good reason to do so... You might be better served by [serialization](https://docs.rs/egglog/latest/egglog/struct.EGraph.html#method.serialize) depending on your application though.

cksac:
I would like to use EGraph with lazy_static
```rust
lazy_static::lazy_static! {
    pub static ref RT:  std::sync::Mutex<EGraph> = std::sync::Mutex::new(EGraph::default());
}
```
you are right, there are numbers of errors, not only the Parser.
```
314 | / pub struct Rc<
315 | |     T: ?Sized,
316 | |     #[unstable(feature = "allocator_api", issue = "32838")] A: Allocator = Global,
317 | | > {
    | |_- doesn't satisfy `Rc<egglog::actions::Program>: Send` or `Rc<egglog::function::index::ColumnIndex>: Send`
    |
    = note: the following trait bounds were not satisfied:
            `(dyn Macro<Vec<GenericCommand<Symbol, Symbol>>> + 'static): Sync`
            which is required by `Mutex<EGraph>: Sync`
            `(dyn Macro<Vec<GenericCommand<Symbol, Symbol>>> + 'static): Send`
            which is required by `Mutex<EGraph>: Sync`
            `(dyn Macro<Vec<GenericAction<Symbol, Symbol>>> + 'static): Sync`
[...]

PR #517, opened 2025-01-31
Thread-safe EGraph struct

yihozhang:
We can make the EGraph implements the `Send` trait by substituting some of the components with thread-safe alternatives so this works:

```
    lazy_static! {
        pub static ref RT: Mutex<EGraph> = Mutex::new(EGraph::default());
    }
```

### DIFF
--- a/src/ast/parse.rs
+++ b/src/ast/parse.rs
@@ -217,16 +217,19 @@ fn map_fallible<T>(
         .collect::<Result<_, _>>()
 }
 
-pub trait Macro<T> {
+pub trait Macro<T>: Send {
     fn name(&self) -> Symbol;
     fn parse(&self, args: &[Sexp], span: Span, parser: &mut Parser) -> Result<T, ParseError>;
 }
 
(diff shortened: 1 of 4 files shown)


src/ast/parse.rs
```diff
@@ -217,16 +217,19 @@ fn map_fallible<T>(
         .collect::<Result<_, _>>()
 }
 
-pub trait Macro<T> {
```
Alex-Fischman, line 220: If you put `Send + Sync` here instead of just `Send` you don't need to repeat the bounds everywhere. I only needed them on `SimpleMacro` IIRC.

Alex-Fischman (approved):

Alex-Fischman:
Fixes #516.

saulshanabrook (approved):

src/ast/parse.rs
yihozhang: Good catch! Yeah I just implemented that and code is simpler.

PR #518, opened 2025-02-04
Add option for a prompt on the CLI

gussmith23:
I'm working on a project that interacts with egglog's CLI programmatically. Being able to detect when a command is done is useful. A prompt is a very easy way to detect when a command completes.

This trips the nits detector because of the use of `print!` -- can I get around that somehow?

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,5 +1,5 @@
 use crate::*;
-use std::io::{self, BufRead, BufReader, Read, Write};
+use std::io::{self, stdout, BufRead, BufReader, Read, Write};
 
 #[cfg(feature = "bin")]
 pub mod bin {
@@ -50,6 +50,9 @@ pub mod bin {
         /// Prevents egglog from printing messages
         #[clap(long)]
         no_messages: bool,
+        /// String to use as the prompt.
+        #[clap(long, default_value = "")]
+        prompt: String,
     }
 
     #[allow(clippy::disallowed_macros)]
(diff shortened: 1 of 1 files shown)


yihozhang:
The interactive mode will print a `(done)` once every command is done (sorry for the poor documentation). Does that work for you?

gussmith23:
> The interactive mode will print a `(done)` once every command is done (sorry for the poor documentation). Does that work for you?

Unfortunately no -- here's why. My interactive routine needs to be able to run multiple commands at once, and only have a single prompt/marker appear at the very end of all of the commands. Currently, I do this by writing a batch of commands to a file and then using `(include <filepath>)`. Because `include` is just a single command at the top level, all of the commands in the file will run and a _single_ prompt will be printed when they're all done. Based on my testing, interactive mode prints a `(done)` after each command, even if the commands were `include`d. This is an issue, because if I don't know how many commands were in the `include`d file, I don't know how many `(done)`s to look for. With the prompt, however, I always know to look for a single prompt at the very end.

Alex-Fischman:
I think your issue can be solved by changing your uses of
```
(include "file.egg")
```
to
```
(set-option interactive_mode 0)
(include "file.egg")
(set-option interactive_mode 1)
```
This snippet should only print one (done) , from the second set-option.

I looked at doing this automatically, but it's not trivial because of the case where interactive_mode is already 0. I think this is perfectly fine for your use case though.

PR #526, opened 2025-03-16
Fix primitives in queries

Alex-Fischman:
This PR doesn't actually work; it causes fewer tests to pass than before. I've looked at the `bignum` test, and it seems to fail by not finding the variable that is returned from a primitive, but I'm not sure how to fix it.


Alex-Fischman:
Fixed after a call with Eli!

PR #527, opened 2025-03-17
eggcc benchmarks

oflatt:


Alex-Fischman:
What does CodSpeed mean by "dropped" benchmarks?

oflatt:
I think that's due to me renaming `example_benchmarks.rs` to `ci_benchmarking.rs`
I'm not sure why only 10 are dropped and the rest are ignored...

oflatt:
Codspeed can't figure out that they are the same benchmarks after the renaming

yihozhang:
That's a lot of code! Maybe we should have a separate repo for `egglog-benchmarks`?

saulshanabrook (commented):
Would you mind renaming the `ci_benchmarking.rs` file to `example_benchmarks.rs` so that they aren't counted as deleted then re-added? I had to manually go in and "ignore" all the smaller benchmarks which have non deterministic timing. If we rename them all then I would have to go and redo that, plus we would lose all our history data for them.

For some reason this benchmark comparison isn't picking up any of the new benchmarks too and the comment is showing a comparison against an earlier one of your commits instead of one in main, maybe from a force push? I can ping on their discord to see why this is.

FYI we should make sure to squash this when we merge it so that the big files aren't included in the git history.

oflatt:
I feel that `example_benchmarks.rs` is a bad name for the file
Perhaps we should ask the codspeed people if there's a rename option?

Alex-Fischman:
I've reviewed and merged #572.

saulshanabrook:
Crosspost from discord response:

> Hey, unfortunately CodSpeed does not support renaming benchmarks, renaming  the file that defines them or moving them to other files at the moment.
> 
> We do have a feature on our roadmap to be able to sync benchmarks when we detect such a thing happen, so we will keep you posted 😉

Alex-Fischman:
You appear to have lost the `subsume-relation` benchmark (or maybe it didn't exist when the branch was created?).

Alex-Fischman:
I've tried one thing and I'm all out of ideas

FTRobbin (approved):
LGTM based on my limited knowledge of how this works.

PR #530, opened 2025-03-21
Add `find_variants` to `Extractor`

bksaiki:
The `EGraph` type implements `extract` and `extract_variants`, but the extractor only implements `find_best`. Adds `find_variants` to `Extractor` which is basically a copy of the original `extract_variants` function.

### DIFF
--- a/src/extract.rs
+++ b/src/extract.rs
@@ -77,33 +77,9 @@ impl EGraph {
         limit: usize,
         termdag: &mut TermDag,
     ) -> Vec<Term> {
-        let output_sort = sort.name();
-        let output_value = self.find(sort, value);
-        let ext = &Extractor::new(self, termdag);
-        ext.ctors
-            .iter()
-            .flat_map(|&sym| {
-                let func = &self.functions[&sym];
-                if !func.schema.output.is_eq_sort() {
-                    return vec![];
-                }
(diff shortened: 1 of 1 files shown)


bksaiki:
As discussed with @oflatt and @Alex-Fischman. Unsure on the correct return type for `Extractor::find_variants` since `find_best` is an `Option`.

Alex-Fischman:
I think that it's fine to not return a `Result` as long as you document that the `*_variants` functions might return fewer than `limit` values (if they don't exist in the e-graph).

FTRobbin (approved):
Looks good to me since it is mostly moving code from egraph to extractor.

But I don't understand the error message `extract_variants should be called after extractor initialization`. It seems to be an impossible reason unless there is some extractor initialization happening somewhere else.

Alex-Fischman:
@bksaiki Can you confirm that you don't need this PR merged for the next month-ish? You're just upstreaming a change that herbie-egglog would like?

I would rather wait to merge this until the new backend is in; it seems just as easy to do this in the new backend as it is in this PR, and it will avoid merge conflicts.

bksaiki:
Not high priority

Alex-Fischman:
Closing after confirming in-person.

PR #531, opened 2025-04-02
New backend containers

Alex-Fischman:
This PR:
- changes the `add_primitive` macro to a procedural macro
- adds support for storing new backend containers and specialized type constraints
- implements Vec, Map, Set, and Multiset (except for map) containers
- cleans up `unstable-fn` to show where the missing pieces need to go


src/sort/multiset.rs
```diff
@@ -1,583 +1,198 @@
...
+        add_primitive!(eg, "multiset-remove" = |mut xs: @MultiSetContainer<Value> (self.clone()), x: # (self.element())| -> @MultiSetContainer<Value> (self.clone()) { MultiSetContainer(xs.0.remove(&x)?) });
+
+        add_primitive!(eg, "multiset-length"       = |xs: @MultiSetContainer<Value> (self.clone())| -> i64 { xs.0.len() as i64 });
+        add_primitive!(eg, "multiset-contains"     = |xs: @MultiSetContainer<Value> (self.clone()), x: # (self.element())| -?> () { ( xs.0.contains(&x)).then_some(()) });
+        add_primitive!(eg, "multiset-not-contains" = |xs: @MultiSetContainer<Value> (self.clone()), x: # (self.element())| -?> () { (!xs.0.contains(&x)).then_some(()) });
+
+        add_primitive!(eg, "multiset-sum" = |xs: @MultiSetContainer<Value> (self.clone()), ys: @MultiSetContainer<Value> (self.clone())| -> @MultiSetContainer<Value> (self.clone()) { MultiSetContainer(xs.0.sum(ys.0)) });
+
+        // let inner_name = self.element.name();
+        // let fn_sort = eg.type_info.get_sort_by(|s: &Arc<FunctionSort>| {
```
ezrosent, line 142: Do we still need this commented out portion?

src/sort/multiset.rs
```diff
@@ -616,3 +231,105 @@
 //         Some(new_multiset.store(&self.multiset))
 //     }
 // }
+
+// Place multiset in its own module to keep implementation details private from sort
+mod inner {
+    use im::OrdMap;
```
ezrosent, line 237: do you have thoughts on `im` vs. `rpds`?

src/sort/multiset.rs
```diff
@@ -616,3 +231,105 @@
...
+    use std::hash::Hash;
+    /// Immutable multiset implementation, which is threadsafe and hash stable, regardless of insertion order.
+    ///
+    /// All methods that return a new multiset take ownership of the old multiset.
+    #[derive(Debug, Hash, Eq, PartialEq, Clone)]
+    pub(crate) struct MultiSet<T: Clone + Hash + Ord>(
+        /// All values should be > 0
+        OrdMap<T, usize>,
+        /// cached length
+        usize,
```
ezrosent, line 247: should we cache the hash code as well

src/sort/set.rs
```diff
@@ -1,593 +1,205 @@
...
-//         Some(set1.store(&self.set))
-//     }
-// }
+use super::*;
+use std::collections::BTreeSet;
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+struct SetContainer<V> {
+    do_rebuild: bool,
+    data: BTreeSet<V>,
```
ezrosent, line 7: why a std container here, but immutable/persistent for multiset?

src/sort/vec.rs
```diff
@@ -1,638 +1,242 @@
...
+            false
+        }
+    }
+    fn iter(&self) -> impl Iterator<Item = core_relations::Value> + '_ {
+        self.data.iter().copied()
+    }
+}
+
+#[derive(Debug)]
+pub struct VecSort {
```
ezrosent, line 23: Long-term, will we need to have these sorts around that carry a Mutex<IndexSet<...>>? The ContainerEnv structure in the backend crate serves a similar purpose.

src/sort/add_primitive/src/lib.rs
```diff
@@ -0,0 +1,395 @@
+use proc_macro::{Span, TokenStream};
+use quote::quote;
+use syn::parse::{Parse, ParseStream, Result};
+use syn::punctuated::Punctuated;
+use syn::{braced, bracketed, parenthesized, parse_macro_input, Expr, Ident, LitStr, Token};
+///
```
ezrosent, line 6: nit: don't need this leading line.

ezrosent (commented):
Change looks good! just a couple of questions.

src/sort/vec.rs
Alex-Fischman: Remember that the current plan is to have both backends running together until we get all (or almost all) of the tests passing, and then to rip out the old backend. You are totally right that this is doing the same job as `ContainerEnv`, but it's necessary for the old backend.

src/sort/multiset.rs
Alex-Fischman: Why would we? We cache the length because `len` should be O(1), but we wouldn't expect `hash` to be O(1).

src/sort/multiset.rs
Alex-Fischman: I think of `im` as the default. I'm not familiar with `rpds`.

src/sort/multiset.rs
Alex-Fischman: I think it's useful to keep until we make a decision on `unstable-fn` (see the Slack thread). I can uncomment it and fill in `todo!`s if that's preferrable.

src/sort/set.rs
Alex-Fischman: Dunno, I just copied what was there. I would guess that they should all be immutable. I'll ask in Slack.

src/sort/multiset.rs
ezrosent: Right... but we hash every one of these at least once because we're always inserting/looking it up in a hash table. May be worth caching the hash code?

Fine to leave this as a follow-up, just a thought.

src/sort/vec.rs
ezrosent: ah yes, of course!

ezrosent (approved):

src/sort/multiset.rs
ezrosent: commented out here is fine I think

PR #534, opened 2025-04-04
Subsumption in frontend

Alex-Fischman:
Wires subsumption up to the new backend. Passes `subsume.egg` if the `extract`s are removed.


ezrosent (approved):
Very clean! thanks for wiring this in

PR #539, opened 2025-04-11
Egglog proof format and checker

oflatt:


src/termdag.rs
```diff
@@ -186,6 +186,12 @@ impl TermDag {
 
         String::from_utf8(result).unwrap()
     }
+
+    pub fn parse(&mut self, str: &str) -> Result<Term, ParseError> {
+        // Ok to create new parser here, since parsing expressions doesn't use symbol gen
```
Alex-Fischman, line 191: Parser also supports macros now

src/proofs/proof.rs
```diff
@@ -0,0 +1,514 @@
...
+}
+
+#[derive(Clone, PartialEq, Eq, Hash, Debug)]
+pub enum Proposition {
+    TOk(Term),
+    TEq(Term, Term),
+}
+
+// todo how to ignore this warning?
+#[warn(clippy::enum_variant_names)]
```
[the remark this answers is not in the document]
Alex-Fischman: Oh if the specific enum names matter you can use `#[allow(clippy::enum_variant_names)]` iirc

src/proofs/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod proof;
```
Alex-Fischman, line 1: nit: file is missing ending newline

src/proofs/proof.rs
```diff
@@ -0,0 +1,514 @@
...
+    PCong {
+        pf_args_eq: Vec<ProofId>,
+        pf_f_args_ok: ProofId,
+        fun_sym: Symbol,
+    },
+}
+
+#[derive(PartialEq, Eq, Debug)]
+pub enum ProofCheckError {
+    Todo,
```
Alex-Fischman, line 79: This is a weird error condition; using `todo!` instead gives both error messages and line number

oflatt: This is a draft PR- I plan to fix these todos

src/termdag.rs
oflatt: Does this mean I can't make a parser here?

Alex-Fischman: Making a new parser here "forgets" all the macros, yes.

Is there a reason that this can't be a method on the `EGraph`? Like
```rust
pub fn parse_expr_and_store_in_termdag(&self, termdag: &mut Termdag, str: &str) -> Result<Term, ParseError> {
    Ok(termdag.expr_to_term(&self.parser.get_expr_from_string(None, str)?))
}
```

(An example usage of a macro is [here](https://github.com/egraphs-good/egglog-experimental/blob/main/src/sugar/with_ruleset.rs).)

Alex-Fischman: Wait actually everything is already public so this method can just accept a `Parser` (or an `EGraph` if you want).

oflatt: I made it for convenience, perhaps I'll just make it a testing function.

src/proofs/proof.rs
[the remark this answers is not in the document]
ezrosent: (I do think that the leading P isn't great.)

src/proofs/proof.rs
```diff
@@ -0,0 +1,514 @@
...
+            }
+        }
+    }
+
+    fn get_rule_by_name<'a>(
+        &self,
+        prog: &'a Vec<Command>,
+        name: Symbol,
+    ) -> Result<&'a Rule, ProofCheckError> {
+        for command in prog {
```
ezrosent, line 110: nit, `prog.iter().find_map()` 
and `prog: &'a [Command]`

src/proofs/proof.rs
```diff
@@ -0,0 +1,514 @@
...
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub enum ProofTerm {
+    /// proves a Proposition based on a rule application
+    /// the subsitution gives the mapping from variables to terms
+    /// the body_pfs gives proofs for each of the conditions in the query of the rule
+    /// the act_pf gives a location in the action of the proposition
+    PRule {
+        rule_name: Symbol,
+        subst: Substitution,
+        body_pfs: Vec<ProofId>,
```
ezrosent, line 45: What do these "body_pfs" mean? are they premises? Will they be a bunch of equality proofs or TOK proofs?

src/proofs/proof.rs
```diff
@@ -0,0 +1,514 @@
...
+#[derive(Clone, PartialEq, Eq, Hash, Debug)]
+pub enum Proposition {
+    TOk(Term),
+    TEq(Term, Term),
+}
+
+// todo how to ignore this warning?
+#[warn(clippy::enum_variant_names)]
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub enum ProofTerm {
```
ezrosent, line 37: Do we need stuff for primitives here?

src/proofs/proof.rs
oflatt: They are proofs for each line of the rule's premises
They produces a `Proposition` that matches the rule's syntax (a TOK proof for an expression and a equality proof for an (= t1 t2) line)

src/proofs/proof.rs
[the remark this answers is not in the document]
oflatt: It's to match the rocq checker, I'm leaving as-is

src/proofs/proof.rs
```diff
@@ -0,0 +1,514 @@
...
+    /// The symmetric equality of eq_pf
+    PSym {
+        eq_pf: ProofId,
+    },
+    PTrans {
+        pfxy: ProofId,
+        pfyz: ProofId,
+    },
+    /// get a proof for the child of a term given a proof of a term
+    PProj {
```
ezrosent, line 63: Does this assume that rules only ever insert one term?

e.g. if I had a rule:

```
(rule ((foo x)) ((bar (Num x)) (baz x))
```
What does the proof of `(bar ...)` look like as opposed to `(baz ...)`. 

I take `PProj` to be helping with getting access to the `(Num ...)` subterm if the proof is of `bar`.

src/proofs/proof.rs
```diff
@@ -0,0 +1,514 @@
...
+}
+
+// todo how to ignore this warning?
+#[warn(clippy::enum_variant_names)]
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub enum ProofTerm {
+    /// proves a Proposition based on a rule application
+    /// the subsitution gives the mapping from variables to terms
+    /// the body_pfs gives proofs for each of the conditions in the query of the rule
+    /// the act_pf gives a location in the action of the proposition
```
ezrosent, line 42: Where is `act_pf` ?

ezrosent (commented):
High-level thoughts on resugaring. We've talked about some of this before but wanted to re-confirm.

* We get these `MappedQuery` and `MappedActions` structures in the typechecking phase that generate `CorrespondingVar`s. 
* The `egglog-bridge` proofs work on "flattened" rules where each `CorrespondingVar` is its own atom.
* To resugar the LHS, take the proof from egglog-bridge, then add `PCong` rules translating a flattened LHS into a properly-nested one. 
  * **Question:** will the resulting proofs be "more gross than they should be?" Maybe there is a simplification pass we can do that eliminates no-op `PCongs`. 
  * Observation: I'm making this sound like a `ProofTerm`->`ProofTerm` transform but really it's a function from `FlatProof->ProofTerm`, the flattened proofs aren't going to be valid proof terms because they don't don't refer back to the resugared rules. 
* **Question** Do we need to do any resugaring of the RHS? Depending on how we encode `act_pf` I assume not?

src/proofs/proof.rs
oflatt: I removed `act_pf` because it seems like the rocq proof checker doesn't use it... need to ask @Tia-Vu about this

src/proofs/proof.rs
oflatt: I don't think we want to make that assumption... as far as I can tell, you can pick any term or sub-term in the rhs of a rule application when you use `PRule`.

That begs the question of why we need `PProj` at all. I'm not really sure, perhaps we won't need it?

oflatt:
> Question: will the resulting proofs be "more gross than they should be?" Maybe there is a simplification pass we can do that eliminates no-op PCongs.
Good question- probably they will have no-op PCongs we should get rid of

> Question Do we need to do any resugaring of the RHS? Depending on how we encode act_pf I assume not?

I don't think so- it seems like we don't need `act_pf` at all looking at the rocq checker??
The current checker I'm working on in rust doesn't need any proofs for actions.

oflatt:
@Tia-Vu and I are discussing-
The rocq checker *does* use action_proofs. I found my checker easier to write by ignoring them. We'll figure out if we need them, still up in the air.

Tia-Vu:
> Sounds like we should provide action proofs because they make the checker much easier to verify!

I actually just took a second look I think you’re good to delete act_pf I just took them out of the rocq checker 👀.

oflatt:
Yay!

Issue #546, opened 2025-04-17
Old and new backends don't agree as to whether deleting an entry counts as a change


RiscInside:
(Apologies if this was known, but I figured this is an edge case and there are no tests in egglog testing against this)

Trying to run new backend on this code would lead to a crash as old and new backend don't agree on if the database has changed.
```
(relation A ())
(A)
(rule ((A)) ((delete (A))))
(run 1)
```
```
[INFO ] Declared function A.
[INFO ] Declared rule (rule ((A ))
          ((delete (A )))
             ).
thread 'main' panicked at src/lib.rs:1081:13:
assertion `left == right` failed
  left: false
 right: true
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
Relevant code:
```rs
            let mut rule_ids = Vec::new();
            collect_rule_ids(ruleset, &self.rulesets, &mut rule_ids);
            let new_updated = self.backend.run_rules(&rule_ids).unwrap();
            assert_eq!(old_updated, new_updated); // <- this assert fails
```
It's not a big deal, but I thought it was worth reporting just to see what the correct answer should be.

Alex-Fischman:
Added as a test in #547; thank you for this! May I ask what you're using the new backend for? I was not expecting anyone outside of the egglog team to use it at this stage.

RiscInside:
> Added as a test in #547; thank you for this! May I ask what you're using the new backend for? I was not expecting anyone outside of the egglog team to use it at this stage.

Nothing really, just wanted to try it out.

Alex-Fischman:
Closing this as #547 adds the test; we are committed to having all tests passing before merging the new backend into main.

PR #547, opened 2025-04-17
Add repro-new-backend tests

Alex-Fischman:
Adds a minimized `prims.egg` test and the test from #546.

Both tests are currently failing.


Alex-Fischman:
Oliver says that we should make sure to add `repro` tests whenever we minimize a test and find a bug.

I have arbitrarily chosen `repro-new-backend` as a prefix for tests where the old backend and the new backend disagree.

yihozhang (approved):

PR #554, opened 2025-04-29
Add test for failing new backend example

saulshanabrook:
Adds a test case that fails on the new backend with `Panic: Illegal merge attempted for function TupleInt_to_vec`.


Alex-Fischman:
Can you change this PR to be onto the `backend-merge` branch?

Alex-Fischman (approved):
Approved for the `backend-merge` branch

PR #559, opened 2025-05-06
Fix parser ignoring errors

Alex-Fischman:
Thanks to @FTRobbin for finding this; on line 652 of `parse.rs`, we were flattening a `Result` iterator instead of propagating the error. This had the effect of silently discarding top-level actions that failed to parse!

I've also added type annotations to the other sites in this file where we use `flatten` to make sure that this isn't happening there.


Alex-Fischman:
This PR is onto `main`, but we should also get this commit onto `backend-merge`.

yihozhang (approved):
great catch, thanks

PR #560, opened 2025-05-07
Make union-find struct externally accessible (for container implementation)

RiscInside:
Currently it's not possible to implement `canonicalize` method of `Sort` trait on container sorts defined outside of this crate due to the use of private `UnionFind` type. This PR exposes `UnionFind` and `Id` types (`Id` was already part of the `union` method on e-graph). The only publicly accessible method on `UnionFind` is `find` (since it already has public EGraph counterpart).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -58,7 +58,7 @@ use std::sync::Arc;
 pub use termdag::{Term, TermDag, TermId};
 use thiserror::Error;
 pub use typechecking::TypeInfo;
-use unionfind::*;
+pub use unionfind::*;
 use util::*;
 pub use value::*;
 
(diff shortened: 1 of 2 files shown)


Alex-Fischman:
I think you can delete `union_raw`?

Alex-Fischman:
We decided to merge this PR for a v0.5 release right before we switch to the new backend. But know that if you depend on this PR, it will break in v0.6, which we expect to release before June 16.

Alex-Fischman:
We think it will be good to fix the issue for v0.5. Thank you for your contribution!

Alex-Fischman (approved):

PR #563, opened 2025-05-09
Reuse Extractor whenever possible

yihozhang:
TODO

Add test


Alex-Fischman:
I think we aren't planning to merge this, so mark as draft?

Issue #551, opened 2025-04-24
web demo does not escape error messages


yihozhang:
![Image](https://github.com/user-attachments/assets/1959d738-8711-4231-b11a-9d63089ba7af)

PR #564, opened 2025-05-09
Fix web demo escaping

saulshanabrook:
Fixes https://github.com/egraphs-good/egglog/issues/551 by using `textContent` instead of `innerHTML` in the web demo so that content is properly escaped. As [recommended by MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#security_considerations)

Screenshot 2025-05-09 at 2 50 38 PM


yihozhang (approved):
thank you for looking into this!

PR #569, opened 2025-05-19
Fix value generation for relations in Change::Subsume code

RiscInside:
This PR de-duplicates code for fresh output generation, fixing a bug with relation subsumption: old code for subsume just grabbed a new value from UF, which is not the right way to generate outputs for relations.

Let me know if `NewOutputGen` stuff is too much - I can do a simpler version going directly to a new fresh `Value` (it would be less code, but it would require an extra unwrap() call).

### DIFF
--- a/tests/subsume-relation.egg
+++ b/tests/subsume-relation.egg
@@ -1,0 +1,9 @@
+(datatype V (A) (B) (C))
+(relation R (V V))
+
+(A) (B) (C)
+(subsume (R (A) (B)))
+(R (A) (C))
+(union (B) (C))
+
+(check (= (R (A) (B)) ()))
--- a/src/actions.rs
+++ b/src/actions.rs
@@ -392,20 +392,29 @@ impl EGraph {
                             function.remove(args, self.timestamp);
                         }
                         Change::Subsume => {
-                            if function.decl.subtype != FunctionSubtype::Constructor
-                                && function.decl.subtype != FunctionSubtype::Relation
-                            {
-                                return Err(Error::SubsumeMergeError(*f));
-                            }
-                            function
-                                .nodes
-                                .insert_and_merge(args, self.timestamp, true, |old| {
-                                    old.unwrap_or_else(|| Value {
-                                        #[cfg(debug_assertions)]
-                                        tag: function.schema.output.name(),
-                                        bits: self.unionfind.make_set(),
-                                    })
-                                });
+                            match function.decl.subtype {
+                                FunctionSubtype::Constructor => function.nodes.insert_and_merge(
+                                    args,
+                                    self.timestamp,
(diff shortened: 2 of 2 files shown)


Alex-Fischman:
I don't think we need a trait, why not just match on the subtype in the Subsume case? You can do
```rust
match subtype {
    Constructor => ...,
    Relation => ...,
    _ => return Err(...),
}
```

RiscInside:
> I don't think we need a trait, why not just match on the subtype in the Subsume case? You can do
> 
> ```rust
> 
> match subtype {
> 
>     Constructor => ...,
> 
>     Relation => ...,
> 
>     _ => return Err(...),
> 
> }
> 
> ```

I was not doing that since I wanted to avoid duplicating the code for output id generation, but I will change the PR to do that if it's not an issue.

tests/subsume-relation.egg
```diff
@@ -0,0 +1,10 @@
+(datatype V (A) (B) (C))
+(relation R (V V))
+
+(A) (B) (C)
+(subsume (R (A) (B)))
+(R (A) (C))
+(union (B) (C))
+
+(rule ((= (B) (C))) ((union (B) (C))))
+(run 1)
```
Alex-Fischman, line 10: I don't understand the last part of the test case; this rule should be a no-op, since if `(B) = (C)` then `(B)` and `(C)` must have been unioned already. Why do you add a rule here?

RiscInside: https://egraphs-good.github.io/egglog/?program=XQAAgABrAAAAAAAAAAAUGQgnfMUD9dO10KKlCABufHf1Kz7EbNMQ5urHa3iQgKHj5tBOo4YVRWUcAnht_pCc4LXP_JubFRajykKtxFtcXVUuxuLi8P7mDMA%253D

it doesn't I think. Rebuilding happens in run_rules seemingly.

Alex-Fischman: Rebuilding should happen without the `(run)` though; otherwise stuff like
```
(union (B) (C))
(extract (B))
```
wouldn't be correct (assuming `(B)` costs more than `(C)`).

Alex-Fischman: Is what you actually want to do a `(check (= (R (A) (B)) ())`? I think this test is not minimal. Unless I'm seeing an XY problem?

RiscInside: Ah, I see what's happening now. Every command runs rebuilding *before* it runs itself (but, crucially, not after). I can add any command after union and it would panic before it even gets to execution (that could be check or run).

With this in mind, should I switch to using run or check?

Alex-Fischman (approved):
Looks good to me. Thanks!

PR #570, opened 2025-05-21
Much Faster Push Implementation

oflatt:
Previously, the push implementation cloned the entire egraph... including previously pushed egraphs

So a sequence of n pushes took `O(n^2 * size_of_initial_egraph)` space
This PR fixes that so it takes n * size_of_initial_egraph space

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -429,7 +429,7 @@ impl FromStr for RunMode {
 pub struct EGraph {
     pub parser: Parser,
     names: check_shadowing::Names,
-    prev_egraph: Box<Option<Self>>,
+    prev_egraph: Option<Box<Self>>,
     unionfind: UnionFind,
     pub functions: IndexMap<Symbol, Function>,
     rulesets: IndexMap<Symbol, Ruleset>,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -429,7 +429,7 @@ impl FromStr for RunMode {
 pub struct EGraph {
     pub parser: Parser,
     names: check_shadowing::Names,
-    egraphs: Vec<Self>,
+    prev_egraph: Box<Option<Self>>,
```
Alex-Fischman, line 432: nit: `pushed_egraph` gives more clarity to readers without context

nit: `Box<Option<T>>` should usually be `Option<Box<T>>`, so there's no heap alloc for None

src/lib.rs
```diff
@@ -496,7 +496,11 @@ impl EGraph {
     }
 
     pub fn push(&mut self) {
-        self.egraphs.push(self.clone());
+        let mut prev_prev: Box<Option<Self>> = Default::default();
+        std::mem::swap(&mut self.prev_egraph, &mut prev_prev);
```
Alex-Fischman, line 500: nit: can you just use `Option::take` here?

Alex-Fischman (approved):
Just some nits

PR #571, opened 2025-05-22
Extraction as a top-level command in the frontend for the new backend

FTRobbin:
This PR implements extraction as a top-level command in the frontend. It uses the default additive tree cost model and can only be triggered through `(extract [term] [variant])`.

### Implementation overview:
- Extract as an action has been removed.
- `query-extract` as a top-level command has also been removed.
- `Extract` is now a `Command` and is interpreted in the frontend `run_command`. `Extract` first runs two backend queries to resolve the eclass and the number of variants to be extracted and then calls `ExtractorAlter` to perform the extraction.
- `ExtractorAlter` first needs to be initialized by passing the list of rootsorts to be extracted. It computes all relevant costs using Bellman-Ford and remembers the necessary information for reconstruction in this step. Then, it can be queried multiple times to reconstruct the terms in termdags. It does not hold a reference to `EGraph`, so it can be reused as long as the relevant tables do not change.
- As an optimization, in the initialization step, `ExtractorAlter` first performs a BFS on the sorts to filter out tables unreachable from the root sort.
- In the Bellman-Ford loop, it calls inner_sorts and inner_values on the sorts to unpack containers and calls the cost model for primitive, container, and function costs.
- After Bellman-Ford finds a fixpoint, `ExtractorAlter` enumerates all rows again to copy information for reconstruction. Ties are broken arbitrarily.
- For extracting a single lowest-cost term, it recursively reconstructs the lowest-cost enode for each eclass used in the lowest-cost enode of the root eclass and writes into a TermDag struct.
- For extracting multiple "diverse" variants, it extracts the multiple enodes of the root eclass, in the order of lowest costs, using a single lowest-cost enode for other classes.

### Additional remarks:
- This implementation does not involve concurrency control or tunneling between the frontend and the backend, so the code logic is simplified compared to #562.
- This new architecture potentially allows extraction to be used more easily from RustAPI.
- This implementation opens up the possibility of reusing extractor results, potentially subsuming #563 after the new backend merge.
- A caveat is that there is no easy way to port the old `Extractor` to this new architecture. So, no extraction in the old backend after this PR.


src/sort/mod.rs
```diff
@@ -51,6 +51,13 @@ pub trait Sort: Any + Send + Sync + Debug {
 
     fn column_ty(&self, backend: &egglog_bridge::EGraph) -> ColumnTy;
 
+    // return the inner sorts if a container sort
+    // remember that containers can contain containers
+    // and this only unfold one level
+    fn inner_sorts(&self) -> Vec<&Arc<dyn Sort>> {
```
Alex-Fischman, line 57: Is `inner_sorts` missing for `FunctionSort`? If it is incorrect to omit for containers, the default implementation should panic if `self.is_container_sort()` is true.

src/sort/multiset.rs
```diff
@@ -203,6 +205,15 @@ impl Sort for MultiSetSort {
         }
         Some((cost, termdag.app("multiset-of".into(), children)))
     }
+    fn reconstruct_termdag_container(
```
Alex-Fischman, line 208: nit: add space between methods

tests/integration_test.rs
```diff
@@ -1,6 +1,228 @@
 use egglog::{ast::Expr, *};
 use symbol_table::GlobalSymbol;
 
+#[test]
```
Alex-Fischman, line 4: Since we aren't testing the new extractor against the old one automatically, should these tests (the ones you added in `integration_tests.rs`) include the expected output?

Cargo.toml
```diff
@@ -57,12 +57,13 @@ rustc-hash = "1.1"
 smallvec = "1.11"
 symbol_table = { version = "0.4.0", features = ["global"] }
 thiserror = "1"
+queues = "1.1.0"
```
Alex-Fischman, line 60: What's wrong with https://doc.rust-lang.org/std/collections/struct.VecDeque.html?

Cargo.toml
```diff
@@ -57,12 +57,13 @@ rustc-hash = "1.1"
...
 
 # Backend
 # TODO: move egglog-backend repo into core egglog repo
-core-relations = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "e1c938d" }
-egglog-bridge = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "e1c938d" }
-numeric-id = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "e1c938d" }
+core-relations = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "4f52312" }
+egglog-bridge = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "4f52312" }
+numeric-id = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "4f52312" }
```
Alex-Fischman, line 66: Am I okay to close https://github.com/egraphs-good/egglog-backend/pull/29?

src/lib.rs
```diff
@@ -662,7 +661,9 @@ impl EGraph {
 
     fn apply_merges(&mut self, func: Symbol, merges: &[DeferredMerge]) -> usize {
         let mut stack = Vec::new();
-        let mut function = self.functions.get_mut(&func).unwrap();
+        let functions_clone = self.functions.clone();
+        let mut functions_borrow_mut = functions_clone;
+        let mut function = functions_borrow_mut.get_mut(&func).unwrap();
```
Alex-Fischman, line 666: nit: is this leftover from action-extraction? It looks unnecessary

src/lib.rs
```diff
@@ -1692,17 +1765,20 @@ impl EGraph {
     }
 
     pub(crate) fn print_msg(&mut self, msg: String) {
-        if let Some(ref mut msgs) = self.msgs {
+        if let Some(msgs) = &mut self.msgs {
```
Alex-Fischman, line 1768: nit: can we remove the diffs for `print_msg` and `flush_msgs`? (I know what caused them but they aren't an improvement)

src/extract.rs
```diff
@@ -70,7 +67,7 @@ impl EGraph {
                     }
                 }
             }
-            Error::ExtractError(value)
+            Error::ExtractError(String::new())
```
Alex-Fischman, line 70: Can we put a real error message here?

src/extract.rs
```diff
@@ -214,3 +211,484 @@ impl<'a> Extractor<'a> {
         }
     }
 }
+
+pub trait CostModel {
```
Alex-Fischman, line 215: I don't think everyone agrees that this is the best cost model yet. Hopefully we can get to it in the meeting tomorrow.

src/extract.rs
```diff
@@ -214,3 +211,484 @@ impl<'a> Extractor<'a> {
...
+    parent_edge:
+        HashMap<Symbol, HashMap<core_relations::Value, (Symbol, Vec<core_relations::Value>)>>,
+}
+
+impl ExtractorAlter {
+    /// Bulk of the computation happens at initialization time
+    /// The later extractions only reuses saved results
+    /// The means a new extractor must be created if the egraph changes
+    /// Holding a reference to the egraph would enforce this but prevents the extractor being reused
+    /// For convenience, if the rootsorts is None, it defaults to extract all extractable rootsorts
```
Alex-Fischman, line 290: nit: doc comments should be complete sentences

Alex-Fischman (changes requested):
I have not reviewed the actual extraction algorithm. My comments are about style and software engineering.

src/sort/mod.rs
```diff
@@ -107,12 +114,12 @@ pub trait Sort: Any + Send + Sync + Debug {
     /// Only container sort need to implement this method,
     fn inner_values(
         &self,
-        egraph: &EGraph,
+        containers: &core_relations::Containers,
```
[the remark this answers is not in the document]
FTRobbin: Yes, extraction still unfolds things.

src/sort/mod.rs
```diff
@@ -139,6 +146,53 @@ pub trait Sort: Any + Send + Sync + Debug {
...
     ) -> Option<(Cost, Term)>;
+
+    /// Default cost for containers when the cost model does not specify the cost
+    fn default_container_cost(
+        &self,
+        _containers: &core_relations::Containers,
+        _value: core_relations::Value,
+        element_costs: &[Cost],
+    ) -> Cost {
+        element_costs.iter().fold(0, |s, c| s.saturating_add(*c))
```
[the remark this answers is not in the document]
FTRobbin: They are free in the old backend.

Cargo.toml
FTRobbin: I don't think there is anything wrong with a deque. But only a queue is needed.

Cargo.toml
FTRobbin: Yes.

src/lib.rs
```diff
@@ -1414,64 +1539,12 @@ impl EGraph {
...
-                let _ = self.backend.run_rules(&[id]).unwrap();
-                self.backend.free_rule(id);
-                self.backend.free_external_func(ext_id);
+                    let value = self.eval_resolved_expr_new(span.clone(), &expr)?;
+                    let expr_type = expr.output_type();
 
-                let results: Vec<_> = std::mem::take(results.lock().unwrap().as_mut());
-                use std::io::Write;
-                for (value, expr_type) in results.into_iter().zip(expr_types) {
                     // hack before extraction for the new backend gets merged:
```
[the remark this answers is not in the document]
FTRobbin: This would require Yihong. I don't know what this code is doing on top of my head.

src/extract.rs
FTRobbin: This is the old extract. I'm not sure what cases would cause an error besides trying to extract an unextractable value.

src/sort/mod.rs
FTRobbin: Fixed.

src/lib.rs
FTRobbin: Fixed.

src/extract.rs
Alex-Fischman: Right, either we should maintain the way it used to work (maybe just use `format!`?) or don't take an argument in this error variant and put the string that you used in the new backend into the `#[error()]` message.

Cargo.toml
Alex-Fischman: If there's no reason to not use the standard library version I would rather not pull in the extra dependency.

src/lib.rs
Alex-Fischman: Unresolving, I still see diffs

src/ast/desugar.rs
```diff
@@ -326,29 +273,13 @@ fn desugar_simplify(
...
     let lhs = parser.symbol_gen.fresh(&"desugar_simplify".into());
-    res.push(NCommand::CoreAction(Action::Let(
-        span.clone(),
-        lhs,
-        expr.clone(),
-    )));
-    res.push(NCommand::RunSchedule(schedule.clone()));
-    res.extend(
-        desugar_command(
-            Command::QueryExtract {
```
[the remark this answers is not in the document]
Alex-Fischman: (But we should probably not let it become silently different.)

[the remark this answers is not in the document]
saulshanabrook: This came up before, I support simplify in the python bindings and use it, but I could just change it to be desugared manually there since it's so simple.

oflatt (changes requested):
did a quick and rough review

src/extract.rs
```diff
@@ -214,3 +211,484 @@ impl<'a> Extractor<'a> {
...
+            func.decl.name,
+            &ch_costs,
+            self.cost_model.enode_cost(egraph, func, row),
+        ))
+    }
+
+    fn bellman_ford(&mut self, egraph: &EGraph) {
+        // We use Bellman-Ford to compute the costs of the relevant eq sorts' terms
+
+        let mut ensure_fixpoint = false;
```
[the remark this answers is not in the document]
Alex-Fischman: In `RunReport` it's called `updated`.

src/extract.rs
```diff
@@ -214,3 +211,484 @@ impl<'a> Extractor<'a> {
...
+        HashMap<Symbol, HashMap<core_relations::Value, (Symbol, Vec<core_relations::Value>)>>,
+}
+
+impl ExtractorAlter {
+    /// Bulk of the computation happens at initialization time.
+    /// The later extractions only reuses saved results.
+    /// This means a new extractor must be created if the egraph changes.
+    /// Holding a reference to the egraph would enforce this but prevents the extractor being reused.
+    /// For convenience, if the rootsorts is None, it defaults to extract all extractable rootsorts.
+    pub fn compute_costs_from_rootsorts(
```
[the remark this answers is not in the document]
FTRobbin: I think `extract` is somewhat overloaded here. Extractors, in general, don't really `compute` costs as costs are given by the cost model as an oracle. I prefer the way it is.

src/extract.rs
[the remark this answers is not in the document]
FTRobbin: I'm using this in combination with the `reconstruction_round` flag. I think it is clearer this way.

src/extract.rs
```diff
@@ -214,3 +211,508 @@ impl<'a> Extractor<'a> {
...
+                Some((best_cost, term))
+            }
+            None => {
+                log::error!("Unextractable root {:?} with sort {:?}", value, sort,);
+                None
+            }
+        }
+    }
+
+    pub fn extract_best(
```
oflatt, line 608: document invariant- extract_best requires a value that matches the rootsort the extractor was created with

oflatt (approved):
nice!

src/extract.rs
Alex-Fischman: Are we okay with merging this temporarily? Should this be `pub` in this PR? Tagging @oflatt @yihozhang.

src/lib.rs
```diff
@@ -1200,6 +1199,69 @@ impl EGraph {
         Ok((sort, value_new_backend))
     }
 
+    fn eval_resolved_expr(
```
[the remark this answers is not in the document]
Alex-Fischman: I guess so, but would this really reduce code duplication? Not sure

src/lib.rs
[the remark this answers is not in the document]
Alex-Fischman: @yihozhang

oflatt (approved):
never mind, the external function is necessary

src/ast/desugar.rs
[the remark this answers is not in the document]
Alex-Fischman: Approval to remove granted by @oflatt on Slack

src/lib.rs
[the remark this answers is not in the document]
yihozhang: This code extracts a string representation for every value that is outputted, so it should just be a call to extract (we may want to reuse the extraction initialization)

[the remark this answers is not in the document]
FTRobbin: Could you double-check that the latest is what you want?

src/extract.rs
FTRobbin: I'm going to resolve this according to the discussion on Slack.

src/lib.rs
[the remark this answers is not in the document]
yihozhang: eval_action does not give you the result, so we have to do this

Alex-Fischman (approved):
All of my comments have been resolved

PR #573, opened 2025-05-22
Rebuild after running commands instead of before

Alex-Fischman:
Running rebuilding before running the commands means that the results from the Rust API would be wrong after running e.g. a top level `union`.

This is an old backend/v0.5 kind of fix.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1200,17 +1200,6 @@ impl EGraph {
     }
 
     fn run_command(&mut self, command: ResolvedNCommand) -> Result<(), Error> {
-        let pre_rebuild = Instant::now();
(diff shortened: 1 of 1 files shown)


Alex-Fischman:
@RiscInside Thanks for bringing this up in https://github.com/egraphs-good/egglog/pull/569#discussion_r2103353024, the new backend fixes this but I think it's worth doing in the old backend too.

yihozhang (approved):

RiscInside:
For the union, wouldn't it be a good idea to expose a non-rebuilding version of it in the API? Eagerly rebuilding after every union seems a bit excessive (and if actions such as extract or check expect the graph to be rebuilt, they can do it themselves).

PR #578, opened 2025-05-24
Merge main into backend-merge

Alex-Fischman:
I want to fix all the merge conflicts before deleting the old backend.

PR #579, opened 2025-05-24
Delete the old backend

Alex-Fischman:
- The big one: deletes `Value` and everything that uses `Value`.
- Change all `core_relations::Value` to `Value` with a top-level import.
- Make all `PrimitiveLike` implement `ExternalFunction` automatically.
- Rename `Primitive` to `PrimitiveWithId` and move it to `typechecking.rs`.
- Rename `PrimitiveLike` to `Primitive`.
- Move `Function` to `lib.rs`, delete `function` folder.
- Remove now-unused `concat_vecs` and `ListDebug` helpers.
- Remove now-unused `smallvec` dependency.
- Make the `backend` and `function` fields private.
- Fix a bug in the new serializer.
- Fix some fragile tests (one is still fragile).
- Fix extraction performance bug in reconstruction by adding a cache.
- Cleanup: don't use `&Value` as an argument type, use `Value` instead.
- Fixes TODO in `eval_expr` now that we don't need to support the old backend.
- Add back `RunReport` after https://github.com/egraphs-good/egglog-backend/pull/31.
- Rename `ExtractorAlter` to `Extractor`.
- Make `RunReport` unions mutable.

(Also note that I slipped #578 in behind y'all's backs.)

It may be easier to review this PR commit-by-commit.


ezrosent:
I figure once we get codespeed running we'll also want to assess performance parity before undrafting?

Alex-Fischman:
I'm okay with doing pure performance in a separate PR, but I don't think we should delete the old backend if not all tests are passing.

Alex-Fischman:
`blame` brings up #487

Alex-Fischman:
@yihozhang Can you look into the failing tests? It seems like you were the one to most recently touch the line in question, and you wrote the new serialization.

Alex-Fischman:
I'm adding everyone as a reviewer because this PR deletes a bunch of stuff, some of which you may depend on.

src/lib.rs
```diff
@@ -422,17 +170,48 @@ pub struct EGraph {
     msgs: Option<Vec<String>>,
 }
 
+#[derive(Clone)]
+pub struct Function {
+    pub(crate) decl: ResolvedFunctionDecl,
+    pub schema: ResolvedSchema,
+    pub(crate) can_subsume: bool,
+    pub backend_id: egglog_bridge::FunctionId,
```
oflatt, line 178: why is the backend_id pub?
Could be useful for some reason I just don't know

Alex-Fischman: It's used in the integration tests, but I can make it private and just move those tests to  `src`.

saulshanabrook:
Great that all the examples are working! Is this benchmark accurate? Are the performance degradations that severe?

Alex-Fischman:
Bugfixes are in, all tests should pass now.

ezrosent (approved):
Looks good. Thanks Alex!

Most of my questions were resolved in the other threads re: run report and failing tests.

FTRobbin (approved):

Alex-Fischman:
I will rename the extractor. What do you mean by "handled by the termdag"? The termdag doesn't know about values, so how would it store the cache?

Alex-Fischman:
Merging now since Yihong is busy; I will fix any concerns he has though.

PR #580, opened 2025-05-25
Remove web demo

saulshanabrook:
Removes the web demo after I moved it to a separate repo at https://github.com/egraphs-good/egglog-demo

## Motivation

We wanted to support changing the web demo to use egglog experimental as well as main. If we kept it in this repo, it would introduce a circular github dependency, which while possible is not intuitive.

Instead, it was decided to move the web demo to a separate repo which can pull in both main and experimental. It currently only pulls in main, since the experimental repo does not work against main, but once it does it will be straightforward to build that as well.

## Details

In order to minimize the disruption to existing users of the web demo, the docs are kept at the same sub-path and a redirect file is added at the root to the new demo.

The [git history was also saved from the `web-demo` path](https://docs.github.com/en/get-started/using-git/splitting-a-subfolder-out-into-a-new-repository) when the new repo was created, so that the author list and changes can still be seen.


Cargo.toml
```diff
@@ -13,7 +13,6 @@ name = "example_benchmarks"
 harness = false
 
 [workspace]
```
Alex-Fischman, line 15: nit: remove empty workspace tag?

Alex-Fischman (approved):
LGTM!

PR #582, opened 2025-05-31
Rust API

Alex-Fischman:
Adds a prelude that can be imported to expose ways to build an e-graph without using string interpolation.

I am ready for people to review, only a draft because I don't want to stomp on the performance work which I think is higher priority.

### DIFF
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -33,21 +33,19 @@ pub mod sort {
 }
 
 /// Create a new ruleset.
-pub fn add_ruleset(egraph: &mut EGraph, ruleset: Symbol) -> Result<(), Error> {
-    egraph.run_program(vec![Command::AddRuleset(span!(), ruleset)])?;
-    Ok(())
+pub fn add_ruleset(egraph: &mut EGraph, ruleset: Symbol) -> Result<Vec<String>, Error> {
+    egraph.run_program(vec![Command::AddRuleset(span!(), ruleset)])
 }
 
 /// Run one iteration of a ruleset.
-pub fn run_ruleset(egraph: &mut EGraph, ruleset: Symbol) -> Result<(), Error> {
+pub fn run_ruleset(egraph: &mut EGraph, ruleset: Symbol) -> Result<Vec<String>, Error> {
     egraph.run_program(vec![Command::RunSchedule(Schedule::Run(
         span!(),
         RunConfig {
             ruleset,
             until: None,
         },
-    ))])?;
-    Ok(())
+    ))])
 }
 
 #[macro_export]
@@ -112,7 +110,7 @@ pub fn rule(
     ruleset: Symbol,
     facts: Facts<Symbol, Symbol>,
     actions: Actions,
-) -> Result<(), Error> {
+) -> Result<Vec<String>, Error> {
     let rule = Rule {
         span: span!(),
         head: actions,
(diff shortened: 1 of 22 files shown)


src/prelude.rs
```diff
@@ -0,0 +1,764 @@
...
+    /// For more information, see `egglog_bridge::TableAction::subsume`.
+    pub fn subsume(&mut self, table: &str, key: &[Value]) {
+        self.get_table_action(table).subsume(self.exec_state, key)
+    }
+
+    /// Panic.
+    /// You should also return `None` from your callback if you call
+    /// this function, which this function hopefully makes easier by
+    /// always returning `None` so that you can use `?`.
+    pub fn panic(&mut self) -> Option<()> {
```
ezrosent, line 186: Should this take a string argument for the panic message?

src/prelude.rs
```diff
@@ -0,0 +1,764 @@
...
+    let rule_name = egraph.parser.symbol_gen.fresh(&"prelude::rule".into());
+    egraph.run_program(vec![Command::Rule {
+        name: rule_name,
+        ruleset: ruleset.into(),
+        rule,
+    }])
+}
+
+/// A wrapper around an `ExecutionState` for rules that are written in Rust.
+pub struct RustRuleContext<'a, 'b> {
```
ezrosent, line 130: I think the methods on this struct should have doctests. Sim. for `query` and `rust_rule`.

src/prelude.rs
```diff
@@ -0,0 +1,764 @@
+use crate::*;
```
ezrosent, line 1: In the doc comments for this module, maybe include a few rust examples of simple rulesets being run, extraction, etc.?

src/prelude.rs
```diff
@@ -0,0 +1,764 @@
...
+pub struct QueryResult {
+    cols: usize,
+    data: Vec<Value>,
+}
+
+impl QueryResult {
+    /// Get an iterator over the query results,
+    /// where each match is a `&[Value]` in the same order
+    /// as the `vars` that were passed to `query`.
+    pub fn iter(&self) -> QueryResultIterator {
```
ezrosent: Is this "just" `self.data.iter().chunks(self.cols)` ?

Is there a reason we need a nameable iterater type?

src/prelude.rs
```diff
@@ -0,0 +1,764 @@
...
+                input: vec![$(stringify!($args).into()),*],
+                output: stringify!($sort).into(),
+            },
+            [$($cost)*].first().copied(),
+            false,
+        )?;)*
+    };
+}
+
+pub trait LeafSort: Any + Send + Sync + Debug {
```
ezrosent, line 420: Documentation (here and ContainerSort)

src/prelude.rs
Alex-Fischman: The problem is that the panic message needs to be known when the primitive is defined. You can see how I did that when the Context gets built. Maybe (tying into the backend performance PR) there's a way to pass the message as a Value efficiently?

src/prelude.rs
Alex-Fischman: Nope, just forgot about chunks.

Alex-Fischman:
Moving this PR to the adfisch-rust-api branch on this repo.

PR #586, opened 2025-06-04
Rust API

Alex-Fischman:
Adds a prelude that can be imported to expose ways to build an e-graph without using string interpolation.

I am choosing to block this on the open performance PR.

### DIFF
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -33,21 +33,19 @@ pub mod sort {
 }
 
 /// Create a new ruleset.
-pub fn add_ruleset(egraph: &mut EGraph, ruleset: Symbol) -> Result<(), Error> {
-    egraph.run_program(vec![Command::AddRuleset(span!(), ruleset)])?;
-    Ok(())
+pub fn add_ruleset(egraph: &mut EGraph, ruleset: Symbol) -> Result<Vec<String>, Error> {
+    egraph.run_program(vec![Command::AddRuleset(span!(), ruleset)])
 }
 
 /// Run one iteration of a ruleset.
-pub fn run_ruleset(egraph: &mut EGraph, ruleset: Symbol) -> Result<(), Error> {
+pub fn run_ruleset(egraph: &mut EGraph, ruleset: Symbol) -> Result<Vec<String>, Error> {
     egraph.run_program(vec![Command::RunSchedule(Schedule::Run(
         span!(),
         RunConfig {
             ruleset,
             until: None,
         },
-    ))])?;
-    Ok(())
+    ))])
 }
 
 #[macro_export]
@@ -112,7 +110,7 @@ pub fn rule(
     ruleset: Symbol,
     facts: Facts<Symbol, Symbol>,
     actions: Actions,
-) -> Result<(), Error> {
+) -> Result<Vec<String>, Error> {
     let rule = Rule {
         span: span!(),
         head: actions,
@@ -124,9 +122,7 @@ pub fn rule(
         name: rule_name,
         ruleset,
         rule,
-    }])?;
-
-    Ok(())
+    }])
 }
 
 /// A wrapper around an `ExecutionState` for rules that are written in Rust.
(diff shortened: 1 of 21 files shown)


Alex-Fischman:
Moved here from #582. Unresolved notes from Eli's review:
- Panic messages: currently Rust rules don't support error messages when panicking.
- Doctests for `RustRuleContext` methods, `query`, and `rust_rule`.
- Doctests for the `prelude` module.

src/sort/add_primitive/src/lib.rs
```diff
@@ -182,6 +191,7 @@ pub fn add_primitive(input: TokenStream) -> TokenStream {
     // This is the big `quote!` block that ties everything together.
     quote! {{
         #[allow(unused_imports)] use ::egglog::{*, constraint::*};
+        #[allow(unused_imports)] use ::std::any::TypeId;
```
yihozhang: I also noticed that Span and ExecutionState are used but not imported, so users of this macro may potentially get a type error.

src/prelude.rs
```diff
@@ -0,0 +1,755 @@
...
+    facts: Facts<Symbol, Symbol>,
+    actions: Actions,
+) -> Result<Vec<String>, Error> {
+    let rule = Rule {
+        span: span!(),
+        head: actions,
+        body: facts.0,
+    };
+
+    let rule_name = egraph.parser.symbol_gen.fresh(&"prelude::rule".into());
```
yihozhang: Would it be better if the rule name were a stringification of the actual rule?

We should also consider allowing the user to optionally pass a rule name if they want.

src/prelude.rs
```diff
@@ -0,0 +1,755 @@
...
+
+/// Add a rule to the e-graph whose right-hand side is a Rust callback.
+pub fn rust_rule(
+    egraph: &mut EGraph,
+    ruleset: &str,
+    vars: &[(&str, ArcSort)],
+    facts: Facts<Symbol, Symbol>,
+    func: impl Fn(&mut RustRuleContext, &[Value]) -> Option<()> + Clone + Send + Sync + 'static,
+) -> Result<Vec<String>, Error> {
+    let rule_name = egraph.parser.symbol_gen.fresh(&"prelude::rust_rule".into());
```
yihozhang: same here

src/sort/add_primitive/src/lib.rs
```diff
@@ -21,10 +21,14 @@ use syn::{braced, bracketed, parenthesized, parse_macro_input, Expr, Ident, LitS
...
 /// - Specialized Constraints: using `T (E)` as a type will use
-///   `E:expr` in the type constraint but `T:ty` in the body.
+///   `(E:expr).clone()` in the type constraint but `T:ty` in the body.
 ///   This is necessary because the relationship between Rust types
 ///   and egglog sorts is not 1-to-1.
 ///
+/// - Context: putting `{x: T}` between the `=` and the argument list
+///   will let you access the expression `x` of type `T` from inside
+///   the body as `self.ctx`. `T` must be the real Rust type of `x`.
+///   `T` must be `Clone` and `'static`.
```
yihozhang, line 33: nit: I got lost by the documentation of add_primitive

yihozhang (approved):

src/prelude.rs
```diff
@@ -0,0 +1,935 @@
...
+
+    pub fn call(f: &str, xs: Vec<Expr>) -> Expr {
+        Expr::Call(span!(), f.into(), xs)
+    }
+}
+
+pub mod sort {
+    use super::*;
+
+    pub fn i64() -> ArcSort {
```
FTRobbin: Why is this here?

src/prelude.rs
```diff
@@ -0,0 +1,935 @@
...
+/// ) (
+///     (set (fib (+ x 2)) (+ f0 f1))
+/// ))
+/// (run 10)
+///     ",
+/// )?;
+///
+/// let big_number = 20;
+///
+/// // check that `fib` does not contain `20`
```
FTRobbin: 20 is not a Fibonacci number. I think you mean `(fib 20)` does not exist in the db yet.

FTRobbin (approved):

PR #587, opened 2025-06-04
Scheduler interface with `choose`

yihozhang:
Closes #431

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,12 +56,13 @@ rustc-hash = "1.1"
 symbol_table = { version = "0.4.0", features = ["global"] }
 thiserror = "1"
 web-time = "1.1"
+dyn-clone = "1.0.17"
 
 # Backend
 # TODO: move egglog-backend repo into core egglog repo
-core-relations = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "4b35d00" }
-egglog-bridge = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "4b35d00" }
-numeric-id = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "4b35d00" }
+core-relations = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "1f8483" }
+egglog-bridge = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "1f8483" }
+numeric-id = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "1f8483" }
 
 # Need to add "js" feature for "graphviz-rust" to work in wasm
 getrandom = { version = "0.2.10", optional = true, features = ["js"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,9 +60,9 @@ dyn-clone = "1.0.17"
 
 # Backend
 # TODO: move egglog-backend repo into core egglog repo
-core-relations = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "1f8483" }
-egglog-bridge = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "1f8483" }
-numeric-id = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "1f8483" }
+core-relations = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "6d74e8" }
+egglog-bridge = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "6d74e8" }
+numeric-id = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "6d74e8" }
 
 # Need to add "js" feature for "graphviz-rust" to work in wasm
 getrandom = { version = "0.2.10", optional = true, features = ["js"] }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28,30 +28,21 @@ pub mod util;
 // both this crate and other crates by referring to `::egglog`.
 extern crate self as egglog;
 pub use add_primitive::add_primitive;
+use scheduler::SchedulerRecord;
 
-use crate::constraint::Problem;
-use crate::core::{AtomTerm, ResolvedAtomTerm, ResolvedCall};
-pub use crate::prelude::*;
-use crate::scheduler::SchedulerRecord;
-use crate::typechecking::TypeError;
 use ast::*;
 #[cfg(feature = "bin")]
 pub use cli::bin::*;
-use constraint::{Constraint, SimpleTypeConstraint, TypeConstraint};
+use constraint::{Constraint, Problem, SimpleTypeConstraint, TypeConstraint};
+use core::{AtomTerm, ResolvedAtomTerm, ResolvedCall};
 use core_relations::{make_external_func, ExternalFunctionId};
 pub use core_relations::{ExecutionState, Value};
 use egglog_bridge::{ColumnTy, IterationReport, QueryEntry};
 use extract::{Extractor, TreeAdditiveCostModel};
+use indexmap::map::Entry;
+use prelude::*;
 pub use serialize::{SerializeConfig, SerializedNode};
 use sort::*;
-pub use termdag::{Term, TermDag, TermId};
-use typechecking::TypeInfo;
-use util::*;
-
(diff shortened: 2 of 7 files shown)


src/scheduler.rs
```diff
@@ -0,0 +1,428 @@
...
+
+    pub fn get_match(&self, idx: usize) -> Match {
+        Match {
+            values: &self.matches[idx * self.tuple_len()..(idx + 1) * self.tuple_len()],
+            vars: &self.vars,
+        }
+    }
+
+    pub fn choose(&mut self, idx: usize) {
+        self.chosen.set(idx, true);
```
Alex-Fischman: Can we avoid the linear retain if we get rid of the bitvec and just immediately do the vec removal and table insertion here? We would want to do something with the idx so that we could use `swap_remove`.

src/scheduler.rs
```diff
@@ -0,0 +1,428 @@
...
+        report.num_matches_per_rule = action_iter_report
+            .rule_reports
+            .iter()
+            .map(|(rule, report)| (rule.as_str().into(), report.num_matches))
+            .collect();
+        report
+    }
+}
+
+fn collect_rules<'a>(
```
Alex-Fischman: Can we factor out this function and `collect_rule_ids` in `lib.rs`? They do the same thing.

src/scheduler.rs
```diff
@@ -0,0 +1,428 @@
...
+}
+
+impl Match<'_> {
+    pub fn get_value(&self, var: Symbol) -> Value {
+        let idx = self.vars.iter().position(|v| v.name == var).unwrap();
+        self.values[idx]
+    }
+}
+
+impl Matches {
```
ezrosent, line 45: Anything that's `pub` should have docs here, and maybe doc tests. (Once Alex is satisfied with the API)

ezrosent (commented):
Seems reasonable. Thanks for explaining things offline as well.

Cargo.toml
```diff
@@ -56,12 +56,17 @@ rustc-hash = "1.1"
...
 
 # Backend
 # TODO: move egglog-backend repo into core egglog repo
-core-relations = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "4236684" }
-egglog-bridge = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "4236684" }
-numeric-id = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "4236684" }
+core-relations = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "32c630" }
+egglog-bridge = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "32c630" }
+numeric-id = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "32c630" }
+# core-relations = { path = "../egglog-backend/core-relations" }
```
ezrosent: clean these up before submitting?

Alex-Fischman: Honestly I'm okay with the comments, I have the same directory structure and it's annoying to switch manually.

src/scheduler.rs
yihozhang: It is doing a slightly different thing since it's collecting ResolvedCoreRules instead of RuleIds. I put the definition inside `step_rule_with_scheduler` to be consistent with `step_rules`

yihozhang:
@Alex-Fischman @ezrosent I have refactored to use a swap_remove-like implementation and added documentation. I cannot directly call swap_remove during `choose` because it makes the index unstable. The correct way is to start at the tail and swap tuples one by one.

Cargo.toml
```diff
@@ -56,12 +56,13 @@ rustc-hash = "1.1"
...
 thiserror = "1"
 web-time = "1.1"
+dyn-clone = "1.0.17"
 
 # Backend
 # TODO: move egglog-backend repo into core egglog repo
-core-relations = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "4b35d00" }
-egglog-bridge = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "4b35d00" }
-numeric-id = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "4b35d00" }
+core-relations = { git = "https://github.com/egraphs-good/egglog-backend.git", rev = "1f8483" }
```
Alex-Fischman: This PR is blocked until the backend PR is merged.

TODO: change the commit hash once the backend PR is merged

src/lib.rs
```diff
@@ -28,20 +29,23 @@ pub mod util;
 extern crate self as egglog;
 pub use add_primitive::add_primitive;
 
+use crate::constraint::Problem;
+use crate::core::{AtomTerm, ResolvedAtomTerm, ResolvedCall};
+pub use crate::prelude::*;
+use crate::scheduler::SchedulerRecord;
+use crate::typechecking::TypeError;
```
Alex-Fischman: Why did you revert these changes? I cleaned up these imports because the `crate::` isn't doing anything. (If it is doing something we need a comment.)

src/prelude.rs
```diff
@@ -237,7 +237,7 @@ impl RustRuleContext<'_, '_> {
     }
 
     fn get_table_action(&self, table: &str) -> egglog_bridge::TableAction {
-        self.table_actions[&Symbol::from(table)]
+        self.table_actions[&Symbol::from(table)].clone()
```
Alex-Fischman, line 240: nit: `TableAction` is `Copy`

src/prelude.rs
```diff
@@ -369,7 +369,7 @@ impl<F: Fn(&mut RustRuleContext, &[Value]) -> Option<()>> Primitive for RustRule
 ///
 ///         let y = ctx.rust_to_value::<i64>(x + 2);
 ///         let f2 = ctx.rust_to_value::<i64>(f0 + f1);
-///         ctx.insert("fib", vec![y, f2]);
```
Alex-Fischman, line 372: nit: don't use `vec!`, just use an array if possible

src/prelude.rs
```diff
@@ -930,7 +930,7 @@ mod tests {
 
                 let y = ctx.rust_to_value::<i64>(x + 2);
                 let f2 = ctx.rust_to_value::<i64>(f0 + f1);
-                ctx.insert("fib", vec![y, f2]);
```
Alex-Fischman, line 933: nit: don't use vec!, just use an array if possible

src/scheduler.rs
```diff
@@ -0,0 +1,449 @@
...
+use egglog_bridge::FunctionId;
+use egglog_bridge::MergeFn;
+use egglog_bridge::RuleId;
+use egglog_bridge::TableAction;
+
+use crate::ast::ResolvedVar;
+use crate::core::GenericAtomTerm;
+use crate::core::ResolvedCoreRule;
+use crate::span;
+use crate::util::IndexMap;
```
Alex-Fischman: `util` defines `HashMap` based on the nondeterminism flag

src/scheduler.rs
```diff
@@ -0,0 +1,449 @@
...
+use crate::BackendRule;
+use crate::HashMap;
+use crate::Ruleset;
+use crate::RunReport;
+use crate::Symbol;
+
+use crate::EGraph;
```
Alex-Fischman: nit, optional: collapse common import prefixes, import `crate::*`

src/scheduler.rs
```diff
@@ -0,0 +1,449 @@
...
+                        collect_rules(*sub_ruleset, rulesets, ids);
+                    }
+                }
+            }
+        }
+
+        let mut rules = Vec::new();
+        let rulesets = std::mem::take(&mut self.rulesets);
+        collect_rules(ruleset, &rulesets, &mut rules);
+        let mut schedulers = std::mem::take(&mut self.schedulers);
```
Alex-Fischman, line 157: nit: I see no reason to use `std::mem::take` when you already have a `&mut`

src/prelude.rs
yihozhang: It was changed to Clone only because it carries a buffer after https://github.com/egraphs-good/egglog-backend/pull/38

src/scheduler.rs
yihozhang: I think this is because a mutable reference to self is used later during rule execution.

PR #588, opened 2025-06-04
fix term reconstruction for bigint

yihozhang:
oops, I thikn we need more tests for the new backend


Alex-Fischman:
How is this bug related to the new backend? I think I wrote this bug when I originally did `BigRat`...

Alex-Fischman (approved):

yihozhang:
Ah okay, I thought it's introduced in the refactoring for the new backend

FTRobbin (commented):
LGTM

PR #589, opened 2025-06-05
Performance related changes

ezrosent:
This brings this branch up to date with the latest version of the backend, and makes changes related to `TableAction` and  the primitive trait.

It also adds `mimalloc` to the benchmark suite. Open to removing it. I'd like to see a codspeed benchmark with both.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -136,6 +136,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
 
 [[package]]
+name = "cc"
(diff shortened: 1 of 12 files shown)


ezrosent:
That codspeed run was with parallelism enabled. on my machine, parallelism doesn't help much. Hoping that the next run will run things serially.

Still. It's already a big improvement over the last few runs (except for `extract-vec-bench` -- still need to figure that one out)

ezrosent:
Okay yeah I just think I don't understand what your proposal is, sorry. 

It sounds like you want to have `BaseTypeWrapper` in the backend, but I don't think that would work here without also linking in `Symbol` etc. to the backend. But I'm pretty sure that this isn't what you mean. Happy to follow up offline too.

PR #596, opened 2025-06-09
Try removing Symbol since the new backend also interns strings

Alex-Fischman:
This PR removes our dependency on `Symbol`. It is unnecessary because:
- the new backend does its own string interning
- I was able to reach performance parity in the frontend


Alex-Fischman:
Why isn't CI running?

yihozhang:
I can confirm the speedup compared to main.

On math-microbenchmark
```
main:     time:   [2.0563 s 2.0854 s 2.1165 s]
this PR:     time:   [1.8570 s 1.8834 s 1.9127 s]
```

src/ast/desugar.rs
```diff
@@ -261,27 +260,27 @@ fn desugar_birewrite(
         conditions: rewrite.conditions.clone(),
     };
     desugar_rewrite(
-        ruleset,
-        format!("{}=>", name).into(),
+        ruleset.clone(),
+        format!("{}=>", name),
```
ezrosent: nit: "{name}=>" (I know this was in the previous code too but may as well fix it. Similarly below.

ezrosent:
Nice! A few notes on why things might be faster here:

* As noted in the PR description, the new backend does its own interning. That means that the interning from `Symbol` is just adding overhead as we effectively have to unbox twice.
* `Symbol` makes hashing and comparison for strings faster but it makes everything else slower. In particular I recall going from `String` to `Symbol` (note all the "<string>".into() calls we're getting rid of) as you essentially need to do a full hashtable lookup and possible insertion under a lock.

PR #597, opened 2025-06-09
User-defined command and improvements to scheduler interface

yihozhang:
This PR follows up on #587 with the following improvements to support scheduler syntax in egglog

* Add user-defined command
* Add a `choose_all` method which is more efficient than individually choosing each match
* Add a `can_stop` method to be consistent with egg's scheduling interface.
* Use DenseIdMap for storing schedulers

### DIFF
--- a/src/ast/parse.rs
+++ b/src/ast/parse.rs
@@ -254,6 +254,7 @@ pub struct Parser {
     commands: HashMap<Symbol, Arc<dyn Macro<Vec<Command>>>>,
     actions: HashMap<Symbol, Arc<dyn Macro<Vec<Action>>>>,
     exprs: HashMap<Symbol, Arc<dyn Macro<Expr>>>,
+    user_defined: HashSet<Symbol>,
     pub symbol_gen: SymbolGen,
 }
 
@@ -263,6 +264,7 @@ impl Default for Parser {
             commands: Default::default(),
             actions: Default::default(),
             exprs: Default::default(),
+            user_defined: Default::default(),
             symbol_gen: SymbolGen::new("$".to_string()),
         }
     }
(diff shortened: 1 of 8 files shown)


src/ast/parse.rs
```diff
@@ -301,13 +303,23 @@ impl Parser {
         self.exprs.insert(ma.name(), ma);
     }
 
+    pub fn add_user_defined(&mut self, name: Symbol) {
+        self.user_defined.insert(name);
```
Alex-Fischman, line 313: TODO: return an error if the name is already taken by a command macro or by a user-defined command

src/scheduler.rs
```diff
@@ -383,18 +433,24 @@ mod test {
...
+        fn filter_matches(
+            &mut self,
+            _rule: Symbol,
+            _ruleset: Symbol,
+            matches: &mut Matches,
+        ) -> bool {
             for i in 0..std::cmp::min(self.n, matches.match_size()) {
                 matches.choose(i);
             }
+            true
```
Alex-Fischman: This scheduler can do much better than always returning `true`, right? Why doesn't it return `matches.len() <= self.n`?

Alex-Fischman (approved):

src/ast/parse.rs
yihozhang: The conflict with user-defined commands was handled by the public-facing `EGraph::add_command`

src/scheduler.rs
yihozhang: Changed to `matches.match_size() < self.n * 2` to ensure the next iter also gets enough matches

src/ast/parse.rs
Alex-Fischman: It doesn't handle command macros.

yihozhang: added check

PR #598, opened 2025-06-09
Improve `input_file` performance

FTRobbin:
This PR optimizes `input_file` performance. Instead of creating as many top-level commands as rows in the csv file, this implementation uses a single rule with a custom callback that inserts all the rows with `TableAction`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1329,23 +1329,24 @@ impl EGraph {
 
         let num_facts = parsed_contents.len();
 
-        let table_action = egglog_bridge::TableAction::new(&self.backend, func.backend_id);
+        let mut table_action = egglog_bridge::TableAction::new(&self.backend, func.backend_id);
 
-        let use_insert = function_type.subtype != FunctionSubtype::Constructor;
-
-        let write_into_db = |es, _| {
-            for row in parsed_contents.iter() {
-                if use_insert {
-                    table_action.insert(es, row.to_vec());
-                } else {
+        if function_type.subtype != FunctionSubtype::Constructor {
+            self.backend.with_execution_state(|es| {
+                for row in parsed_contents.iter() {
+                    table_action.insert(es, row.iter().copied());
+                }
+                Some(unit_val)
+            });
+        } else {
+            self.backend.with_execution_state(|es| {
+                for row in parsed_contents.iter() {
                     table_action.lookup(es, row);
                 }
-            }
-            Some(unit_val)
-        };
+                Some(unit_val)
+            });
+        }
 
-
-        self.backend.with_execution_state(write_into_db);
         self.backend.flush_updates();
 
         log::info!("Read {num_facts} facts into {func_name} from '{file}'.");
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -1230,48 +1230,99 @@ impl EGraph {
...
+
+        let unit_val = self.backend.primitives().get(());
+
         for line in contents.lines() {
-            str_buf.clear();
-            str_buf.extend(line.split('\t').map(|s| s.trim()));
-            if str_buf.is_empty() {
+            let mut row: Vec<Value> = Vec::with_capacity(row_schema.len());
+
+            for (sort, raw) in row_schema.iter().zip(line.split('\t').map(|s| s.trim())) {
```
Alex-Fischman: This will silently fail if a line is too long. You should collect the `line.split` and make sure it has the right length _before_ calling `zip`.

src/lib.rs
```diff
@@ -1230,48 +1230,99 @@ impl EGraph {
         let mut contents = String::new();
         f.read_to_string(&mut contents).unwrap();
 
-        let span: Span = span!();
-        let mut actions: Vec<Action> = vec![];
-        let mut str_buf: Vec<&str> = vec![];
+        let mut parsed_contents: Vec<Vec<Value>> = Vec::with_capacity(contents.lines().count());
```
Alex-Fischman, line 1269: Add comment about potential optimization: flatten Vec<Vec<>> into row-major Vec<>

src/lib.rs
```diff
@@ -1230,48 +1230,99 @@ impl EGraph {
...
+        }
+
+        log::debug!("Successfully loaded file.");
+
+        let num_facts = parsed_contents.len();
+
+        let table_action = egglog_bridge::TableAction::new(&self.backend, func.backend_id);
+
+        let unit_id = self.backend.primitives().get_ty::<()>();
+        let use_lookup = function_type.subtype != FunctionSubtype::Constructor;
```
Alex-Fischman: This condition seems wrong to me. It should be `lookup` for `Constructor`, `insert` for `Custom`, and I think either will work for `Relation`.

src/lib.rs
```diff
@@ -1230,48 +1230,99 @@ impl EGraph {
...
+        translator.rb.call_external_func(
+            ext_id,
+            &[],
+            egglog_bridge::ColumnTy::Primitive(unit_id),
+            || "this function will never panic".to_string(),
+        );
+
+        let id = translator.build();
+        let _ = self.backend.run_rules(&[id]).unwrap();
+        self.backend.free_rule(id);
```
Alex-Fischman: nit: https://github.com/egraphs-good/egglog-backend/pull/38 adds `with_execution_state` which is what you want here.

src/lib.rs
Alex-Fischman: I see now that it is just named incorrectly.

Alex-Fischman (changes requested):

Alex-Fischman (approved):
Switching review to approved-with-nits

PR #599, opened 2025-06-09
Tweaking `CostModel` Trait

FTRobbin:
This PR:
1. Added more documentation to the trait
2. Moved default cost functions from the sorts into the default cost model
  2.1. Changed the default cost for unstable-fn from `1 + sum of children's costs` to just `sum`
3. Made the cost model interface generic
4. Made the extractor interface generic to the cost type with `Cost: SaturatingAdd + Zero + One + Ord + Eq + Copy + Debug`

### DIFF
--- a/src/extract.rs
+++ b/src/extract.rs
@@ -4,8 +4,6 @@ use crate::util::{HashMap, HashSet};
 use crate::*;
 use std::collections::VecDeque;
 
-pub type Cost = usize;
-
 /// An interface for custom cost model
 /// For common case usage, the model should guarantee a term has a no-smaller cost
 /// than its subterms to avoid cycles in the extracted terms.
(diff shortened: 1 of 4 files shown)


src/extract.rs
```diff
@@ -4,44 +4,58 @@ use crate::util::{HashMap, HashSet};
...
-pub type Cost = usize;
-
 /// An interface for custom cost model
-/// For common case usage, the model should guarantee a term has a no-smaller cost
-/// than its subterms to avoid cycles in the extracted terms.
-/// For more niech usage, a term can have a cost less than its subterms.
+/// To use it with the default extractor, the cost type must be usize
+/// Additionally, the model should guarantee a term has a no-smaller cost
+/// than its subterms to avoid cycles in the extracted terms for common case usages.
+/// For more niech usages, a term can have a cost less than its subterms.
```
Alex-Fischman: typo: niech -> niche

src/extract.rs
```diff
@@ -4,44 +4,58 @@ use crate::util::{HashMap, HashSet};
...
+        element_costs: &[C],
+    ) -> C;
 
     /// Compute the cost of a primitive, non-container, value
-    fn leaf_primitive(&self, egraph: &EGraph, sort: &ArcSort, value: Value) -> Cost;
+    fn leaf_primitive(&self, egraph: &EGraph, sort: &ArcSort, value: Value) -> C;
 }
 
+/// default cost type
+pub type Cost = usize;
```
Alex-Fischman: nit: doc comments should be full sentences.
nit: maybe we should rename this to DefaultCost?

src/extract.rs
```diff
@@ -4,44 +4,58 @@ use crate::util::{HashMap, HashSet};
...
+        element_costs: &[Cost],
+    ) -> Cost {
+        element_costs.iter().fold(0, |s, c| s.saturating_add(*c))
+    }
+
+    /// Default cost for leaf primitives is the constant one
+    fn default_leaf_cost(&self, _egraph: &EGraph, _value: Value) -> Cost {
+        1
+    }
+}
```
Alex-Fischman: Is there a reason for these to be separate methods? If they are only called from one place I think we should inline them.

PR #601, opened 2025-06-10
Alex's tutorial revisions

Alex-Fischman:
I made a branch so I could generate a diff. I don't care if we incorporate this feedback by merging this PR or by treating each line of diff as a comment.

### DIFF
--- a/tutorial/linear-algebra-compiler/src/defn.egg
+++ b/tutorial/linear-algebra-compiler/src/defn.egg
@@ -10,9 +10,6 @@
     (MMul Matrix Matrix)
     (Transpose Matrix))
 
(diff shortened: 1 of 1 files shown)


tutorial/01-basics.egg
```diff
@@ -103,8 +100,10 @@
 ;;
 ;; 1. Define the distributivity rule for multiplication over addition. Find two expressions that 
 ;;    can be proven equivalent using this rule.
+;;        TODO: haven't we already done this above?
```
yihozhang, line 103: Oops, I think I accidentally added these rules after I make the exercise... Removed

PR #602, opened 2025-06-11
Expose message buffer, Function, and Function Row

yihozhang:


src/lib.rs
```diff
@@ -35,7 +35,9 @@ use constraint::{Constraint, Problem, SimpleTypeConstraint, TypeConstraint};
 use core::{AtomTerm, ResolvedAtomTerm, ResolvedCall};
 use core_relations::{make_external_func, ExternalFunctionId};
 pub use core_relations::{ExecutionState, Value};
+pub use egglog_bridge::FunctionRow;
 use egglog_bridge::{ColumnTy, IterationReport, QueryEntry};
+use extract::{CostModel, DefaultCost};
 use extract::{Extractor, TreeAdditiveCostModel};
```
Alex-Fischman: nit: combine these

src/lib.rs
```diff
@@ -283,14 +285,40 @@ pub trait UserDefinedCommand: Send + Sync {
...
+    pub fn name(&self) -> Symbol {
+        self.decl.name
+    }
+
+    /// Get the schema of the function.
+    pub fn schema(&self) -> &ResolvedSchema {
+        &self.schema
+    }
+
+    /// Whether the function can subsume other functions.
```
Alex-Fischman: nit: this doc is wrong

src/lib.rs
```diff
@@ -283,14 +285,40 @@ pub trait UserDefinedCommand: Send + Sync {
...
+    pub fn can_subsume(&self) -> bool {
+        self.can_subsume
+    }
+
+    /// Get the corresponding [`FunctionId`] of the function in the backend.
+    pub fn function_id(&self) -> egglog_bridge::FunctionId {
+        self.backend_id
+    }
```
Alex-Fischman: Do we need to expose this? I would rather not expose backend details

src/lib.rs
```diff
@@ -719,25 +747,40 @@ impl EGraph {
...
     /// Extract a value to a string for printing.
     /// See also `extract_value` for more control.
-    pub fn extract_value_to_string(&self, sort: &ArcSort, value: Value) -> Result<String, Error> {
-        let (termdag, term) = self.extract_value(sort, value)?;
-        Ok(termdag.to_string(&term))
+    pub fn extract_value_to_string(
+        &self,
+        sort: &ArcSort,
+        value: Value,
+    ) -> Result<(String, DefaultCost), Error> {
```
Alex-Fischman, line 776: nit: given that this is just a convenience method I don't see a reason to return the cost as well

src/lib.rs
```diff
@@ -719,25 +747,40 @@ impl EGraph {
         }
     }
 
+    /// Extract a value to a [`TermDag`] and [`Term`] in the [`TermDag`] using the default cost model.
+    pub fn extract_value(
```
Alex-Fischman, line 746: nit, optional: add "See `extract_value_with_cost_model` for more control." to docs

Alex-Fischman (commented):
Mostly nits, the main thing is to not expose backend ids if possible.

src/lib.rs
yihozhang: Updated

PR #609, opened 2025-06-12
Use `String::extend_from_within` in `TermDag::to_string`

aDotInTheVoid:
Followup to https://github.com/egraphs-good/egglog/pull/496.

This method wasn't stable at the time, but now is. Requires bumping the toolchain to 1.87.0.

Fun fact, it was stabilized specifically for this function: https://github.com/rust-lang/rust/issues/103806#issuecomment-2546776745.

### DIFF
--- a/src/termdag.rs
+++ b/src/termdag.rs
@@ -1,5 +1,5 @@
 use crate::*;
-use std::io::Write;
+use std::fmt::Write;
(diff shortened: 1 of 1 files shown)


aDotInTheVoid:
Fixed clippy.

PR #613, opened 2025-06-13
Include more literals in prelude

yihozhang:
Question: Should they be macros? I think macros will allow us to expand and inject the correct line information with `span!` instead of being in `prelude.rs`


Alex-Fischman (approved):
LGTM.

We can make new macros for them as well, or we can improve the `expr` macro to call them.

PR #615, opened 2025-06-15
Free variable exercise in section 3

Alex-Fischman:


tutorial/03-analysis.egg
```diff
@@ -195,3 +195,41 @@
 ;; `extract` can also take a second argument, which causes it to extract that many different "variants" of the
 ;; first argument. This is useful when trying to figure out why one e-class is failing to be unioned with another.
 (extract expr3 3)
+
+;; **Exercises:**
+;;
+;; 1. One analysis that is frequently useful is free variable sets. However, sometimes it is tedious and/or
+;;    slow to construct "sets" using functions. egglog provides _containers_ to solve this problem. A container
+;;    is essentially a value that can store other values; some examples are `Set`, `Map`, and `Vec`.
+;;    However, before we can construct a container, we have to tell `egglog` what sort to to use inside
```
yihozhang: to to

tutorial/03-analysis.egg
```diff
@@ -195,3 +195,41 @@
...
+;;    ```
+;;    (extract (set-of "1" "1"))
+;;    ```
+;;    We will need a function to store the results of our free variable analysis. We have to use set intersection
+;;    for the merge function, because of rewrites like `x / x => 1`.
+;;    ```
+;;    (function FreeVars (Expr) FreeVarSet :merge (set-intersect old new))
+;;    ```
+;;    Finally, we will need you to write the rules for the free variable analysis. You should have one rule for
+;;    every variant of `Expr`. We'll give you the `Add` rule for free:
```
yihozhang: Here's an example rule for `Add`:

tutorial/03-analysis.egg
```diff
@@ -195,3 +195,41 @@
 ;; `extract` can also take a second argument, which causes it to extract that many different "variants" of the
 ;; first argument. This is useful when trying to figure out why one e-class is failing to be unioned with another.
 (extract expr3 3)
+
+;; **Exercises:**
+;;
+;; 1. One analysis that is frequently useful is free variable sets. However, sometimes it is tedious and/or
+;;    slow to construct "sets" using functions. egglog provides _containers_ to solve this problem. A container
```
yihozhang: we have not mentioned how to construct sets using functions?

Issue #614, opened 2025-06-13
Comparison of kinds of High Level Rust API  (On Graph Revision)


MilkBlock:
Egraph is a compressed data structure and that means some info has been lost when you add nodes to Egraph.

For example `(let expr1 (plus A (plus B C)))` and `(let expr2 (plus D (plus B C))` share common subexpression, so they will be merged together, which means you can't distinguish between expr1's `plus B C` and expr2 `plus B C` .

If users want to update B to be M, the only way is to retype the whole tree they have to maintain their own Graph when they want to do some changes and finally parse it to Egraph.

I wrapped EGraph with a Working Area Graph which provides a suite of traits to abstract some behaviors. With such suite of traits users can quickly defines different Working Area Graph with different costs (they have to store additional data to track relations between nodes to compensate the Compression of EGraph). 

Currently they are implemented by sending Egglog String to EGraph for simplicity.

# High Level Rust API

## `new` API
```rust
fn main() {
    let node1 = Cons::new_value(3, &Cons::<MyRx>::new_end());
    let node2 = Cons::new_value(2, &node1);
    let node3 = Cons::new_value(1, &node2);
    let root = Root::new_v(&VecCon::new(vec![&node1, &node2, &node3]));
    let m = root.cur_sym();
    MyRx::rx().to_dot("egraph.dot".into());
}

basic_rx_no_vt!(MyRx);
```
## `set` API
Because EGraph is a kind of "immutable" so we need to clone part of the DAG if they have different nodes.

```rust
fn main() {
    let node1 = Cons::new_value(3, &Cons::<MyRx>::new_end());
    let mut node2 = Cons::new_value(2, &node1);
    let node3 = Cons::new_value(1, &node2);
    let root = Root::new_v(&VecCon::new(vec![&node1, &node2, &node3]));
    **node2.set_v(5);**
    MyRx::rx().to_dot("egraph.dot".into());
}
```

Check [dsl_set.rs](https://github.com/MilkBlock/egglog_wrap_test/blob/master/examples/dsl_set.rs)

## Version Control API (Commit & Stage)

we may stage some revison and  commit them to EGraph as a single version. Literally, we need a buffer between User Operations and EGraph. This API will incrementally send  Egglog String to EGraph and track WorkAreaNode versions. Mention that it will not call (pop) and (push) API of Egglog, which clones the whole EGraph and may be heavy in memory usage.

```rust 
fn main() {
[...]

Issue #583, opened 2025-05-31
Proc macro to generate egglog by rust enum/struct


MilkBlock:
I wrote a proc-macro lib by which you can easily define a suit of Predefined Egglog Type by #[egglog_ty] attribute. 

```rust
#[egglog_ty]
enum Fn {
    Builtin {
        function: BuiltinF,
    },
    WasmGuestExtern {
        name: String,
    },
}
#[egglog_ty]
enum Point {
    FixedPoint {
        offset: Offset,
    },
    OffsetPoint {
        offset: Offset,
        base: Box<Point>,
    },
    CurAnchorOf {
        object: Box<BRabject>,
    },
    PointAtIdx {
        shape: Shape,
        index: i64,
    },
}
#[egglog_ty]
struct Points { v:Vec<Point>}
```
Ref:
[https://github.com/MilkBlock/egglog_wrap_test/blob/master/crates/egglog_wrapper/src/lib.rs](https://github.com/MilkBlock/egglog_wrap_test/blob/master/crates/egglog_wrapper/src/lib.rs)

And you can simply call such generated function to generate egglog statemnt in String.

```rust
let p = Point::<Rx>::new_fixed_point(&Offset::new_d_vec2(1.0, 0.));
```
while the Rx type specifiy a Receiver (Host). It will automatically call user defined `receive` function.

All code that user need write is here.

```rust
fn main() {
    let p1 = Point::<Rx>::new_fixed_point(&Offset::new_d_vec2(1.0, 0.));
    let p2 = Point::new_fixed_point(&Offset::new_d_vec2(1.0, 2.));
    let ps = Points::new(vec![&p1,&p2]);
    Rx::singleton().to_dot(PathBuf::from("egraph"));
}
```
which also enable users to inspect the graph they generated.
[https://github.com/MilkBlock/egglog_wrap_test/blob/master/examples](https://github.com/MilkBlock/egglog_wrap_test/blob/master/examples)

I'm not sure whether this feature helpful enough.

What do you think of this? Appreciate your comments.

Alex-Fischman:
Your links seem to be broken for me.

We are currently working on making it easier to write egglog rules in Rust in #586 (https://github.com/egraphs-good/egglog/blob/adfisch-rust-api/src/prelude.rs).

Alex-Fischman:
I think I understand. You have made a macro for translating Rust types into egglog strings, and for your use case going through strings is easier than going through the Rust API.

[...]

PR #618, opened 2025-06-17
feat. proc macro & wrapper

MilkBlock:
High Level Rust API for Egglog #614 #583 , please check examples.

All traits are listed at wrap.rs. They provide both string & native way to execute commands while native is the default one.

Pull API is on the way to Extract or Inspect the EGraph.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -423,6 +423,41 @@ dependencies = [
 ]
 
 [[package]]
+name = "darling"
+version = "0.20.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fc7f46116c46ff9ab3eb1597a45688b6715c6e628b5c133e288e709a29bcb4ee"
+dependencies = [
+ "darling_core",
+ "darling_macro",
+]
+
+[[package]]
+name = "darling_core"
+version = "0.20.11"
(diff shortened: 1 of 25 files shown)


MilkBlock:
> I see there are some good ideas there, but just as a precaution, I think it's unlikely for us to merge this PR (e.g., the diff is large and egglog devs were not involved in the design). Can you package your high-level APIs in a separate crate? What APIs do we need to expose for your use cases?

You are right. I think it would be better to separate it. I would call it `eggplant`. 

I'm doing Pull API to query the result that Egglog inferred. Well, I did lots of workarounds because I didn't find some API. For example, you can see that I use Sym as an Identity of Node which is a pointer to &'static str, and build High Level  Command by cloning String.  This way is nothing with elegance and slow down performance. 

 I thought it's plausible for you because I notice your API and structs are designed for Python mostly.

It would be better to use Value to ref the node. I hope there could be some API about Value like:  

```rust
fn let(sort:&'static str, children:  Vec<Value>)-> Value{
  // define a parent of type `sort` 
}
fn extract(val:Value) -> IndexSet<(sort,Vec<Value>)>{
}
``
which is String-free.

By the way, next step I'm going to build a Rule Define API like Bevy's System, they are similar in some way and is more elegant comparing to macro.

PR #625, opened 2025-06-27
Fix BigRat Extraction

saulshanabrook:
Previously BigRat referenced the `from_string` function to construct a BigInt, but the name is actually `from-string`


yihozhang (approved):
Thank you!

PR #626, opened 2025-07-02
Union=>set desugar

### LINKED ISSUE #350: Inefficient desugaring for `rewrite`
We desugar code like `(rewrite (Add x y) (Add y x))` into something like:

```
(rule ((= e (Add x y))) ((union (Add y x) e)))
```

But  using `union` is only needed if the right-hand side of a rewrite is an identifier. When it's an expression, we could do:

```
(rule ((= e (Add x y))) ((set (Add y x) e)))
```
This should be safe, but it gives egglog less work to do: if `(Add y x)` is in the database already, setting it will call union anyway as part of the merge process. If it _isn't_ we skip creating a new noncanonical id that rebuilding has to clean up.

On the following program, this is a significant speedup:

```
(datatype Num (N i64) (Add Num Num))
; instead of (rewrite (Add a b) (Add b a))
(rule ((= e (Add a b))) ((set (Add b a) e)))
; instead of (rewrite (Add a (Add b c)) (Add (Add a b) c))
(rule ((= e (Add a (Add b c)))) ((set (Add (Add a b) c) e)))

(let _lit0 (N 0))
(let _lit1 (N 1))
(let _lit2 (N 2))
(let _lit3 (N 3))
(let _lit4 (N 4))
(let _lit5 (N 5))
(let _lit6 (N 6))
(let _lit7 (N 7))
(let _lit8 (N 8))
(let _lit9 (N 9))
(let _lit10 (N 10))
(let _lit11 (N 11))
(let _v0 (Add _lit0 _lit1))
(let _v1 (Add _lit2 _v0))
(let _v2 (Add _lit3 _v1))
(let _v3 (Add _lit4 _v2))
(let _v4 (Add _lit5 _v3))
(let _v5 (Add _lit6 _v4))
(let _v6 (Add _lit7 _v5))
(let _v7 (Add _lit8 _v6))
(let _v8 (Add _lit9 _v7))
(let _v9 (Add _lit10 _v8))
(let _v10 (Add _lit11 _v9))
(let _r0 (Add _lit11 _lit10))
(let _r1 (Add _r0 _lit9))
(let _r2 (Add _r1 _lit8))
(let _r3 (Add _r2 _lit7))
(let _r4 (Add _r3 _lit6))
(let _r5 (Add _r4 _lit5))
(let _r6 (Add _r5 _lit4))
(let _r7 (Add _r6 _lit3))
(let _r8 (Add _r7 _lit2))
(let _r9 (Add _r8 _lit1))
(let _r10 (Add _r9 _lit0))
(run 10000)
(check (= _r10 _v10))
```

This program takes 22.5 seconds to saturate on my machine. The usual rules for associativity and commutativity take 29s.

ezrosent:
Desugar `rewrite`s into calls to `set` where possible. Also bump the egglog-backend version.

* This speeds a lot of stuff up as it avoids creating a new Id that needs to be rewritten later
* The PR handles the case where such a rewrite is impossible and we need to keep `union` around.
* The current implementation of proofs needs to support minting new Ids in this case anyway. We shouldn't pessimize rules run without proofs just for proofs to work.l

Fixes #350

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -116,6 +116,27 @@ impl Display for ResolvedCall {
     }
 }
 
+/// A trait encapsulating the ability to query a [`TypeInfo`] to determine
+/// whether or not a symbol is bound as a function in the current egglog program.
+pub trait IsFunc {
+    fn is_func(&self, type_info: &TypeInfo) -> bool;
+}
+
+impl IsFunc for ResolvedCall {
+    fn is_func(&self, type_info: &TypeInfo) -> bool {
+        match self {
+            ResolvedCall::Func(func) => type_info.get_func_type(&func.name).is_some(),
+            ResolvedCall::Primitive(_) => false,
+        }
+    }
+}
+
+impl IsFunc for String {
+    fn is_func(&self, type_info: &TypeInfo) -> bool {
+        type_info.get_func_type(self).is_some()
+    }
+}
+
 #[derive(Debug, Clone)]
 pub enum GenericAtomTerm<Leaf> {
     Var(Span, Leaf),
@@ -456,7 +477,7 @@ where
 #[allow(clippy::type_complexity)]
 impl<Head, Leaf> GenericActions<Head, Leaf>
 where
-    Head: Clone + Display,
+    Head: Clone + Display + IsFunc,
     Leaf: Clone + PartialEq + Eq + Display + Hash,
 {
     pub(crate) fn to_core_actions<FG: FreshGen<Head, Leaf>>(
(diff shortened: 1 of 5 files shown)


src/core.rs
```diff
@@ -116,6 +116,27 @@ impl Display for ResolvedCall {
...
+/// A trait encapsulating the ability to query a [`TypeInfo`] to determine
+/// whether or not a symbol is bound as a function in the current egglog program.
+pub trait IsFunc {
+    fn is_func(&self, type_info: &TypeInfo) -> bool;
+}
+
+impl IsFunc for ResolvedCall {
+    fn is_func(&self, type_info: &TypeInfo) -> bool {
+        match self {
+            ResolvedCall::Func(func) => type_info.get_func_type(&func.name).is_some(),
```
yihozhang: Does this returning a false signal an error? Does it make sense to do

```
assert!(type_info.get_func_type(&func.name).is_some());
true
```

src/core.rs
```diff
@@ -542,18 +563,74 @@ where
...
-                        e2.to_core_actions(typeinfo, binding, fresh_gen, &mut norm_actions)?;
-                    norm_actions.push(GenericCoreAction::Union(
-                        span.clone(),
-                        mapped_e1.get_corresponding_var_or_lit(typeinfo),
-                        mapped_e2.get_corresponding_var_or_lit(typeinfo),
-                    ));
-                    mapped_actions
-                        .0
-                        .push(GenericAction::Union(span.clone(), mapped_e1, mapped_e2));
+                    match (e1, e2) {
```
yihozhang, line 569: Does it work both ways?
```
(GenericExpr::Var(..), GenericExpr::Call(_, f, args)) | (GenericExpr::Call(_, f, args), GenericExpr::Var(..))
                            if f.is_func(typeinfo)
```

yihozhang: Do you also need to check that `f` is a constructor? E.g., consider a function with(out) a merge function
```
(datatype E (g) (h))
(function f () E :no-merge)
(set (f) (g))


(union (f) (h))
;; is different from
;; (set (f) (h))

```

Setting and unioning have different semantics.

yihozhang (changes requested):

src/core.rs
ezrosent: Made both changes!

src/core.rs
ezrosent: No I don't think so, particularly now that we are just looking at constructors.

yihozhang (approved):
Thank you!!

PR #632, opened 2025-07-17
Add walltime benchmarks

saulshanabrook:
This PR will changes benchmarks to be measured with walltime (https://codspeed.io/docs/instruments/walltime) as well as instrumented.

The walltime should capture the effect of system calls, like with accessing files or lock contention with multithreading. However, it won't be as repeatable, so is run on the custom codspeed runner which is set up for repeatable benchmark timing.

It also:

* Switches the benchmarks to use divan over criterion. Divan is already being used by some of our other packages so it reduces the number of total dependencies we have. It is [also recommended by codspeed](https://codspeed.io/docs/benchmarks/rust/divan) and runs a lot quicker, 10 minutes versus 30 minutes on CI, even though it seems to run the same number of samples.
* Rename the benchmark file to what @oflatt recommended in a previous PR since the names change anyways with divan. 
* Runs all benchmarks with no-messages, so the CI stdout isn't clogged up running them multiple times. This also lets us move all benchmarks files to the top level so we don't have a separate `no-messages-folder.


ezrosent (approved):

saulshanabrook:
So I did some data munging to see if we could stay within the free tier of codspeed's hosted runners:
- In egglog, we ran the `build.yml` workflow [[141 times in the past 90 days](https://github.com/egraphs-good/egglog/actions/metrics/usage?dateRangeType=DATE_RANGE_TYPE_LAST_90_DAYS)](https://github.com/egraphs-good/egglog/actions/metrics/usage?dateRangeType=DATE_RANGE_TYPE_LAST_90_DAYS). That's about 47 per month. The most recent one took [10 minutes to run](https://github.com/egraphs-good/egglog/actions/runs/18333892576/job/52214142926), so in total that 470 minutes per month
- In egglog python, we ran `CI.yml` workflow 153 times in the past 90 days, averaging 51 per month. They took about [[14 minutes to run](https://github.com/egraphs-good/egglog-python/actions/runs/18329661866/job/52202075168?pr=362)](https://github.com/egraphs-good/egglog-python/actions/runs/18329661866/job/52202075168?pr=362), so in total that 714 minutes per month.


[[Codspeed increased the number of free minutes](https://codspeed.io/changelog/2025-09-08-more-free-macro-runners-minutes)](https://codspeed.io/changelog/2025-09-08-more-free-macro-runners-minutes) from 120 to 600 per month. But we can ask for more if we are an open source project that needs it. I am going to email to see if our current usage could fit within their limits. In total it looks like we use around 1200 minutes per month for the org.

ezrosent (approved):

Issue #370, opened 2024-04-25
Error when using `vec-of` in rule is not readable


gussmith23:
Error:
![image](https://github.com/egraphs-good/egglog/assets/1203329/eeb1de5c-11c9-443f-bbed-def60d59c891)

Rule:
```
(rule
 ((= ha1_s (GetOutput (ModuleInstance "HalfAdd" (vec-of "a" "b") (vec-of a b)) "sum"))
  (= ha1_c (GetOutput (ModuleInstance "HalfAdd" (vec-of "a" "b") (vec-of a b)) "carry"))
  (= fa_s (GetOutput (ModuleInstance "HalfAdd" (vec-of "a" "b") (vec-of ha1_s c)) "sum"))
  (= ha2_c (GetOutput (ModuleInstance "HalfAdd" (vec-of "a" "b") (vec-of ha1_s c)) "carry"))
  (= fa_c (Op2 (Or) ha1_c ha2_c)))
 ((union fa_s (GetOutput (ModuleInstance "FullAdd" (vec-of "a" "b" "c") (vec-of a b c)) "sum"))
  (union fa_c (GetOutput (ModuleInstance "FullAdd" (vec-of "a" "b" "c") (vec-of a b c)) "carry")))
 :ruleset adders)
```

@oflatt says this is due to the use of `vec-of`. Working around it with a cons list, but the error isn't great!

Thanks @zsisco for finding this.

yihozhang:
Fixed by #635

PR #635, opened 2025-07-21
Add grounded-ness check during lowering to core

### LINKED ISSUE #616: Check for ungrounded query variables at rule declaration time
In the old backend, ungrounded variables gave a runtime error that was at least understandable.
In the new backend, this program
```
(sort S)
(constructor S0 () S)
(constructor S1 (S) S)
(birewrite (S1 x) x)
(run 10)
```
panics witth
```
thread 'main' panicked at /home/alex/.cargo/git/checkouts/egglog-backend-94ca14661ab46e5e/a0b98d3/core-relations/src/table_spec.rs:463:36:
called `Option::unwrap()` on a `None` value
```
which is basically impossible to understand as a user. We need to check for this statically so that we can give a good error message.

yihozhang:
This PR adds a groundedness check that occurs between lowering to ResolvedCoreRule and canonicalization. It reuses the current constraint-solving mechanism for type inference and adds an `implies` constraint.

The constraint is a bit more general than what groundedness needs, but this constraint can be used for other purposes. The new constraint requires `update` to take a mutable self, so the implementation for `xor` needs to be updated to work with that.

Fixes #616, #370.


src/constraint.rs
```diff
@@ -73,6 +78,79 @@ where
...
+{
+    Box::new(Implies {
+        name,
+        watch_vars,
+        constraint: DelayedConstraint::Delayed(constraint),
+    })
+}
+
+pub type DelayedConstraintFn<Var, Value> =
+    Arc<dyn Fn(&[&Value]) -> Box<dyn Constraint<Var, Value>>>;
```
ezrosent: do we need Arc here vs. Rc? I think the frontend should be single-threaded for the most part.

src/constraint.rs
```diff
@@ -73,6 +78,79 @@ where
...
+
+impl<Var, Value> Constraint<Var, Value> for Implies<Var, Value>
+where
+    Var: cmp::Eq + PartialEq + Hash + Clone + Debug,
+    Value: Clone + Debug,
+{
+    fn update(
+        &mut self,
+        assignment: &mut Assignment<Var, Value>,
+        key: fn(&Value) -> &str,
```
ezrosent, line 144: why not `impl Fn(&Value) -> &str` or similar?

src/constraint.rs
ezrosent, line 1: I think the functions and methods in this file could use some doc comments. Particularly anything that's pub.

src/constraint.rs
```diff
@@ -73,6 +78,79 @@ where
     Box::new(Impossible { constraint })
 }
 
+pub fn implies<Var, Value>(
```
ezrosent, line 105: why are these `pub` and not `pub(crate)`?

src/constraint.rs
```diff
@@ -73,6 +78,79 @@ where
...
+    Var: cmp::Eq + PartialEq + Hash + Clone + Debug,
+    Value: Clone + Debug,
+{
+    fn update(
+        &mut self,
+        assignment: &mut Assignment<Var, Value>,
+        key: fn(&Value) -> &str,
+    ) -> Result<bool, ConstraintError<Var, Value>> {
+        let mut updated = false;
+        if let DelayedConstraint::Delayed(delayed) = &self.constraint {
```
ezrosent, line 148: can this be an exhaustive match?

also, why is the `Constraint` branch unreachable?

src/constraint.rs
```diff
@@ -961,3 +1060,68 @@ impl TypeConstraint for AllEqualTypeConstraint {
...
+                }
+            }
+        }
+    }
+
+    let _assignment = problem.solve(|_| "grounded").map_err(|err| match err {
+        ConstraintError::UnconstrainedVar(ResolvedAtomTerm::Var(span, v)) => {
+            TypeError::Ungrounded(v.to_string(), span)
+        }
+        _ => panic!("unexpected constraint error in groundedness check"),
```
ezrosent: can we debug-print the unexpected error?

ezrosent (approved):

src/constraint.rs
yihozhang: The implies constraint is useful not only for groundedness check but also type checking in general (e.g., type-level literals). We want users to use these constraints for their primitives.

src/constraint.rs
yihozhang: Interesting, we want the E-graph to be [thread safe](https://github.com/egraphs-good/egglog/pull/517), but since the constraint is not stored as part of the e-graph, Rc is fine. Fixed

src/constraint.rs
yihozhang: The existing trait interface uses `fn(&Value)->&str` and this PR respects that. I think closure is more general, but `fn(&Value) -> &str` suffices for our purposes, since the`key` function is always a getter.

src/constraint.rs
yihozhang: The first check `if let DelayedConstraint::Delayed ...` converts either exists or converts a delayed constraint to an immediate constraint, so when it reaches `let DelayedConstraint::Constraint ...` the constraint must be an immediate one.

Add a small comment to clarify.

yihozhang (commented):
@ezrosent Thanks for the review! I have made the delayed constraint an `Rc` instead of `Arc`, debug-print the error, added some documentation to the file (with help from LLM).

PR #637, opened 2025-07-21
Speedup CLI

saulshanabrook:
Speeds up CLI by using options in benchmark mode of a custom allocator and reducing the number of threads.

Locally, this decreased the time of running `time ./target/release/egglog tests/no-messages/python_array_optimize.egg --no-messages` from about 6 seconds to 3 seconds after running `cargo build --release`.

We could add more options in the future to configure these things from public API, but for now I wanted to at least speed up performance so we can better profile things in the CLI and have it be representative.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,8 @@ rustc-hash = "2.1"
 thiserror = "2.0"
 web-time = "1.1"
 dyn-clone = "1.0.17"
+rayon = "1.10.0"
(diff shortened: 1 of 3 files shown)


ezrosent (approved):

ezrosent:
Last time I added a custom allocator to egglog the webdemo broke. We may want to go back and only enable mimalloc on some targets. But supposedly mimalloc works on webassembly so we should try it!

PR #646, opened 2025-08-13
Expose whether serialization is cut off, fixes #643

### LINKED ISSUE #643: Expose whether serialization is cut off
When serializing, we should also return a boolean on whether we cut off any nodes:

https://github.com/egraphs-good/egglog/blob/main/src/serialize.rs#L90

We should also expose them on the CLI when visualizing, emitting a warning if it's cut off.

saulshanabrook:
Implement functionality to indicate when serialization exceeds specified limits, enhancing user feedback during serialization processes. Fixes #643

### DIFF
--- a/src/serialize.rs
+++ b/src/serialize.rs
@@ -87,49 +87,55 @@ impl EGraph {
     /// - Nodes will have consistant IDs throughout execution of e-graph (used for animating changes in the visualization)
     /// - Edges in the visualization will be well distributed (used for animating changes in the visualization)
     ///   (Note that this will be changed in `<https://github.com/egraphs-good/egglog/pull/158>` so that edges point to exact nodes instead of looking up the e-class)
-    pub fn serialize(&self, config: SerializeConfig) -> egraph_serialize::EGraph {
-        // First collect a list of all the calls we want to serialize
-        let all_calls: Vec<(
+    pub fn serialize(&self, config: SerializeConfig) -> (egraph_serialize::EGraph, Option<String>) {
+        let mut msg = String::new();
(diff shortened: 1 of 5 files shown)


src/serialize.rs
```diff
@@ -87,49 +87,55 @@ impl EGraph {
     /// - Nodes will have consistant IDs throughout execution of e-graph (used for animating changes in the visualization)
     /// - Edges in the visualization will be well distributed (used for animating changes in the visualization)
     ///   (Note that this will be changed in `<https://github.com/egraphs-good/egglog/pull/158>` so that edges point to exact nodes instead of looking up the e-class)
-    pub fn serialize(&self, config: SerializeConfig) -> egraph_serialize::EGraph {
-        // First collect a list of all the calls we want to serialize
-        let all_calls: Vec<(
+    pub fn serialize(&self, config: SerializeConfig) -> (egraph_serialize::EGraph, Option<String>) {
```
yihozhang: Instead of an optional string, I think it would be better to return a struct
```
enum SerializeStatus {
  egraph: egraph_serialize::EGraph,
  ignored_functions: Vec<String>,
  truncated_functions: Vec<String>
}
impl SerializeStatus {
  fn is_full(&self) -> bool;
}
```
Strings do not carry semantic meaning.

tests/integration_test.rs
```diff
@@ -748,3 +751,57 @@ fn test_shadowing_push() {
     let s = "(push) (let x 1) (pop) (let x 1)";
     EGraph::default().parse_and_run_program(None, s).unwrap();
 }
+
```
yihozhang, line 754: nice tests!

PR #647, opened 2025-08-13
raw_add_rule_with_name API & expose dependency

MilkBlock:
Currently I've built a suite of API with which you can implement constant propagation and define a DSL. Only 50 lines of code is needed! [constant_prop.rs](https://github.com/MilkBlock/eggplant/blob/main/examples/constant_prop.rs).

It depends on an unmerged API called raw_add_rule_with_name in [prelude.rs](https://github.com/MilkBlock/egglog/blob/eggplant_test_use/src/prelude.rs).

Could you please have a look and give some suggestion on which struct shouldn't I expose and which is not good?

For detailed API desgin please have a look at [intro](https://zhuanlan.zhihu.com/p/1933523964161942218).

Thank you!

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -37,14 +37,14 @@ impl<Head> HeadOrEq<Head> {
 }
 
 #[derive(Debug, Clone)]
-pub(crate) struct SpecializedPrimitive {
+pub struct SpecializedPrimitive {
     pub(crate) primitive: PrimitiveWithId,
     pub(crate) input: Vec<ArcSort>,
     pub(crate) output: ArcSort,
 }
 
 #[derive(Debug, Clone)]
-pub(crate) enum ResolvedCall {
+pub enum ResolvedCall {
     Func(FuncType),
     Primitive(SpecializedPrimitive),
 }
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -10,6 +10,9 @@ use crate::*;
 use std::any::{Any, TypeId};
 
 // Re-exports in `prelude` for convenience.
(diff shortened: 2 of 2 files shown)


MilkBlock:
@yihozhang

yihozhang:
IMO there are several ways of calling egglog:

* You can make an egglog program string and pass that to egglog. I believe what Alex meant is that for now, we have not decided to provide facilities for generating egglog strings. However, using egglog strings to interact with egglog is still a good idea, and maybe the easiest way. Eggcc and maybe program optimization project use this approach.
* You can make a series of `Command`s and pass that command to `EGraph::run_program`. It is basically an AST-level representation of the egglog program, but compared to egglog string, this is more structured. Egglog-python binding uses this approach. One downside of this approach is it cannot be expanded by macros (e.g., those in egglog-experimental).
* Finally, there are certain Rust-level APIs that don't have an egglog program/AST counterpart. Examples are user-defined primitives (`add_primitive`) and custom extraction (`extract_with_cost_model`). This also includes some functions in prelude.rs

All three ways of interacting with egglog are as safe as possible. On the other hand, low-level APIs like those requiring users to construct resolved ASTs (`EGraph::add_rule_with_name`) can be unsafe. These low-level APIs assume the program it received has been properly typechecked and contains no type or scope errors. We don't want to expose those interfaces.

For your use case, why can't you use `prelude::rust_rule`? This is the safer version of `raw_add_with_name` because it does type checking.

MilkBlock:
This sounds plausible. For engineering problem, I would keep it in my fork before next version bump of `egglog`. It's would be more easy to do revision on my fork than AST with type checking when I want to implement some new feature. I would close this PR, thank you!

PR #648, opened 2025-08-14
Remove globals in show egglog only mode

saulshanabrook:
This PR removes a conditional branch to always run the globals removal and shadow checking, even if in in show egglog only mode.

This simplifies the logic and all re-sugarring tests seem to pass fine.


yihozhang:
The relevant code was added in https://github.com/egraphs-good/egglog/pull/607 for making the resugaring test in egglog-experimental pass.

saulshanabrook:
Ah I see, thanks!

PR #650, opened 2025-08-19
Update Core Readme for Release

FTRobbin:


README.md
```diff
@@ -1,81 +1,96 @@
...
 
-Since many of the shorter running benchmarks have unstable timings due to non deterministic performance ([like in the memory allocator](https://github.com/oxc-project/backlog/issues/89)),
-we ["ignore"](https://docs.codspeed.io/features/ignoring-benchmarks/) them in codspeed. That way, we still
-capture their performance, but their timings don't show up in our reports by default.
+Move to experimental readme
 
-We use 50ms as our cutoff currently, any benchmarks shorter than that are ignored. This number was selected to try to ignore
-any benchmarks with have changes > 1% when they haven't been modified. Note that all the ignoring is done manually,
-so if you add another example that's short, an admin on the codspeed project will need to manually ignore it.
+* [@hatoo](https://github.com/hatoo) maintains an [egglog-language extension](https://marketplace.visualstudio.com/items?itemName=hatookov.egglog-language) in VS Code (just search for "egglog" in VS Code). (Outdated)
```
saulshanabrook, line 72: It might be nice to keep the community extensions here since they should work with non experimental egglog too?

README.md
```diff
@@ -89,7 +104,37 @@ samply record ./target/profiling/egglog tests/extract-vec-bench.egg
 env RUST_LOG=error samply record ./target/profiling/egglog --dont-print-messages tests/extract-vec-bench.egg
 ```
 
-# Documentation
-
-To view documentation, run `cargo doc --open`.
+**Pay Attention to Parallelism** egglog has support to run programs in parallel
```
yihozhang: @ezrosent By the way, how many threads are used if we call egglog's rust `parse_and_run_program` method? Is it also 1, or all threads?

saulshanabrook: I think it will be all threads by default. It only changes the number of threads when running from the CLI or in benchmarks.

yihozhang: This seems like a weird behavior...

yihozhang (approved):
Addressed reviews and pushed some tweaks. Left a note about parallelism in CLI vs as library

PR #651, opened 2025-08-19
Merge `egglog-backend` into `egglog`

FTRobbin:
Now, this has the directory structure that people agreed upon.


FTRobbin:
I did some research: the `Instant` crate is no longer maintained, and the `web_time::Instant` we are using is actually the best practice. Max put in the rule in `clippy.toml` 3 years ago, and things have changed.

saulshanabrook:
Sounds good to modify the `clippy.toml` then!

PR #653, opened 2025-08-20
Return structured output

saulshanabrook:
This PR changes command outputs from strings to structured data. The goal of this was to have a more unified way to process the outputs of running egglog commands programmatically and also to allow the experimental backend to [expose the extracted report](https://github.com/egraphs-good/egglog-experimental/blob/549948dd1548b51bc3a3f324b2cab1de05206f5e/src/set_cost.rs#L245-L253) in its custom extraction command. This partially addresses https://github.com/egraphs-good/egglog-experimental/issues/25 by letting us tackle it in the experimental repo.

Overall by it makes a cleaner distinction between generating outputs and printing them. I originally added the `ExtractReport` and `RunReport` as fields on the EGraph since it seemed like the least invasive way to expose them programmatically. As we are getting ready for 1.0 though it seemed also prudent to think more holistically about how to handle this case and support it more directly.

It also:

* Removes the need to store messages and most reports on the EGraph data structure. This removes some code also potentially reduces memory usage so that reports and extractions are not stored after they are returned.
* Unifies all the output modes into a `—-mode` on the CLI. These are `normal`, `resugar` (previously called the desusgarred egglog), `interactive`, and `no-messages`. This makes it clear that these options are mutually exclusive and only one can be active at a time. It also moves the logic for processes these to the CLI instead of in the core e-graph.
* Moves the logic to run re-sgar commands instead of running them to a seperate function instead of an option on the e-graph, to make it compatible with the new output type.
* Removes `set-option` because the interactive mode is moved to a CLI parameter. It also suggested it could be used in the future for enabling experimental egglog features. This was added before we had the experimental repo so I imagine that other things we wanted to add as options could be added there or as option in the Rust API in egglog.
* Fixes a bug where `(error)` was only output after the next command was run in interactive mode, instead of directly after the error.
* Only runs computation for logging if the right logging level is enabled.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -43,10 +43,6 @@ pub mod bin {
         /// Number of times to inline leaves
         #[clap(long, default_value = "0")]
         serialize_n_inline_leaves: usize,
-        /// Prevents egglog from printing messages
-        #[clap(long)]
-        no_messages: bool,
-
         #[clap(short = 'j', long, default_value = "1")]
         /// Number of threads to use for parallel execution. Passing `0` will use the maximum
         /// inferred parallelism available on the current system.


src/cli.rs
```diff
@@ -197,19 +192,86 @@ fn should_eval(curr_cmd: &str) -> bool {
...
+    Normal,
+    ShowDesugaredEgglog,
+    Interactive,
+    NoMessages,
+}
+
+impl Display for RunMode {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        // A little bit unintuitive but RunMode is specified as command-line
+        // argument with flag `--mode`, so `--mode none` means a normal run.
```
yihozhang: Do you mean `--mode normal`?

src/lib.rs
```diff
@@ -301,7 +295,7 @@ pub struct EGraph {
 /// it has an exclusive access to the e-graph.
 pub trait UserDefinedCommand: Send + Sync {
     /// Run the command with the given arguments.
-    fn update(&self, egraph: &mut EGraph, args: &[Expr]) -> Result<(), Error>;
+    fn update(&self, egraph: &mut EGraph, args: &[Expr]) -> Result<Option<CommandOutput>, Error>;
```
yihozhang, line 337: if `update` returns an optional `CommandOutput`, then we need to have a CommandOutput variant for custom user outputs.

src/lib.rs
```diff
@@ -671,135 +625,117 @@ impl EGraph {
...
-                file.write_all(buf.as_bytes())
-                    .expect("Failed to write to file");
+                file.write_all(lines.as_bytes())
+                    .expect("Error writing to file");
+                Ok(None)
             }
             None => {
-                self.print_msg(buf);
+                if is_csv {
+                    panic!("Cannot print a function to CSV if a filename is not provided")
```
yihozhang: Why not? The user might be redirecting stdout and expects output to be in the CSV format

src/lib.rs
```diff
@@ -1267,22 +1183,12 @@ impl EGraph {
...
-                        msg += "(\n";
-                        assert!(!terms.is_empty());
-                        for expr in &terms {
-                            let str = termdag.to_string(expr);
-                            log::info!("   {str}");
-                            msg += &format!("   {str}\n");
-                        }
-                        msg += ")";
-                        self.print_msg(msg);
+                    if log_enabled!(Level::Info) {
```
yihozhang, line 1172: Under the setting LOG_LEVEL=INFO and `--mode no-messages`, the new implementation would still produce a big output, unlike the old behavior, is this right?

src/lib.rs
```diff
@@ -1500,61 +1407,58 @@ impl EGraph {
...
 
-        let mut program = self.typecheck_program(&program)?;
-
-        if !show_egglog_only {
-            program = remove_globals::remove_globals(program, &mut self.parser.symbol_gen);
-
-            for command in &program {
-                self.names.check_shadowing(command)?;
-            }
+        program = remove_globals::remove_globals(program, &mut self.parser.symbol_gen);
```
yihozhang, line 1399: Is this change compatible with egglog-experimental's resugar tests?

src/lib.rs
```diff
@@ -1500,61 +1407,58 @@ impl EGraph {
...
-                if show_egglog_only {
-                    // In show_egglog mode, we still need to run scope-related commands (Push/Pop) to make
-                    // the program well-scoped.
-                    match &processed {
-                        ResolvedNCommand::Push(..) | ResolvedNCommand::Pop(..) => {
-                            self.run_command(processed.clone())?;
-                        }
-                        _ => {}
-                    };
-                    self.print_msg(processed.to_command().to_string());
```
yihozhang, line 1540: Previously, no-messages is a property of the e-graph, so run_program would respect that. With this change, there's no way for the user to turn off the messages when they call `run_program`, e.g., via methods in prelude? Is this desired?

yihozhang (changes requested):
Gave a rough first review.. There seems to be some nuanced changes to the behavior so I want to be careful here. If we want to merge this, we also need to make sure egglog-experiemental runs with this PR.

src/lib.rs
saulshanabrook: Sure, I can change this to be allowed.

src/lib.rs
saulshanabrook: I was more thinking of they wanted to return and existing command output like the new extract command would do in experimental. But sure I can also add support for custom command output with a trait.

src/cli.rs
saulshanabrook: Yes I'll fix this. Copied comment from old code and didn't update.

src/lib.rs
saulshanabrook: I believe so because it still allows you to run just the desugarring that doesn't run remove globals?

src/lib.rs
saulshanabrook: Hm, I mean now by default calling run program won't produce outputs, only return structured data, so if you don't format the results it won't ever been turned into a string. So this is similar to no messages? 

And logging in independent of course.

src/lib.rs
saulshanabrook: The old behavior did turn off some logging with no messages, and now they are entirely independent. If there is any logging that seems to verbose, we should probably change the log level or remove it?

saulshanabrook:
Thank you @yihozhang for your thorough review. I opened https://github.com/egraphs-good/egglog-experimental/pull/19 to show how these changes would work in experimental. It demonstrates how the custom extract command can return an output and have that be handled properly. It also shows how to create a custom output for the scheduler, so that other rust users or Python users who use the custom schedule could get all the run reports. I wasn't sure on the semantics there, it might also be better to just return one run report and union all the times, but I added that as an example at least of how it can work.

src/lib.rs
saulshanabrook: I opened a branch that has experimental working with this.

src/lib.rs
saulshanabrook: My intent was to keep all the messages exactly the same under this change, but the logging might change.

src/cli.rs
```diff
@@ -201,19 +196,84 @@ fn should_eval(curr_cmd: &str) -> bool {
...
+                Some(err)
+            }
+        });
+    };
+
+    Ok(match egraph.parse_and_run_program(filename, command) {
         Ok(msgs) => {
-            for msg in msgs {
-                writeln!(output, "{msg}")?;
+            if mode == RunMode::Interactive {
```
yihozhang, line 222: Currently RunMode is Interactive xor NoMessages, but is the following a legit use case: the user wants egglog to print `(done)` after each command (to confirm the command finishes), but not the entire message?

tests/integration_test.rs
```diff
@@ -752,6 +711,45 @@ fn test_shadowing_push() {
...
+    let s = "(function f () i64 :no-merge) (set (f) 2) (print-function f :mode csv)";
+    let outputs = EGraph::default().parse_and_run_program(None, s).unwrap();
+    assert_eq!(outputs[0].to_string(), "f,2\n");
+}
+
+#[test]
+fn test_print_stats() {
+    let s = "(run 1) (print-stats)";
+    let outputs = EGraph::default().parse_and_run_program(None, s).unwrap();
+    assert_eq!(
```
yihozhang, line 739: Comparing the equality of the output strings of a print-stats might be unstable. We can just check if the output has the desired variant?

tests/integration_test.rs
```diff
@@ -356,20 +341,18 @@ fn test_extract_variants1() {
              "#,
         )
         .unwrap();
-
-    let report = egraph.get_extract_report().clone().unwrap();
-    let ExtractReport::Variants { terms, .. } = report else {
-        panic!();
-    };
-    assert_eq!(terms.len(), 2);
+    assert_eq!(
```
yihozhang, line 344: Could `(extract tb 3)` produce variants in a different order? Maybe it would be better if we just check if it has the variant `ExtractVariants` with two terms?

yihozhang (approved):
Looks good to me! Left some nits. Let's merge this PR maybe after the 1.0.0 release

PR #661, opened 2025-08-24
Only install mimalloc for bin

saulshanabrook:
Currently building egglog is failing for the web demo because of an issues installing the mimalloc allocator (https://github.com/egraphs-good/egglog-demo/actions/runs/17182776958/job/48747498611)

I also had issues building mimalloc in CI when building one target of the Python bindings.

This PR changes mimalloc to only be a bin dependency, so its not installed when just installing the package.


ezrosent (approved):
Thanks!

PR #666, opened 2025-09-02
Use CSV library for writing

saulshanabrook:
Switches to using a CSV library to handle printing functions, so that things like escaping are taking care of upstream.


src/lib.rs
```diff
@@ -43,12 +44,12 @@ use log::{Level, log_enabled};
 use numeric_id::DenseIdMap;
 use prelude::*;
 use scheduler::{SchedulerId, SchedulerRecord};
-pub use serialize::{SerializeConfig, SerializedNode};
+pub use serialize::{SerializeConfig, SerializeOutput, SerializedNode};
```
saulshanabrook, line 47: Woops this unrelated change got added n this PR by accident. It exposes `SerializeOutput` publicly, otherwise there doesn't seem to be a way to access it from a third party library.

PR #670, opened 2025-09-10
Move generic ast to egglog-bridge in preparation for proofs

oflatt:


Cargo.toml
```diff
@@ -70,6 +71,7 @@ divan = "0.1.15"
 # member crates
 ######################
 egglog-core-relations = { path = "core-relations" }
+egglog-ast = { path = "egglog-ast" }
```
saulshanabrook, line 74: I notice the most of the other crates don't have the word `egglog` in their path, even if that's their name. egglog-bridge does, but that was also added before the other ones in the merge.

Not sure if it makes sense here to then make the folder just `ast`?

oflatt: I was thinking that this makes it less confusing, especially if you use it from another crate
I don't care too much tho

oflatt: it does also distinguish from ast.rs files
naming convection of egglog-* for crates?

yihozhang (changes requested):
The move looks good to me. I appreciate that the moved ASTs are `pub use`d in the `ast/mod.rs`, so this PR won't cause any interface breakage.

Looking at all the changes from OO-style function calling style to standalone function calling style (e.g., `to_canonicalized_rule` and `output_type`), IMO this should not be considered as a "strict improvement", especially since we have decided the OO-style is more preferable.  So this PR as it is right now makes some changes that are expected to be removed later. Even if later you decide more AST declarations need to be moved to `egglog-ast`, supposedly they will still use the OO convention, so these function calls still need to be reverted.

I will take a stab at using Rust traits for preserving the OO-style definitions to help merge this PR. Hopefully it will also minimize the big diffs in core.rs due to indentations.

egglog-ast/src/generic_ast.rs
```diff
@@ -0,0 +1,69 @@
...
+    Head: Clone + Display,
+    Leaf: Clone + PartialEq + Eq + Display + Hash,
+{
+    pub span: Span,
+    pub head: GenericActions<Head, Leaf>,
+    pub body: Vec<GenericFact<Head, Leaf>>,
+}
+
+#[derive(Clone, Debug, Copy, PartialEq, Eq, Hash)]
+pub enum Change {
```
yihozhang, line 96: This is declared twice here and in the main crate. The main one has documentation. We also need to make sure to `pub use Change` in the main crate if we end up deleting that one.

src/ast/mod.rs
```diff
@@ -1189,304 +1123,13 @@ pub type Action = GenericAction<String, String>;
 pub(crate) type MappedAction = GenericAction<CorrespondingVar<String, String>, String>;
 pub(crate) type ResolvedAction = GenericAction<ResolvedCall, ResolvedVar>;
 
-#[derive(Clone, Debug, PartialEq, Eq, Hash)]
```
yihozhang, line 1192: I noticed that the original documentation of GenericAction and GenericFact has been removed. Could you add them back?

oflatt: ooh good catch

egglog-ast/src/generic_ast.rs
oflatt: good catch!

Cargo.toml
saulshanabrook: @yihozhang do you have an opinion here since you just refactored things? Sorry I don't have as much context.

PR #659, opened 2025-08-24
Default to WARNING log level in CLI

saulshanabrook:
Removes explicit setting of INFO for logging and change to use default of WARN, allowing it to be overridden if desired as in https://docs.rs/env_logger/latest/env_logger/#specifying-defaults-for-environment-variables.

Closes https://github.com/egraphs-good/egglog/issues/655 by not displaying most log messages by default and also when messages are disabled.

From @rtzam

> A side note: why is the debug to log so much? I believe the default should be to only diplay the extraction results, print statements, etc. In other words, the things that are explicitly requested by the user.

Updated default behavior:

```bash
$ cargo run -- tests/web-demo/typecheck.egg
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.12s
     Running `target/debug/egglog tests/web-demo/typecheck.egg`
(TArr (TUnitConst) (TUnitConst))
(TUnitConst)
(TArr (TUnitConst) (TUnitConst))
(TArr (TUnitConst) (TArr (TArr (TUnitConst) (TUnitConst)) (TUnitConst)))
```

With logs still enabled:

```bash
$ RUST_LOG=info cargo run -- tests/web-demo/typecheck.egg
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.12s
     Running `target/debug/egglog tests/web-demo/typecheck.egg`
[INFO ] Declared sort Type.
[INFO ] Declared function TArr.
[INFO ] Declared function TUnitConst.
[INFO ] Declared function TUnit.
[INFO ] Declared sort Expr.
[INFO ] Declared function Lam.
[INFO ] Declared function App.
[INFO ] Declared function Var.
[INFO ] Declared function MyUnitConst.
[INFO ] Declared function MyUnit.
[INFO ] Declared sort Ctx.
[INFO ] Declared function Cons.
[INFO ] Declared function NilConst.
[INFO ] Declared function Nil.
[INFO ] Declared function typeof.
[INFO ] Declared rule (rewrite (typeof ctx MyUnit) TUnit).
[INFO ] Declared rule (rewrite (typeof (Cons x t ctx) (Var x)) t).
[INFO ] Declared rule (rule ((= (typeof ctx (App f e)) t2))
          ((typeof ctx f)
           (typeof ctx e))
             ).
[INFO ] Declared rule (rule ((= (typeof ctx (App f e)) t1)
           (= (typeof ctx f) (TArr (typeof ctx e) t2)))
          ((union t1 t2))
             ).
[...]

PR #672, opened 2025-09-10
Emit prompt and debug info when running from REPL

saulshanabrook:
Emits a `> ` prompt and the welcome message when running from the CLI with no file inputs and no stdin piped.

Adds a test for this mode in the repl tests.

```
$ ./target/debug/egglog
Welcome to Egglog REPL! (build: 1.0.0_2025-09-10_34c2f2de)
> (extract 10)
10
> ^C⏎
$ echo '(extract 1)' | ./target/debug/egglog
1
```

Follow up of #659 now that INFO is not printed by default.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -154,6 +154,7 @@ pub mod bin {
     }
 }
 
+#[cfg(feature = "bin")]
 impl EGraph {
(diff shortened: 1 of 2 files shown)


yihozhang (approved):
Thanks!

PR #673, opened 2025-09-11
Move name and ruleset into genericrule

oflatt:


egglog-ast/src/generic_ast_helpers.rs
```diff
@@ -28,6 +28,50 @@ macro_rules! impl_from {
     };
 }
 
+impl<Head: Display, Leaf: Display> Display for GenericRule<Head, Leaf>
```
yihozhang, line 31: Nice! Rule can finally implement the Display trait!

src/prelude.rs
```diff
@@ -226,17 +226,17 @@ pub fn rule(
...
+        name: "".into(),
+        ruleset: ruleset.into(),
     };
 
-    egraph.run_program(vec![Command::Rule {
-        name: format!("{rule:?}"),
-        ruleset: ruleset.to_owned(),
-        rule,
-    }])
+    rule.name = format!("{rule:?}");
```
yihozhang, line 237: Nits: If this rule later gets `format`ted again (e.g., printing to the user), it will print both the rule and `:name <exact copy of the rule>`, right? Maybe we can change our implementation of rule's `Display` to always not print the name?

yihozhang (approved):
Nice! I have been wanting this refactor for a long time.

Next step is maybe to move `Rewrite` to egglog_ast and embed ruleset/subsume inside Rewrite?

src/prelude.rs
oflatt: Actually that's what is already happening (:

egglog-ast/src/generic_ast_helpers.rs
```diff
@@ -28,6 +28,50 @@ macro_rules! impl_from {
...
+            } else {
+                write!(f, "{}", action)?;
+            }
+        }
+        let ruleset = if !self.ruleset.is_empty() {
+            format!(":ruleset {}", self.ruleset)
+        } else {
+            "".into()
+        };
+        let name = if !self.name.is_empty() {
```
yihozhang, line 66: Doesn't the current implementation print the name when it's not empty?

saulshanabrook:
Is this also meant to target main?

oflatt:
I was using other branches as the base temporarily for good diffs

egglog-ast/src/generic_ast_helpers.rs
oflatt: oh I see what you mean
This PR preserves the current behavior I think, let's make an issue

oflatt: https://github.com/egraphs-good/egglog/issues/684

Issue #217, opened 2023-08-24
Distinguishing pattern variables and globals


oflatt:
Right now pattern variables and globals sharing the same syntax in rules often causes confusion or bugs. 
Here's my new proposal- instead of making pattern variables have a special syntax, require that globals do. 

For example:

```
(let @add2 (Add (Var "x") (Num 2)))

(rewrite (Add @add2 (Num 2)) (Add (Var "x") (Num 4)))

```

Idk what symbol to distinguish globals with. Maybe we could use `!add2`. 
Or it should interact with modules like so:  `g.add2` from the `g` module.

saulshanabrook:
Idea: Could have it as a lint rule as well to allow these symbols in names and then warn if you don't use them

oflatt:
Our favorite proposal from the last egglog meeting is this:

Define global variables with this syntax:
```
(global-let (myglobal) 2)
```

Refer to global variables with parenthesis:
```
(rule ((= (myglobal) 2))
         ((+ (myglobal) 3)))
```

oflatt:
Implementing this proposal also helps reduce errors:
https://github.com/egraphs-good/egglog/issues/329

yihozhang:
Note that `(global-let (myglobal) 2)` is inconsistent with function definition `(function myfunction (...) ...)`, which does not have parentheses around the function to be defined. In fact I believe all current definition forms do not have parens around the name to be defined.

Another observation is that this adds a new kind of function that can be read-accessed in actions (besides EqSort). We have been doing this implementation-wise but this proposed form makes it more explicit.

oflatt:
I like the difference in syntax- it makes it clear that they are two different things. global-let is an alias for something that exists already.

I agree that it makes it more explicit that you can use it in actions. It's another reason I like this proposal

Alex-Fischman:
I think this is not a big deal with shadowing banned in #565. Reopen if you disagree.

PR #676, opened 2025-09-11
Serialize let bindings

### LINKED ISSUE #376: Display bound variables in serialized e-graph
Currently variables created with let bindings don't show up in the e-graph. Previously, they had shown as constants unified with their values.

I believe @oflatt had suggested instead treating them like metadata on e-classes. Seems like a nice idea! So we could modify the e-class to somehow show all the labels at the top. There is a place in graphviz to add a cluster label. Previously I had added types to it, so we could use that instead for labels.

We would also have to modify the serialization format to store the labels.

saulshanabrook:
This closes #376 by removing the special case where functions created by let bindings will not be serialized.

This is helpful when looking at examples/tutorial, to easily orient in the e-graph where the expressions you saved ended up.

It doesn't special case this in the visualization at all so far. For example the program `(let x 10)` would show up like:

Screenshot 2025-09-11 at 3 43 01 PM

Previously, it would be empty.

### DIFF
--- a/src/ast/mod.rs
+++ b/src/ast/mod.rs
@@ -889,9 +889,6 @@ where
     pub merge: Option<GenericExpr<Head, Leaf>>,
     pub cost: Option<DefaultCost>,
     pub unextractable: bool,
-    /// Globals are desugared to functions, with this flag set to true.
-    /// This is used by visualization to handle globals differently.
-    pub ignore_viz: bool,
     pub span: Span,
 }
 
@@ -949,7 +946,6 @@ impl FunctionDecl {
             merge,
             cost: None,
             unextractable: true,
-            ignore_viz: false,
             span,
         }
     }
@@ -969,7 +965,6 @@ impl FunctionDecl {
             merge: None,
             cost,
             unextractable,
-            ignore_viz: false,
             span,
         }
     }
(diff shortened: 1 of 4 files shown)


yihozhang:
We have had a lot of back-and-forth about global let bindings. For example,

* In the initial implementation of egglog, global variables = nullary functions
* To facilitate proofs, we make global variables not nullary functions (https://github.com/egraphs-good/egglog/pull/176, https://github.com/egraphs-good/egglog/pull/158)
* Again to facilitate, we make global variables something not nullary functions but implemented as nullary functions (https://github.com/egraphs-good/egglog/pull/338)

There are also many discussions about whether globals should have parentheses, should be shown in the visualizer, etc. I think we should avoid introducing more speculative semantics changes to global bindings.

Personally, I'm okay with the proposal in #217 to avoid global variable footguns, as long as we still allow current forms of let-bindings. We have converged to a relatively stable interface, and we should not break users' code.
```
(global (MyGlobal) (F 1))
```
or
```
(global-let (MyGlobal) (F 1))
```

saulshanabrook:
Thanks for the recap @yihozhang. I'll close this then and eventually figure out how we want to extand the realization format and the two visualizers to support this metadata.

Issue #677, opened 2025-09-13
Cache index fault in `reconstruct_term`


MilkBlock:
I implemented TableAction's tracing feature in my fork and found Proof's `Subst` field is not expected.

Given fact

<img width="160" height="230" alt="Image" src="https://github.com/user-attachments/assets/6a7fe2bd-03e6-4d53-a989-8da51d1028be" />

It infers to be 

<img width="318" height="250" alt="Image" src="https://github.com/user-attachments/assets/56a1fb35-631e-4137-9d02-6acb08b44a1b" />

When querying equal proof between term 6 and term 4, it should generate such proof

```
PSym(PRule[Equality]("AddPat", Subst {
     expr5 => ([2]Mul ([0]Const [3]BaseValueId(3) 3), ([1]Const [2]BaseValueId(3) 2))
    , expr6 => ([3]Const [4]BaseValueId(3) 4)
    , expr7 => ([4]Add ([2]Mul ([0]Const [3]BaseValueId(3) 3), ([1]Const [2]BaseValueId(3) 2)), ([3]Const [4]BaseValueId(3) 4))
    , expr6num => [4]BaseValueId(3) 4
    , expr5num => [6]BaseValueId(3) 6
    
    },...
```
But it generates

```
PSym(PRule[Equality]("AddPat", Subst {
     expr5 => ([2]Mul ([0]Const [3]3), ([1]Const [2]2))
    , expr6 => [3]3
    , expr7 => [4]4
    , expr6num => [4]4
    , expr5num => [6]6
    
    },
```
expr6 and expr7 are recognized as base value which is not expected.

This is because cache fault in `reconstruct_term` fn, in this case the `term_id 3` and `base value 3` conflicted since they have same value and cache is indexed by Value but without ColumnTy, and the solution is to change key of cache to be (Value, ColumnTy).

PR #678, opened 2025-09-13
fix cache fault in `reconstruct_term` fn

MilkBlock:
see [issue 677](https://github.com/egraphs-good/egglog/issues/677)


MilkBlock:
@oflatt

oflatt (approved):
Wow nice fix!
I'm currently refactoring proofs and this memo might go away, just a heads up

PR #680, opened 2025-09-15
Add `*Ext` traits for member functions

yihozhang:
This PR reduces the diffs of the original PR by ~200 lines.


oflatt (approved):
Oh great! This is easier than I thought, I was imagining the trait would be in the generic ast crate and you'd have import problems with sorts. 

Thanks!

PR #681, opened 2025-09-15
Support Clone for custom cost models

saulshanabrook:
I am implementing a custom cost model with [Python objects](https://docs.rs/pyo3/latest/pyo3/struct.Py.html#method.into_ptr) as the cost.

Python objects cannot be copied, so currently this would require doing something unsafe like storing a pointer as an integer and de-referencing that on lookup.

By changing the extraction algorithm to work with cloneable objects, this use case can be more easily supported.

### DIFF
--- a/src/extract.rs
+++ b/src/extract.rs
@@ -5,7 +5,7 @@ use std::collections::VecDeque;
 
 /// An interface for custom cost model.
 ///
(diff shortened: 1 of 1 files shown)


yihozhang (approved):
Thanks!

PR #682, opened 2025-09-16
remove redundant clone of exec_state

MilkBlock:
as title

### DIFF
--- a/src/sort/add_primitive/src/lib.rs
+++ b/src/sort/add_primitive/src/lib.rs
@@ -176,7 +176,7 @@ pub fn add_primitive(input: TokenStream) -> TokenStream {
                 match is_container {
                     false => quote!(exec_state.base_values().get::<#t>(#y)),
                     true => quote!(
-                        exec_state.clone().container_values().register_val::<#t>(#y, exec_state)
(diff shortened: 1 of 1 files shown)


oflatt (approved):
nice

MilkBlock:
Hello, could you give me some clue where these *.egg test file comes from ?
I observed that their naming is regular, maybe I should build a similar codegen which transform `?source` to `Rust Rule` API so I can differ the performance between `egglog` and `eggplant`.
@oflatt

saulshanabrook:
The naming of the test files is defined [here](https://github.com/egraphs-good/egglog/blob/main/tests/files.rs).

Basically, if it has `fail-typecheck` in the path it's an expected failure. If it's in `no-messages` it's run without messages on, for benchmarking purposes. Otherwise, the names are not that meaningful. The ones in `web-demo` are surfaced in the web demo, so are more tutorial like, but should act the same.


You could also open a PR merging your fork in `egglog`, not to be merged, but then the codspeed benchmarks should run automatically there and we could see any diff. It mainly runs the larger benchmarks, not the smaller ones that have a lot of variation.

PR #687, opened 2025-09-18
Make fields public in FunctionContainer for better API access

saulshanabrook:
Expose name and partially applied fields in FunctionContainer to align with other containers and support the `value_to_base` functionality in the Python API for custom cost models.

### DIFF
--- a/src/sort/fn.rs
+++ b/src/sort/fn.rs
@@ -13,7 +13,7 @@
 use super::*;
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1517,6 +1518,11 @@ impl EGraph {
         self.backend.base_values().get::<T>(x)
     }
 
+    /// Convert from an egglog value to a Rust container type.
```
yihozhang: Could you add a comment similar to what @MilkBlock suggested? The returned reference must be freed promptly to avoid deadlock.

PR #693, opened 2025-09-27
Fix rule report and variable ordering

yihozhang:
* The same description may correspond to multiple plans.
* The number of matches should only count succeeding ones.
* This PR also recovers (and slightly improves) the variable ordering used in the old backend.

### DIFF
--- a/core-relations/src/action/mod.rs
+++ b/core-relations/src/action/mod.rs
@@ -517,6 +517,7 @@ impl<'a> ExecutionState<'a> {
 }
 
 impl ExecutionState<'_> {
+    /// Returns the number of matches that make it to the end of the instructions
     pub(crate) fn run_instrs(&mut self, instrs: &[Instr], bindings: &mut Bindings) -> usize {
         if bindings.var_offsets.next_id().rep() == 0 {
             // If we have no variables, we want to run the rules once.
--- a/core-relations/src/action/mod.rs
+++ b/core-relations/src/action/mod.rs
@@ -531,7 +531,7 @@ impl ExecutionState<'_> {
             }
             self.run_instr(&mut mask, instr, bindings);
         }
-        return mask.count_ones();
+        mask.count_ones()
     }
     fn run_instr(&mut self, mask: &mut Mask, inst: &Instr, bindings: &mut Bindings) {
         fn assert_impl(
--- a/core-relations/src/action/mod.rs
+++ b/core-relations/src/action/mod.rs
@@ -517,7 +517,7 @@ impl<'a> ExecutionState<'a> {
 }
 
 impl ExecutionState<'_> {
-    pub(crate) fn run_instrs(&mut self, instrs: &[Instr], bindings: &mut Bindings) {
+    pub(crate) fn run_instrs(&mut self, instrs: &[Instr], bindings: &mut Bindings) -> usize {
         if bindings.var_offsets.next_id().rep() == 0 {
             // If we have no variables, we want to run the rules once.
             bindings.matches = 1;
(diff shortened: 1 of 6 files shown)


yihozhang:
The new variable ordering makes python-array-optimize slower:


This PR:
```
Ruleset combined_ruleset_4774957744: search 0.049s, merge 0.254s, rebuild 0.442s
Ruleset combined_ruleset_4774954096: search 0.004s, merge 0.021s, rebuild 0.056s
Ruleset combined_ruleset_4774912912: search 1.180s, merge 2.009s, rebuild 2.275s
```

Current main (`e28e4e0`):
```
Ruleset combined_ruleset_4774957744: search 0.047s, merge 0.258s, rebuild 0.446s
Ruleset combined_ruleset_4774954096: search 0.004s, merge 0.022s, rebuild 0.057s
Ruleset combined_ruleset_4774912912: search 0.558s, merge 1.959s, rebuild 2.230s
```

core-relations/src/action/mod.rs
```diff
@@ -517,7 +517,7 @@ impl<'a> ExecutionState<'a> {
 }
 
 impl ExecutionState<'_> {
-    pub(crate) fn run_instrs(&mut self, instrs: &[Instr], bindings: &mut Bindings) {
+    pub(crate) fn run_instrs(&mut self, instrs: &[Instr], bindings: &mut Bindings) -> usize {
```
ezrosent, line 521: document the return value?

core-relations/src/free_join/execute.rs
```diff
@@ -1128,28 +1131,24 @@ impl MatchCounter {
...
-            cover,
-            to_intersect,
-            ..
-        } => (
-            binding_info.subsets[cover.to_index.atom].size(),
-            -(to_intersect.len() as i32),
-        ),
+        JoinStage::Intersect { scans, .. } => scans.len() as i32,
+        JoinStage::FusedIntersect { .. } => {
+            // TODO: currently to_intersect is always empty. FusedIntersect in basically
```
ezrosent: I think this is only true for MinSize, not MinCover. IOW this is true for how we use this crate but it's possible to call core-relations directly and generate FusedIntersects like this?

Should we just do to_intersect.len() here?

core-relations/src/free_join/execute.rs
```diff
@@ -1159,8 +1158,56 @@ fn sort_plan_by_size(
     instrs: &[JoinStage],
     binding_info: &mut BindingInfo,
 ) {
-    order.data[start..]
-        .sort_by_key(|index| join_stage_order_key(&instrs[*index as usize], binding_info));
+    let mut times_refined = with_pool_set(|ps| ps.get::<DenseIdMap<AtomId, i64>>());
```
ezrosent, line 1158: nit: I think we could use an explanation for how this works. I think your explanation in the meeting was helpful, but it may be a little unclear just from reading the code.

ezrosent (approved):
Minor comments. Thanks for the fixes!

ezrosent:
I think it's worth checking with Saul on whether we should gate old vs. new ordering via a flag or something to avoid hitting python performance more, depending on how representative python-array-optimize is. I know the new backend is already a bit slower than 0.5 so pesimizing further isn't great. But up to y'all; I'm good with merging.

saulshanabrook:
Yeah I wouldn't block on the performance slowdown here. It's bad, but not like orders of magnitude worse. I think the benchmark did it's job of highlighting the issue, it's OK if we choose to move on for the sake of code clarity/ease, if there isn't an elegant way to address it right now or that would take too much effort.

core-relations/src/free_join/execute.rs
yihozhang: Used to_intersect.len() + 1 to account for `cover`

PR #694, opened 2025-09-30
Invert condition for serialized output completeness check

saulshanabrook:
Thanks to @ajpal for catching this bug. We should print a warning if the serialized output is not complete.


ajpal (approved):
thanks!

PR #698, opened 2025-10-06
Better performance debugging support

yihozhang:


saulshanabrook:
@codex review?

egglog-reports/src/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+};
+use web_time::Duration;
+
+pub(crate) type HashMap<K, V> = hashbrown::HashMap<K, V, BuildHasherDefault<FxHasher>>;
+// pub(crate) type HashSet<T> = hashbrown::HashSet<T, BuildHasherDefault<FxHasher>>;
+pub(crate) type IndexSet<T> = indexmap::IndexSet<T, BuildHasherDefault<FxHasher>>;
+// pub(crate) type IndexMap<K, V> = indexmap::IndexMap<K, V, BuildHasherDefault<FxHasher>>;
+
+#[derive(ValueEnum, Default, Serialize, Debug, Clone, Copy)]
+pub enum ReportLevel {
```
saulshanabrook, line 15: Could you add some comments here to differentiate the report levels?

egglog-reports/src/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+    #[default]
+    SizeOnly,
+    WithPlan,
+    StageInfo,
+}
```
saulshanabrook, line 23: I was also thinking it might be nice to add an option to not save any reporting info, to make it more performant. Just a fast pass that skips any of this logic that could be optionally enabled. However, this could be added in a follow up PR.

profile-viz/index.html
```diff
@@ -0,0 +1,709 @@
...
+
+    <script>
+        let profileData = null;
+        let iterationChart = null;
+        let ruleChart = null;
+        let selectedIteration = 0;
+        let selectedRule = null;
+        let showAllIterations = false;
+
+        document.getElementById('fileInput').addEventListener('change', handleFileUpload);
```
saulshanabrook, line 278: Would it be too much work to hoist out a function here that takes an element, and a JSON representation of the report, and renders the editor there?

I would love to be able to embed this in the Juptyer output in Python, passing in an explicit report and render it.

egglog-reports/src/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+use rustc_hash::FxHasher;
+use serde::Serialize;
+use std::{
+    fmt::{Display, Formatter},
+    hash::BuildHasherDefault,
+};
+use web_time::Duration;
+
+pub(crate) type HashMap<K, V> = hashbrown::HashMap<K, V, BuildHasherDefault<FxHasher>>;
+// pub(crate) type HashSet<T> = hashbrown::HashSet<T, BuildHasherDefault<FxHasher>>;
```
saulshanabrook: nit: remove comments

egglog-bridge/src/rule.rs
```diff
@@ -25,10 +25,25 @@ use crate::{
...
+            name: None,
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Variable {
+    pub id: VariableId,
+    pub name: Option<Box<str>>,
+}
```
saulshanabrook, line 45: I am not that familiar with this code. @ezrosent could you review this?

core-relations/src/free_join/execute.rs
```diff
@@ -222,10 +244,13 @@ impl Database {
...
-            let mut reservation = rule_reports.get_mut(desc).unwrap();
-            let RuleReport { num_matches, .. } = reservation.value_mut();
-            *num_matches += match_counter.read_matches(*action);
+        for (_plan, desc, _symbol_map, action) in rule_set.plans.values() {
+            let reports = rule_reports.get_mut(desc).unwrap();
+            let i = reports
+                .iter()
+                .position(|r| r.num_matches == usize::MAX)
+                .unwrap();
+            reports[i].num_matches = match_counter.read_matches(*action);
```
saulshanabrook, line 259: If you could also look at this file changes @ezrosent

saulshanabrook (approved):
Thanks @yihozhang! This looks good overall and exciting to have some more reports.

I didn't do a thorough review of a lot of the logic changes. I don't think we really have may tests to verify the output of our run reports.

Some high level tests running the EGraph and checking the report output with the three different report styles might be nice, just to make sure it wasn't doing something crazy.

ezrosent (changes requested):
What what is causing these performance issues? (submitted this early, more comments coming)

core-relations/src/free_join/execute.rs
```diff
@@ -181,16 +193,19 @@ impl Database {
...
-                        let mut rule_report: RefMut<'_, String, RuleReport> =
-                            rule_reports.entry(desc.clone()).or_default();
-                        *rule_report = rule_report.union(&RuleReport {
+                        let mut rule_report: RefMut<'_, String, Vec<RuleReport>> =
+                            dash_rule_reports.entry(desc.clone()).or_default();
+                        rule_report.value_mut().push(RuleReport {
+                            plan: report_plan,
                             search_and_apply_time,
-                            num_matches: 0,
+                            num_matches: usize::MAX,
```
ezrosent, line 201: explain why usize::MAX is being used here

core-relations/src/free_join/plan.rs
```diff
@@ -157,6 +167,88 @@ pub(crate) struct Plan {
     pub atoms: Arc<DenseIdMap<AtomId, Atom>>,
     pub stages: JoinStages,
 }
+impl Plan {
+    pub(crate) fn to_report(&self, symbol_map: &SymbolMap) -> egglog_reports::Plan {
```
ezrosent, line 171: I think this could be pretty expensive. Can we cache the output of this method somehow? Or make report generation optional?

saulshanabrook: Yeah IMHO the default should be no report generation, to give the best performance and then it can be turned on for debugging purposes.

src/lib.rs
```diff
@@ -1120,12 +977,22 @@ impl EGraph {
...
+                        self.overall_run_report.clone(),
+                    )));
+                }
+                Some(path) => {
+                    let mut file = std::fs::File::create(&path)
+                        .map_err(|e| Error::IoError(path.clone().into(), e, span.clone()))?;
+                    log::info!("Printed overall statistics to json file {}", path);
+
+                    serde_json::to_writer(&mut file, &self.overall_run_report)
+                        .expect("error serializing to json");
```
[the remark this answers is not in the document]
yihozhang: I don't think we need to return a value here.

egglog-reports/src/lib.rs
yihozhang: Added!

egglog-reports/src/lib.rs
yihozhang: Reporting the timing information should be cheap though (it's what's implemented before this PR and used in commands like `(print-stats)`)

core-relations/src/free_join/plan.rs
yihozhang: By default (at the TimeOnly level) this method won't be called. This will only be called when the user opts in for the WithPlan reporting level.

This method is also linear in the size of the outputted report... though the problem is the report tends to be large because rule texts need to be duplicated for each iteration. I can make a follow-up PR if this turns out to be unbearable.

yihozhang:
Thanks for the review! @ezrosent, @saulshanabrook

Issue #665, opened 2025-09-02
Performance issues with merge functions for containers


yihozhang:
Merge functions are often used with containers like `Set` to track things like sets of free variables or to implement aggregation (with `set-contains`). However, they are currently very slow for non-trivial workloads:

Both of the programs below require gigabytes of memory and never finish.

```
(sort IntSet (Set i64))
(relation R (i64))
(function f () IntSet :merge (set-union old new))

(R 0)
(rule 
    ((R i) (< i 50000))
    ((R (+ i 1))
     (set (f) (set-of i))))

(run-schedule
    (saturate (run)))

```

```
(sort IntSet (Set i64))
(relation R (i64))
(function f () IntSet :merge (set-union old new))

(R 0)
(ruleset init)
(rule 
    ((R i) (< i 50000))
    ((R (+ i 1)))
    :ruleset init)

(rule ((R i))
      ((set (f) (set-of i))))


(run-schedule
    (saturate (run init)
    (run)))
```

MilkBlock:
@saulshanabrook 

Hello Saul,

Currently I'm trying to Instruments egglog, would you mind teach me some knowledge about Instruments on Mac?

I use

``` cargo instruments -t "time" --release -- ./tests/eggcc-extraction.egg```

but I did't got source code available to read on Instruments.

**Comment by saulshanabrook** [comment]
Yeah you have to do it with a profile that preserves debug symbols, like:


```bash
cargo instruments -t time --profile profiling -- ./tests/no-messages/repro-665-set-merge.egg
```

Might be good to add this to the readme

PR #702, opened 2025-10-12
Immutable Containers

saulshanabrook:
Try out immutable containers to see if helps with performance for things like https://github.com/egraphs-good/egglog/issues/665


saulshanabrook:
I am going to close this for now since it doesn't speed up any of our benchmarks. It could be re-opened if there is a use case where containers are slow possibly because of the lack of re-use

PR #705, opened 2025-10-12
Don't Panic on None Primitive in Action

### LINKED ISSUE #697: Fallable in Primitives
I wanted to open an issue to discuss best practices for primitives which may error. 

For example, if you run this currently: `(extract (<< 1 1000))`, it will panic `Failed to evaluate expression '(<< 1 1000)': Panic: In 1:10-20 of web-demo.egg: (<< 1 1000): call of primitive << failed`.

This is from this code:

https://github.com/egraphs-good/egglog/blob/6160ea984fa591be92caf7a4abe081580ab477a2/src/lib.rs#L1707-L1709

I was wondering about the possibility of an alternative that wouldn't panic but would bubble up an Error. I know that things can run in parallel, but my understanding of how this would work is that if any threads raise an error, stop all other work ASAP and bubble up the error from the `run`.

I am asking this, because in the primitives that calls user defined Python code, it is very common for errors to come up. Currently, they are turned into panics, which means they don't give a nice traceback in Python:

```
  File "<doctest egglog.exp.any_expr.AnyExpr[2]>", line 1, in <module>
  File "/Users/saul/p/egg-smol-python/python/egglog/runtime.py", line 678, in _defined_method
    return fn(*args, **kwargs)
  File "/Users/saul/p/egg-smol-python/python/egglog/exp/any_expr.py", line 41, in __bytes__
    return self.egglog_any_expr_bytes().egglog_any_expr_eval()
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/saul/p/egg-smol-python/python/egglog/exp/any_expr.py", line 77, in egglog_any_expr_eval
    egraph.run(any_expr_schedule)
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/Users/saul/p/egg-smol-python/python/egglog/egraph.py", line 934, in run
    return self._run_schedule(limit_or_schedule)
           ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/Users/saul/p/egg-smol-python/python/egglog/egraph.py", line 939, in _run_schedule
    (command_output,) = self._egraph.run_program(cmd)
                        ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^
pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: PyErr { type: <class 'AttributeError'>, value: AttributeError("'bytes' object has no attribute '__bytes'"), traceback: Some("Traceback (most recent call last):\n  File \"<string>\", line 1, in <module>\n  File \"/Users/saul/p/egg-smol-python/python/egglog/exp/any_expr.py\", line 104, in <lambda>\n    yield rewrite(AnyExpr(p).egglog_any_expr_bytes()).to(AnyExpr(py_eval_fn(lambda b: b.__bytes())(p)))\n") }
```

You can see the traceback points to the run, but then the inner traceback is wrapped inside there.

There are other options I could use if I wanted to not panic, like storing a `Result` instead of a PyObject directly, and only returning the error when I try to unwrap it. But this would still keep executing, when an error is reached.

saulshanabrook:
Closes #697 by returning an error when running a ruleset that contains a primitive which returns `None` instead of panicking.

This seems like generally good practice and also will let me handle Python errors better, by re-raising them if I get an error instead of just panicking.

I was also thinking it might make sense to rename "Panic" to like "Error". This isn't a panic, it's just when a primitive returns none, which is an expected case.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -832,6 +832,8 @@ impl EGraph {
     ///
     /// This applies every match it finds (under semi-naive).
     /// See [`EGraph::step_rules_with_scheduler`] for more fine-grained control.
+    ///
+    /// This will return an error if an egglog primitive returns None in an action.
     pub fn step_rules(&mut self, ruleset: &str) -> Result<RunReport, Error> {
         fn collect_rule_ids(
             ruleset: &str,


src/lib.rs
```diff
@@ -814,25 +814,25 @@ impl EGraph {
...
-        report
+        Ok(report)
     }
 
     /// Runs a ruleset for an iteration.
     ///
     /// This applies every match it finds (under semi-naive).
     /// See [`EGraph::step_rules_with_scheduler`] for more fine-grained control.
-    pub fn step_rules(&mut self, ruleset: &str) -> RunReport {
+    pub fn step_rules(&mut self, ruleset: &str) -> Result<RunReport, Error> {
```
ezrosent, line 837: nit: explain why this method can fail (e.g. what can cause failures here, including "egglog-level panics").

PR #708, opened 2025-10-14
struct LazyMapOfIndexSet:

MilkBlock:
opt DashMap<Value,IndexSet<Value>> by allowing lazily insert when read

### DIFF
--- a/core-relations/src/containers/mod.rs
+++ b/core-relations/src/containers/mod.rs
@@ -18,6 +18,7 @@ use std::{
 use crate::numeric_id::{DenseIdMap, IdVec, NumericId, define_id};
 use crossbeam_queue::SegQueue;
 use dashmap::SharedValue;
+use hashbrown::HashSet;
 use rayon::{
     iter::{ParallelBridge, ParallelIterator},
     prelude::*,
@@ -202,15 +203,15 @@ struct ContainerEnv<C: Eq + Hash> {
     to_id: DashMap<C, Value>,
     to_container: DashMap<Value, (usize /* hash code */, usize /* map */)>,
     /// Map from a Value to the set of ids of containers that contain that value.
-    val_index: LazyMapOfIndexSet,
+    val_index: LazyContainerIdx,
 }
 #[derive(Clone)]
-struct LazyMapOfIndexSet {
-    val_index: DashMap<Value, IndexSet<Value>>,
+struct LazyContainerIdx {
+    val_index: DashMap<Value, HashSet<Value>>,
     // keys and value to insert
     // if user want to insert same value for all keys in IndexSet<Value>, LazyMap will put them
     // in pending_insert and do the insertion for single key and remove this key in pending_insert when user want to read LazyMap
-    pending_operations: Arc<Mutex<Vec<(IndexSet<Value>, InsertOrRemove)>>>,
+    pending_operations: Arc<Mutex<Vec<(HashSet<Value>, InsertOrRemove)>>>,
 }
 enum InsertOrRemove {
     Insert(Value),
@@ -218,9 +219,8 @@ enum InsertOrRemove {
 }
 
 const LAZY_BOUND: usize = 30;
-use dashmap::mapref::one::{Ref, RefMut};
-#[allow(dead_code)]
-impl LazyMapOfIndexSet {
+use dashmap::mapref::one::Ref;
+impl LazyContainerIdx {
     /// Creates a new, empty `LazyMapOfIndexSet`.
     pub fn new() -> Self {
         Self {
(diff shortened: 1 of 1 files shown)


core-relations/src/containers/mod.rs
```diff
@@ -201,7 +202,149 @@ struct ContainerEnv<C: Eq + Hash> {
...
+                .lock()
+                .unwrap()
+                .push((keys, InsertOrRemove::Remove(value)));
+        }
+    }
+
+    /// Flushes all pending lazy insertions to the underlying map.
+    fn flush_pending_operations_for_key(&self, key: &Value) {
+        let mut pending_ops = self.pending_operations.lock().unwrap();
+        if !pending_ops.is_empty() {
```
yihozhang: this if statement is redundant?

yihozhang: Currently, `pending_operations` never shrinks, even when the actual val_index is small.

We can maybe do a `pending_ops.retain(|(keys, op)| !keys.is_empty())` when we find there are too many empty sets during the enumeration

core-relations/src/containers/mod.rs
```diff
@@ -500,18 +640,17 @@ impl<C: ContainerValue> ContainerEnv<C> {
                                 self.to_container.remove(&old_val);
                                 self.to_container.insert(result, (hc as usize, target_map));
                                 *val_slot.get_mut() = result;
-                                for val in container.iter() {
-                                    let mut index = self.val_index.entry(val).or_default();
-                                    index.swap_remove(&old_val);
-                                    index.insert(result);
-                                }
+                                self.val_index
```
yihozhang, line 611: `swap_remove` is always used together with an `insert`, so maybe we can have an `update_for_all_keys` that takes both the old value and the new value. This way you don't need to materialize the container twice.

yihozhang: As another optimization, currently we always materialize the container (`container.iter().collect()`), but when the container only has a fewer entries than (`LAZY_BOUND`), the materialized container is immediately discarded. Can we make insert/remove_for_all_keys take an iterator instead of an IndexSet?

core-relations/src/containers/mod.rs
```diff
@@ -201,7 +202,149 @@ struct ContainerEnv<C: Eq + Hash> {
...
     /// Map from a Value to the set of ids of containers that contain that value.
+    val_index: LazyMapOfIndexSet,
+}
+#[derive(Clone)]
+struct LazyMapOfIndexSet {
     val_index: DashMap<Value, IndexSet<Value>>,
+    // keys and value to insert
+    // if user want to insert same value for all keys in IndexSet<Value>, LazyMap will put them
+    // in pending_insert and do the insertion for single key and remove this key in pending_insert when user want to read LazyMap
+    pending_operations: Arc<Mutex<Vec<(IndexSet<Value>, InsertOrRemove)>>>,
```
yihozhang: nit: it does not need to be an `IndexSet`, and a `HashSet` should be good enough?

core-relations/src/containers/mod.rs
```diff
@@ -201,7 +202,149 @@ struct ContainerEnv<C: Eq + Hash> {
     to_id: DashMap<C, Value>,
     to_container: DashMap<Value, (usize /* hash code */, usize /* map */)>,
     /// Map from a Value to the set of ids of containers that contain that value.
+    val_index: LazyMapOfIndexSet,
+}
+#[derive(Clone)]
+struct LazyMapOfIndexSet {
```
yihozhang: nit: consider renaming to just `LazyValIndex` or `LazyContainerIndex`?

core-relations/src/containers/mod.rs
```diff
@@ -201,7 +202,149 @@ struct ContainerEnv<C: Eq + Hash> {
...
+    pending_operations: Arc<Mutex<Vec<(IndexSet<Value>, InsertOrRemove)>>>,
+}
+enum InsertOrRemove {
+    Insert(Value),
+    Remove(Value),
+}
+
+const LAZY_BOUND: usize = 30;
+use dashmap::mapref::one::{Ref, RefMut};
+#[allow(dead_code)]
```
yihozhang: I think we should be able to remove the dead code? Unused functions can be re-implemented later easily if we want to.

core-relations/src/containers/mod.rs
MilkBlock: 2 ways to implement this.
```rust
pub trait ContainerValue: Hash + Eq + Clone + Send + Sync + 'static {
    /// Rebuild an additional container in place according the the given [`Rebuilder`].
    ///
    /// If this method returns `false` then the container must not have been modified (i.e. it must
    /// hash to the same value, and compare equal to a copy of itself before the call).
    fn rebuild_contents(&mut self, rebuilder: &dyn Rebuilder) -> bool;

    /// Iterate over the contents of the container.
    ///
    /// Note that containers can be more structured than just a sequence of values. This iterator
    /// is used to populate an index that in turn is used to speed up rebuilds. If a value in the
    /// container is eligible for a rebuild and it is not mentioned by this iterator, the outer
    /// [`Containers`] registry may skip rebuilding this container.
    fn iter(&self) -> impl Iterator<Item = Value> + '_;
}
```
1. trait revision: add `fn len()`to this trait
2. make iterator sized_iterator

Maybe we could consider it in next PR this may require many changes.

core-relations/src/containers/mod.rs
MilkBlock: Plausible, added

saulshanabrook:
> But in CLI mode, the E-graph is mem::forget, so that cost is avoided.

I opened https://github.com/egraphs-good/egglog/issues/718 to track forgetting in the benchmark as well to get more similar performance

core-relations/src/containers/mod.rs
```diff
@@ -201,7 +203,116 @@ struct ContainerEnv<C: Eq + Hash> {
...
+            self.pending_operations
+                .lock()
+                .unwrap()
+                .push((keys, InsertOrRemove::Insert(value)));
+        }
+    }
+
+    /// Lazily removes a value for all keys in the given index set.
+    pub fn remove_for_all_keys(&self, keys: HashSet<Value>, value: Value) {
+        if keys.len() < LAZY_BOUND {
```
yihozhang, line 255: Need to flush all the updates before eagerly removing?

update: I think you also need to do flush_pending_operations_for_key for eager insertion, not just eager removal

core-relations/src/containers/mod.rs
```diff
@@ -201,7 +203,116 @@ struct ContainerEnv<C: Eq + Hash> {
...
-    val_index: DashMap<Value, IndexSet<Value>>,
+    val_index: LazyContainerIdx,
+}
+#[derive(Clone)]
+struct LazyContainerIdx {
+    val_index: DashMap<Value, HashSet<Value>>,
+    // keys and value to insert
+    // if user want to insert same value for all keys in IndexSet<Value>, LazyMap will put them
+    // in pending_insert and do the insertion for single key and remove this key in pending_insert when user want to read LazyMap
+    pending_operations: Arc<Mutex<Vec<(HashSet<Value>, InsertOrRemove)>>>,
```
yihozhang, line 214: This will be very slow when many threads are contending to insert to the index

yihozhang:
I can confirm similar speedups after rebasing from main. I think it's because of the #709

core-relations/src/containers/mod.rs
yihozhang: Eli suggested crossbeam_queue for concurrent access.

core-relations/src/containers/mod.rs
yihozhang: Can we just change the interface of iter to return an [ExtractSizeIterator](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html)?

PR #709, opened 2025-10-14
"Local" PredictedVals

ezrosent:
This change includes work for the "local" predicted vals implementation that doesn't use dashmap to get consistency and instead relies on congruence closure to make things consistent later.

To get this working with proofs we introduce a whole extra union-find that is only used at proof reconstruction time to resolve inconsistent term ids. This is a bit subtle, and there's a big comment explaining what is going on there.

This change also fixes an off-by-one error in the proofs machinery based on a test that Oliver found.

### DIFF
--- a/egglog-bridge/src/lib.rs
+++ b/egglog-bridge/src/lib.rs
@@ -98,9 +98,10 @@ pub struct EGraph {
     /// uses of `(add y z)` will still see `id` within the same rule.
     ///
     /// This cache is _local_: so different threads can potentially see different values for `id`.
-    /// This is fine as all provisional values for `id` will get unioned together and the database
-    /// will be consistent after congruence closure runs. This same logic, however, does not hold
-    /// true for proofs.
+    /// In other words, each thread will have its unique id for the new row being added. This is
+    /// fine as all provisional values for `id` will get unioned together and the database will be
+    /// consistent after congruence closure runs. This same logic, however, does not hold true for
+    /// proofs.
     ///
     /// When proofs mint a new `id` for a row it is used both as the canonical e-class id in the main
     /// e-graph and also the *term* id for that specific row. Term ids are used to reconstruct
@@ -118,9 +119,9 @@ pub struct EGraph {
     /// > (add x 0) => term: id1, canon: id0
     /// > (add 0 x) => term: id2, canon: id0
     ///
-    /// So in other words, while canonical ids change over time and need not be unique to a term,
-    /// term ids are specific to the specific shape of the (head of the) term when it was first
-    /// instantiated. This is a problem for local caching, because it means that we can have
+    /// In other words, while canonical ids change over time and need not be unique to a term, term
+    /// ids are unique to the specific shape of the (head of the) term when it was first
+    /// instantiated. This is a problem for local caching because it means that we can have
     /// multiple term values for the same row. How do we pick the real one?
     ///
     /// We have a separate union-find. Duplicate insertions into the term table cause `union`s on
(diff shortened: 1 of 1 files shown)


egglog-bridge/src/lib.rs
```diff
@@ -84,6 +85,49 @@ pub struct EGraph {
...
+    /// "predicted value" to handle lookups for rows on tables that haven't been created yet. For
+    /// example, in the associativity rewrite:
+    ///
+    /// > (rewrite (add (add x y) z) (add x (add y z)))
+    ///
+    /// The RHS of the rule will check if `(add y z)` is present in the database, and if it isn't
+    /// it will create a new id and insert `(add y z) => id`. This `id` is cached so subsequent
+    /// uses of `(add y z)` will still see `id` within the same rule.
+    ///
+    /// This cache is _local_: so different threads can potentially see different values for `id`.
```
oflatt, line 100: To clarify- they see different values for `id` but there is never any overlap, right? They are still unique ids, just different.

egglog-bridge/src/lib.rs
```diff
@@ -84,6 +85,49 @@ pub struct EGraph {
...
+    /// So in other words, while canonical ids change over time and need not be unique to a term,
+    /// term ids are specific to the specific shape of the (head of the) term when it was first
+    /// instantiated. This is a problem for local caching, because it means that we can have
+    /// multiple term values for the same row. How do we pick the real one?
+    ///
+    /// We have a separate union-find. Duplicate insertions into the term table cause `union`s on
+    /// this union-find, and future lookups of this table canonicalize ids with respect to this
+    /// union-find. In this way, it's a subset of the full union-find in the `uf_table` row, only
+    /// used to resolved temporary inconsistencies in cached term values.
+    term_consistency_table: TableId,
```
oflatt, line 131: Ok, this makes sense to me and I don't know how you would do it otherwise. Parallel egglog sure makes things complicated! Afaict this is only necessary in parallel mode, right? It's fine to keep it on regardless, but could consider disabling in sequential mode.

egglog-bridge/src/lib.rs
```diff
@@ -259,19 +306,63 @@ impl EGraph {
...
+            .unwrap_or(term_id)
+    }
+
     fn term_table(&mut self, table: TableId) -> TableId {
         let spec = self.db.get_table(table).spec();
         match self.term_tables.entry(spec.n_keys) {
             Entry::Occupied(o) => *o.get(),
             Entry::Vacant(v) => {
+                let term_index = spec.n_keys + 1;
+                let term_consistency_table = self.term_consistency_table;
```
oflatt, line 338: Nit: unnecessary extra variable- I would just refer to self.term_consistency_table everywhere

egglog-bridge/src/lib.rs
ezrosent: Yup! clarified extra.

egglog-bridge/src/lib.rs
ezrosent: It should only be required in parallel, though the code right now could easily cause us to add more `ExecutionState`s than there are threads. I added a comment further down that in principle this should only be needed when parallelism is enabled.

egglog-bridge/src/lib.rs
ezrosent: We're binding it here because we `move` it in the closure further down and want to avoid moving self along with it. We're basically using this to "build the closure by hand" like in C++, but without the nice capture group syntax.

egglog-bridge/src/lib.rs
oflatt: Oh I see, okay thanks

egglog-bridge/src/lib.rs
oflatt: Ohh makes sense

ezrosent:
Good questions. We don't need consistency for two reasons:

* We only read term consistency table during proof reconstruction time; it shouldn't impact any queries or actions.
* After a full `merge` of the database, term consistency and uf should have the correct "subset" relationship where uf encodes a coarser (i.e. more equalities) equivalence relation than term consistency. But we guarantee this by "double writing" certain equalities, not by tying the two UFs together in some way.

PR #713, opened 2025-10-17
Merge1 from eggplant fork

MilkBlock:
as title

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1538,6 +1538,13 @@ impl EGraph {
         self.backend.container_values().get_val::<T>(x)
     }
 
+    /// Convert from a Rust container type to an egglog value.
+    pub fn container_to_value<T: ContainerValue>(&mut self, x: T) -> Value {
+        self.backend.with_execution_state(|state| {
+            self.backend.container_values().register_val::<T>(x, state)
+        })
+    }
+
     /// Get the size of a function in the e-graph.
     ///
     /// `panics` if the function does not exist.
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -255,10 +255,8 @@ impl RustRuleContext<'_, '_> {
     }
 
     /// Convert from an egglog value to reference of Rust container type.
-    /// requires mutable access to execution state
     ///
-    /// # warnning
-    /// return type contains a read guard, don't leak this guard or will get stuck when `rebuild`
+    /// See [`EGraph::value_to_container`].
     pub fn value_to_container<T: ContainerValue>(
         &mut self,
         x: Value,


src/lib.rs
```diff
@@ -1561,6 +1561,21 @@ impl EGraph {
...
+    /// For large tables, this is unlikely to give particularly useful output.
+    pub fn dump_debug_info(&self) {
+        self.backend.dump_debug_info();
+    }
+
+    /// Get the canonical representation for `val` based on type.
+    ///
+    /// For [`ColumnTy::Id`], it looks up the union find; otherwise,
+    /// it returns the value itself.
+    pub fn get_canon_repr(&self, val: Value, ty: ColumnTy) -> Value {
```
yihozhang: Would something like the following work for you?
```
    pub fn get_canonical_value(&self, val: Value, sort: ArcSort) -> Value {
       self.backend.get_canon_repr(val, sort.column_ty(&self.backend))
    }
```

src/prelude.rs
```diff
@@ -254,11 +254,30 @@ impl RustRuleContext<'_, '_> {
         self.exec_state.base_values().unwrap::<T>(x)
     }
 
+    /// Convert from an egglog value to reference of Rust container type.
+    /// requires mutable access to execution state
+    ///
+    /// # warnning
+    /// return type contains a read guard, don't leak this guard or will get stuck when `rebuild`
```
yihozhang:
```suggestion
    /// Convert from an egglog value to a reference of a Rust container type.
    ///
    /// See [`EGraph::value_to_container`].
```

src/sort/set.rs
```diff
@@ -3,7 +3,7 @@ use std::collections::BTreeSet;
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct SetContainer {
-    do_rebuild: bool,
+    pub do_rebuild: bool,
```
yihozhang, line 6: Just curious, what do you use this field for?

src/ast/parse.rs
```diff
@@ -201,6 +201,7 @@ impl Parser {
             || self.exprs.contains_key(&name)
             || self.commands.contains_key(&name)
         {
+            use egglog_ast::span::{RustSpan, Span};
```
yihozhang, line 204: Is this a bug fix? Could you provide more context?

src/lib.rs
MilkBlock: Thanks, I think your version is more plausible.

src/ast/parse.rs
MilkBlock: Yeah this is a bug fix. Original version embed the egglog_ast path into span macro but I can't get egglog_ast path when in eggplant.

src/sort/set.rs
MilkBlock: Origin version Container don't provide new fn and I just pub this field so that I can build struct by field init. Add a new function is a better choice. 

If my understanding is true, I can use this do_rebuild field to canonicalize the inner values of container when rebuild so that I can track eclass merging in this container in action closure.

src/prelude.rs
```diff
@@ -254,11 +254,30 @@ impl RustRuleContext<'_, '_> {
         self.exec_state.base_values().unwrap::<T>(x)
     }
 
+    /// Convert from an egglog value to reference of Rust container type.
```
saulshanabrook, line 257: Just to note this is mirrored from the egraph API: https://github.com/egraphs-good/egglog/blob/ba822a70773d56bb944d8b321495bbf5ddf59efc/src/lib.rs#L1528-L1539

src/prelude.rs
```diff
@@ -254,11 +254,30 @@ impl RustRuleContext<'_, '_> {
...
+
     /// Convert from a Rust type to an egglog value.
     pub fn base_to_value<T: BaseValue>(&self, x: T) -> Value {
         self.exec_state.base_values().get::<T>(x)
     }
 
+    /// Convert from a Rust container type to an egglog value.
+    pub fn container_to_value<T: ContainerValue>(&mut self, x: T) -> Value {
+        self.exec_state
+            .container_values()
```
saulshanabrook, line 275: If we are adding this here, I would think it would also make sense to add it to the top level API like `value_to_container` is?

MilkBlock: Quite plausible, added.

yihozhang (approved):

PR #714, opened 2025-10-17
Remove `lazy_static`

genos:
Hi! First, thanks very much for `egglog`, it's super interesting. I've had fun working with `egg` in the past, and I look forward to digging into `egglog` more!

Given that the `rust-toolchain.toml` pins the channel to `"1.87.0"`, I was planning on replacing any uses of `lazy_static` I found with `std::sync::LazyLock`, which has been in the standard library [since 1.80.0](https://doc.rust-lang.org/std/sync/struct.LazyLock.html). However, I only found one use of it in a test module to ensure that `Mutex<EGraph>` implements `Send` and `Sync`. I've replaced that "test" with a `[#test]` whose successful compilation demonstrates that `EGraph` implements `Send` while `Mutex<EGraph>` implements `Sync`. Interestingly—though perhaps only to me!—I found that `EGraph` itself does _not_ implement `Sync`.

While noodling around, I also took the liberty of appeasing `clippy`. It complained about a direct implementation of `ToString` in `benches/common.rs`: 

```
 1  warning: direct implementation of `ToString`
   --> benches/common.rs:25:1
    |
 25 | / impl ToString for BenchCase {
 26 | |     fn to_string(&self) -> String {
 27 | |         self.name.clone()
 28 | |     }
 29 | | }
    | |_^
    |
    = help: prefer implementing `Display` instead
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#to_string_trait_impl
    = note: `#[warn(clippy::to_string_trait_impl)]` on by default
```

Thanks again for this great project!

### DIFF
--- a/egglog-bridge/src/rule.rs
+++ b/egglog-bridge/src/rule.rs
@@ -94,8 +94,11 @@ impl From<ExternalFunctionId> for Function {
     }
 }
 
-trait Brc: Fn(&mut Bindings, &mut CoreRuleBuilder) -> Result<()> + dyn_clone::DynClone + Send {}
-impl<T: Fn(&mut Bindings, &mut CoreRuleBuilder) -> Result<()> + Clone + Send> Brc for T {}
+trait Brc:
+    Fn(&mut Bindings, &mut CoreRuleBuilder) -> Result<()> + dyn_clone::DynClone + Send + Sync
+{
+}
+impl<T: Fn(&mut Bindings, &mut CoreRuleBuilder) -> Result<()> + Clone + Send + Sync> Brc for T {}
 dyn_clone::clone_trait_object!(Brc);
 type BuildRuleCallback = Box<dyn Brc>;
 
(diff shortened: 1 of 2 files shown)


genos:
Apologies, I neglected `cargo fmt` and put imports in `benches/common.rs` out of alphabetical order.

yihozhang:
Oh interesting! It seems if we change the definition in rule.rs from
```
trait Brc: Fn(&mut Bindings, &mut CoreRuleBuilder) -> Result<()> + dyn_clone::DynClone + Send {}
impl<T: Fn(&mut Bindings, &mut CoreRuleBuilder) -> Result<()> + Clone + Send> Brc for T {}
```
to
```
trait Brc: Fn(&mut Bindings, &mut CoreRuleBuilder) -> Result<()> + dyn_clone::DynClone + Send + Sync {}
impl<T: Fn(&mut Bindings, &mut CoreRuleBuilder) -> Result<()> + Clone + Send + Sync> Brc for T {}
```
then EGraph is sync & send?

yihozhang (approved):
Many thanks!

PR #716, opened 2025-10-20
Clippy pedantic

genos:
**Notice**: This is a _large_ PR, and I've got one failing test, so please feel free to ignore it!

I took the liberty of running 
```
cargo clippy --all-targets -- -D warnings -D clippy::pedantic
```
on each crate individually, updating them one-per-commit.

Nearly everything is perfectly fine (there are some `#[allow(…)]` directives in here that others might wish to address, but I think this is a good first step at maximum `clippy` pedantry). However, I've currently got one failing test:

```
test scheduler::test::test_first_n_scheduler ... FAILED

failures:

---- scheduler::test::test_first_n_scheduler stdout ----

thread 'scheduler::test::test_first_n_scheduler' panicked at src/scheduler.rs:480:13:
assertion `left == right` failed
  left: [("test-rule", 20)]
 right: [("test-rule", 10)]
```

I'll leave this as a `DRAFT` for now and come back to it later; if anyone manages to figure out what I messed up in the meantime, I'd be much obliged!

_Edit_: For anyone trying to track down the bug I've introduced: one of the things that `clippy::pedantic` advises is switching conditionals from `if thing != stuff { … } else { … }` to `if thing == stuff { … } else { … }`. I imagine I've probably screwed up flipping the order of the two branches somewhere.

_Edit^2_: Looking through the different commits, the tests pass as of 1253602 when I updated the `union-find` crate. Due to interdependence between crates, I couldn't get compilation working c6bf607 (`core-relations`) until dee63f4 (`main crate`) when the test failed. So! The culprit hides somewhere among `core-relations`, `egglog-bridge (& rm once_cell)`, and `main crate`.

### DIFF
--- a/core-relations/src/base_values/mod.rs
+++ b/core-relations/src/base_values/mod.rs
@@ -113,7 +113,7 @@ impl BaseValues {
     /// Get the base value of type `P` corresponding to the given [`Value`].
     ///
     /// # Panics
-    /// if types aren't registered first.
+    /// If types aren't registered first.
     #[must_use]
     pub fn unwrap<P: BaseValue>(&self, v: Value) -> P {
         if P::MAY_UNBOX {


genos:
> Does this change the current `make nits` to run in pedantic? If not it probably should.

Great question! I didn't update CI/CD at all, since I wasn't sure of the response this PR would receive. If we want `nits` to be _maximally pedantic_, I'll happily update that as well.

genos:
Or maybe breaking it up by theme/type of pedantry, instead? As in, a series of PRs, such as:

- change `for x in stuff.iter()` to `for x in &stuff` (& variations)
- change `*` imports to explicit ones
- format strings: `"{}", x` becomes `"{x}"` (& variations)
- add `#[must_use]` before functions needing it
- document all `# Errors` and `# Panics`
- flip conditionals to have `==` be first, not `!=` (& variations)
- …

That might be a lot of PRs, though. Perhaps a single PR with commits ordered this way rather than by crate, though? 


_Edit_: Or `egglog` is just fine with the level of `clippy` it's using, and I'll happily close this PR!

genos:
This would be a large merge and, as others have noted, a rather large burden to put on other developers. What's more, `clippy pedantic` recommended switching some arguments to reference rather than value, which probably means there are semver-level breaking changes in here, too.

Happily closing this without merging! Thanks very much for your time and very interesting project.

PR #719, opened 2025-10-24
Index catalog optimized for small set of indices

yihozhang:
We currently stores indexes as a DashMap from access patterns to indices, which can be slow for several reasons:

* DashMap by default has available_parallelism() * 4 many shards, so to enumerate the DashMap requires visiting 128 shards on a 32-core desktop, even when very few indices are stored. This has been a bottleneck for workloads with many iterations.
* DashMap is optimized for concurrently accessing a large amount of data. An associative array is usually faster when data size is small (i.e < 32-64)

This PR addresses this issue by replacing the DashMap with a lightweight read-optimized array.

### DIFF
--- a/concurrency/src/lib.rs
+++ b/concurrency/src/lib.rs
@@ -43,6 +43,15 @@ pub struct ReadOptimizedLock<T> {
     data: UnsafeCell<T>,
 }
 
+impl<T: Default> Default for ReadOptimizedLock<T> {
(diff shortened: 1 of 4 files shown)


ezrosent (approved):
Very nice!! This is a cool idea and a nice speedup. I think that since these indexes are stored per-TableInfo, we're unlikely to have that many and having an O(n) map implementation is fine.

PR #722, opened 2025-10-30
Warn when globals lack the `$` prefix

oflatt:


src/typechecking.rs
```diff
@@ -705,6 +706,11 @@ pub enum TypeError {
     AllAlternativeFailed(Vec<TypeError>),
     #[error("{}\nCannot union values of sort {}", .1, .0.name())]
     NonEqsortUnion(ArcSort, Span),
+    #[error(
+        "{span}\nGlobal `{name}` must start with `{}`.",
+        crate::GLOBAL_NAME_PREFIX
+    )]
+    GlobalMissingDollar { name: String, span: Span },
```
saulshanabrook: Maybe rename to GlobalMissingPrefix to match constant?

tests/integration_test.rs
```diff
@@ -1,6 +1,53 @@
 use egglog::{extract::DefaultCost, *};
 use egglog_ast::span::{RustSpan, Span};
 
+#[test]
+fn globals_missing_prefix_warns_by_default() {
+    let _ = env_logger::builder().is_test(true).try_init();
+
+    let mut egraph = EGraph::default();
+    egraph
```
saulshanabrook, line 9: Could this test that the warning is triggered somehow?

src/util.rs
```diff
@@ -2,6 +2,17 @@ use crate::{
     ast::ResolvedVar,
     core::{ResolvedCall, SpecializedPrimitive},
 };
+use std::borrow::Cow;
+
+pub const INTERNAL_SYMBOL_PREFIX: &str = "@";
+
+pub fn sanitize_internal_name(name: &str) -> Cow<'_, str> {
```
saulshanabrook, line 11: Could you add a little doc to describe this behavior?

src/ast/check_shadowing.rs
```diff
@@ -57,44 +91,40 @@ impl Names {
     }
 
     fn check_shadowing_query(&mut self, query: &[ResolvedFact]) -> Result<(), Error> {
-        // we want to allow names in queries to shadow each other, so we first collect
-        // all of the variable names, and then we check each of those names once
-        fn get_expr_names(expr: &ResolvedExpr, inner: &mut Names) {
+        fn collect_expr_names(expr: &ResolvedExpr, out: &mut HashMap<String, Span>) {
```
saulshanabrook, line 96: What's the change here for? Is this about not allowing shadowing? If so could we add a test for this changed behavior?

src/ast/check_shadowing.rs
```diff
@@ -1,18 +1,46 @@
...
             Err(Error::Shadowing(name, old.clone(), new))
         } else {
-            self.0.insert(name, new);
+            self.seen.insert(name, new);
             Ok(())
         }
     }
 
+    fn track_global_alias(&mut self, name: &str, span: &Span) {
+        if let Some(stripped) = name.strip_prefix(GLOBAL_NAME_PREFIX) {
```
saulshanabrook, line 20: Why do we only track globals with the prefix? What if we have the warning on, then wouldn't we have globals without the prefix?

saulshanabrook (changes requested):
Thanks for this! Looks good overall, had a few small questions and suggestions. 

Also would it be possible to document in one place the different rules for variable names? Like it seems like we have global let's, local let's,  bound variables, and then also internal and external versions of all of those. It's a little unclear from me looking at the code what the prefix should be or has to be for each of those and where shadowing is or isn't allowed. 

It would also be helpful once we have that documented to have tests for all of those behaviors.

src/ast/check_shadowing.rs
oflatt: Without the prefix we can't tell if the user intended a global or not in a query variable, so we can't give an error. global_aliases tracks the ones we can error for

src/ast/check_shadowing.rs
oflatt: This is just a re-name of some variables

tests/integration_test.rs
oflatt: It prints to stderr so I'm not sure how to
Right now it just tests no error happens

src/lib.rs
```diff
@@ -349,6 +355,39 @@ impl EGraph {
...
+        }
+        if self.strict_mode {
+            Err(TypeError::GlobalMissingDollar {
+                name: name.to_owned(),
+                span: span.clone(),
+            })
+        } else {
+            if !self.warned_about_missing_global_prefix {
+                self.warned_about_missing_global_prefix = true;
+                eprintln!(
```
yihozhang: I thought we have banned eprintln in clippy.toml. Can we use log::warn! instead?

src/ast/check_shadowing.rs
yihozhang: Is the deletion of the comment unintentional?

yihozhang (commented):
nit: could you also make sure rules like
```
;; when no global named $x is defined
(rule ((= $x 1)) ())
```
and
```
(rule () (let $y 1))
```
throw a warning?

src/lib.rs
oflatt: Edit: nvm, I think log::warn! does show up by default

src/ast/check_shadowing.rs
oflatt: Yeah I'll restore it

src/lib.rs
yihozhang: I think we have set the default logging level to warn, so it should show up
```
env_logger::Builder::from_env(Env::default().default_filter_or("warn"))
```

tests/integration_test.rs
saulshanabrook: Ah yeah, maybe if you use warm then you can verify that it warns in the test?

egglog-ast/src/generic_ast_helpers.rs
```diff
@@ -215,6 +221,32 @@ where
     Head: Clone + Display,
     Leaf: Clone + Eq + Display + Hash,
 {
+    pub fn visit_vars(&self, f: &mut impl FnMut(&Span, &Leaf)) {
```
yihozhang, line 224: Can you just implement `visit_vars` in terms of `visit_exprs`?

```rust
pub fn visit_vars(self, f: &mut impl FnMut(Span, Leaf) -> (Span, Leaf)) -> Self {
    self.visit_exprs(|e| if let Var(s, v) =  e { let (s, v) = f(s, v); Var(s, v) } else { e })
}
```

saulshanabrook:
Could you add some documentation somewhere to describe what prefixes are recommended and which are not allowed for global lets, local lets, and variables for users?

src/ast/check_shadowing.rs
saulshanabrook, line 20: Do we need to strip the name for the global aliases?

What if you set `(let $x …)` and `(let x …)` will these conflict? I would assume these would be different variables?

oflatt: I want these to conflict to avoid footguns

saulshanabrook (approved):

PR #739, opened 2025-11-15
Fix Incorrect Unstable Function Behavior

### LINKED ISSUE #738: UnstableFn Bug Partial Different Sorts
This program panics:

```
(sort X)
(sort Y)
(sort Y_to_X (UnstableFn (Y) X))

(constructor x () X)
(constructor fn (X Y) X)
(constructor wrap (Y_to_X) X)

(extract (unstable-fn "fn" (x)))
```

Also, if I look at the visualization for this modified program I can see that the partially function value is an empty e-class:

```
(sort X)
(sort Y)
(sort Y_to_X (UnstableFn (Y) X))

(constructor x () X)
(constructor fn (X Y) X)
(constructor wrap (Y_to_X) X)

(wrap (unstable-fn "fn" (x)))
```

Image

Interestingly, if I remove the Y sort and change all references to X this doesn't happen and I get the correct result:

```
(sort X)
(sort X_to_X (UnstableFn (X) X))

(constructor x () X)
(constructor fn (X X) X)
(constructor wrap (X_to_X) X)

(wrap (unstable-fn "fn" (x)))
```

Image

saulshanabrook:
Closes #738 by fixing incorrect behavior of the unstable higher order functions. Previously, it assumed that the sorts of the partially applied args were the same as the sort of the remaining args, or the sorts that the function sort was parametrized over. Of course, they may be different. This fixes the bug by storing the arcsorts on the function container value.

However, the extractor currently needs to ask each sort for all of it's possible children sorts when deciding which sorts to extract from. So we also have to store a list of all the arcsorts used in any of the partial functions on the sort. Alternatively, we could change the extractor to not require to know from each sort all possible child sorts, but for now I have opted to keep the interface the same and only make changes to the function sort.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1533,11 +1533,11 @@ impl<'a> BackendRule<'a> {
             } else {
                 panic!("no callable for {name}");
             };
-            let do_rebuild = prim.input.iter().skip(1).cloned().collect();
+            let partial_arcsorts = prim.input.iter().skip(1).cloned().collect();
 
             qe_args[0] = self.rb.egraph().base_value_constant(ResolvedFunction {
                 id,
-                partial_arcsorts: do_rebuild,
+                partial_arcsorts,
                 name: name.clone(),
             });
         }


oflatt (approved):
nice catch

PR #741, opened 2025-11-19
Enable type-aware macros for fresh! sugar

oflatt:
Enables this PR:
https://github.com/egraphs-good/egglog-experimental/pull/34/files#diff-863d9a2b76bfdb514c1569041f1f009c0625e89aee367305d42aaad6405f5737


src/lib.rs
```diff
@@ -3,14 +3,63 @@
...
+//! introduces a constructor (name generated via the parser's symbol generator):
+//!
+//! ```text
+//! (constructor GeneratedFreshTableN (<cols...> i64) <Sort>)
+//! ```
+//!
+//! where:
+//! - `<cols...>` are the types of all subexpressions appearing in the rule's
+//!   query (LHS). These are obtained by typechecking the query.
+//! - The trailing `i64` is a per-`fresh!`-call unique index within the rule,
```
saulshanabrook: Could this just be added to the constructor name?

src/lib.rs
```diff
@@ -3,14 +3,63 @@
...
+//! - `fresh!` is a surface sugar and does not appear in the core AST.
+//!
+//! #### Warning and Tips
+//! - Since the generated constructor is a pure function of the query subexpressions
+//!   (plus the per-call index), two different rule matches that later become
+//!   equivalent (e.g., via `union` effects in the e-graph) will produce equal
+//!   arguments to the constructor and therefore equal "fresh" results. In other
+//!   words, "fresh" is only as unique as the tuple of its query arguments for a
+//!   given `fresh!` site.
+//! - This behavior can be useful when you want to intentionally create cycles in
```
saulshanabrook: Do we have a test that uses it to produce a cycle? From a first look at the tests it seems like they don't do this but something else? I am a little confused why you would want this except to introduce cycles that are able to be unioned together.

src/lib.rs
oflatt: Yes but would create a lot of noise with potentially many constructors.

src/lib.rs
oflatt: done

src/ast/desugar.rs
```diff
@@ -185,6 +192,199 @@ fn desugar_datatype(span: Span, name: String, variants: Vec<Variant>) -> Vec<NCo
...
+        .map(|(_, sort)| sort.name().to_string())
+        .collect();
+    constructor_inputs.push("i64".to_string());
+
+    let mut table_names: IndexMap<String, String> = IndexMap::default();
+    let mut constructor_decls: Vec<NCommand> = Vec::new();
+
+    for (sort_name, _sort) in fresh_sorts.iter() {
+        if !table_names.contains_key(sort_name) {
+            let fresh_table_name = parser.symbol_gen.fresh("GeneratedFreshTable");
```
saulshanabrook: I'm looking at the code now and wondering how hard it would be to create a new kind of macro to be able to support this in experimental... Not only would reduce the complexity in main but maybe there would end up being other uses for having type aware macros?

It seems like we might not have to expose the normalized commands too? And typeinfo is already exposed.

src/ast/desugar.rs
```diff
@@ -185,6 +167,199 @@ fn desugar_datatype(span: Span, name: String, variants: Vec<Variant>) -> Vec<NCo
...
```
yihozhang: Does this need to have an internal function prefix?

src/ast/desugar.rs
```diff
@@ -1,29 +1,13 @@
...
-
-/// Desugars a single command into the normalized form.
-/// Gets rid of a bunch of syntactic sugar, but also
-/// makes rules into a SSA-like format (see [`NormFact`]).
+/// Desugars a single command, removing syntactic sugar.
+/// The `type_info` contains type infromation for already desugared parts of the program.
 pub(crate) fn desugar_command(
     command: Command,
     parser: &mut Parser,
-    seminaive_transform: bool,
```
yihozhang, line 26: Good catch

src/lib.rs
```diff
@@ -1333,46 +1357,85 @@ impl EGraph {
...
-                    outputs.push(output);
+        let mut desugared_commands = Vec::new();
+        let mut program_queue: Vec<Command> = program;
+        program_queue.reverse();
+
+        while let Some(before_expanded_command) = program_queue.pop() {
+            // First do user-provided macro expansion for this command,
+            // which may rely on type information from previous commands.
+            let macro_expanded = self.command_macros.apply(
+                before_expanded_command.clone(),
```
yihozhang: Is the clone unnecessary?

src/lib.rs
```diff
@@ -1333,46 +1357,85 @@ impl EGraph {
...
+        let desugared = desugar_command(command, &mut self.parser)?;
+        Ok(self.typecheck_program(&desugared)?)
     }
 
-    /// Run a program, represented as an AST.
-    /// Return a list of messages.
-    pub fn run_program(&mut self, program: Vec<Command>) -> Result<Vec<CommandOutput>, Error> {
+    /// Run a program, returning the desugared outputs as well as the CommandOutputs.
+    /// Can optionally not run the commands, just adding type information.
+    fn run_program_internal(
```
yihozhang: This program is called `run_program`, but it can optionally not run the program, which feels a bit weird.. Why can't it just be two functions, one for running and one for desugaring?

yihozhang: It also feels uncommon to me that `desugar` returns a typechecked program. It should just return a program that is `CommandMacro`-desugared, followed by `desugar.rs`-desugared?

tests/test_command_macros.rs
```diff
@@ -0,0 +1,355 @@
...
+        _type_info: &TypeInfo,
+    ) -> Result<Vec<Command>, Error> {
+        match command {
+            Command::Rule { rule } => {
+                let rule_name = rule.name.clone();
+                Ok(vec![
+                    Command::Rule { rule },
+                    // Comments don't exist in the AST but we can use PrintSize with None
+                    // as a marker that the macro ran
+                    Command::PrintSize(span!(), Some(format!("rule_{}", rule_name))),
```
yihozhang: Consider rewriting this example macro? It does not do what its name suggests

yihozhang (changes requested):
Looks good to me! I have a question about the interface of `run_program_internal` and `desugar_program`.

egglog-ast/src/generic_ast_helpers.rs
```diff
@@ -104,16 +108,22 @@ where
...
-                rhs
-            ),
+            GenericAction::Set(_, lhs, args, rhs) => {
+                if args.is_empty() {
+                    write!(f, "(set ({}) {})", lhs, rhs)
+                } else {
+                    write!(
+                        f,
+                        "(set ({} {}) {})",
+                        lhs,
```
saulshanabrook, line 118: This change was to remove the extra space for displaying a set without args?

src/ast/desugar.rs
```diff
@@ -1,29 +1,13 @@
...
-        res.extend(desugared);
-    }
-    Ok(res)
-}
-
-/// Desugars a single command into the normalized form.
-/// Gets rid of a bunch of syntactic sugar, but also
-/// makes rules into a SSA-like format (see [`NormFact`]).
+/// Desugars a single command, removing syntactic sugar.
+/// The `type_info` contains type infromation for already desugared parts of the program.
```
saulshanabrook: Is this comment still accurate? I don't see a `type_info`.

src/cli.rs
```diff
@@ -286,7 +286,7 @@ impl Display for RunMode {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         match self {
             RunMode::Normal => write!(f, "normal"),
-            RunMode::ShowDesugaredEgglog => write!(f, "resugar"),
+            RunMode::ShowDesugaredEgglog => write!(f, "desugar"),
```
saulshanabrook, line 289: So this changes the CLI to be able to show de-sugared instead of re-sugarred commands… This seems fine, but I am curious why this was necessary?

src/ast/mod.rs
```diff
@@ -1047,7 +1048,7 @@ where
 }
 
 pub type Fact = GenericFact<String, String>;
-pub(crate) type ResolvedFact = GenericFact<ResolvedCall, ResolvedVar>;
+pub type ResolvedFact = GenericFact<ResolvedCall, ResolvedVar>;
 pub(crate) type MappedFact<Head, Leaf> = GenericFact<CorrespondingVar<Head, Leaf>, Leaf>;
```
saulshanabrook, line 1052: Same Q for these two made public, is that required?

src/core.rs
```diff
@@ -40,14 +40,36 @@ impl<Head> HeadOrEq<Head> {
...
+
+    /// Get the input sorts of this primitive
+    pub fn input(&self) -> &[ArcSort] {
+        &self.input
+    }
+
+    /// Get the external function ID of this primitive
+    pub(crate) fn external_id(&self) -> ExternalFunctionId {
+        self.primitive.1
+    }
```
saulshanabrook, line 68: Not that it really matters, but it seems a bit simpler to make these all pub and then make the external ID on the external function id `pub(crate)` to hide it?

src/core.rs
```diff
@@ -40,14 +40,36 @@ impl<Head> HeadOrEq<Head> {
...
 }
 
 #[derive(Debug, Clone)]
-pub(crate) enum ResolvedCall {
+pub enum ResolvedCall {
```
saulshanabrook, line 72: Same Q with this, if it needs to be exposed

src/lib.rs
```diff
@@ -1333,46 +1357,85 @@ impl EGraph {
...
-            // because push and pop create new scopes
-            for processed in self.process_command(command)? {
-                let result = self.run_command(processed)?;
-                if let Some(output) = result {
-                    outputs.push(output);
+        let mut desugared_commands = Vec::new();
+        let mut program_queue: Vec<Command> = program;
+        program_queue.reverse();
+
+        while let Some(before_expanded_command) = program_queue.pop() {
```
saulshanabrook: nit: Why do we make a queue then pop off it instead of just iterating through the program?

src/lib.rs
```diff
@@ -1333,46 +1357,85 @@ impl EGraph {
...
+                    let s = std::fs::read_to_string(file)
+                        .unwrap_or_else(|_| panic!("{span} Failed to read file {file}"));
+                    let included_program = self
+                        .parser
+                        .get_program_from_string(Some(file.clone()), &s)?;
+                    // run program internal on these include commands
+                    let (included_outputs, _included_desugared) =
+                        self.run_program_internal(included_program, run_commands)?;
+                    outputs.extend(included_outputs);
+                    desugared_commands.push(ResolvedCommand::Include(span.clone(), file.clone()));
```
saulshanabrook, line 1393: Could we leave Include as it was before as happening in the de-sugarring pass, so we don't need this special case?

tests/files.rs
saulshanabrook, line 37: I think now I am a bit confused why we are even testing desugaring now, if we aren't trying to re-sugar… Like should we just not? I am trying to understand what bugs this desugar run would pick up that the normal run wouldn't.

saulshanabrook (changes requested):
This looks good overall, had some comments and questions. Thanks!

src/core.rs
saulshanabrook: Nevermind I see why you need all these since you expose typecheck_facts

egglog-ast/src/generic_ast_helpers.rs
oflatt: yep

src/ast/desugar.rs
oflatt: good catch

src/cli.rs
oflatt: I think "resugar" was a misnomer, fixed that here. Does the same thing as before.

src/core.rs
oflatt: I like the getters because it's easier to refactor the struct later if we reorganize the data. It's a practice I learned from the Cedar project.

src/lib.rs
oflatt: Because we push to the queue when we include new commands, and we need to process them for typechecking infromation before doing macro expansion.

src/lib.rs
oflatt: No because we want to run include before user-provided macros I think

tests/files.rs
oflatt: This does the same thing as before- tests that the program is still runnable after desugaring.

src/ast/desugar.rs
oflatt: I think `fresh` already does that right?

src/lib.rs
oflatt: good catch.

src/lib.rs
oflatt: the thing is we have to typecheck in order to run CommandMacro anyways so we would just be throwing away types. 

Yeah it's weird that it's one function, but having two functions is error-prone because they share a ton of logic.

tests/test_command_macros.rs
oflatt: done

src/ast/desugar.rs
oflatt: I'm confused by this question- maybe it's because I already moved fresh to experimental?

saulshanabrook: Yes this was for the old code, it was for a different commit even.

src/lib.rs
saulshanabrook: Maybe we could look at this together, I just seem to be missing something obvious, because I don't see program_queue reference after it was created besides in the loop iteration.

src/lib.rs
saulshanabrook: I see, so in this work we run the macros on the fully expanded program after include? I was wondering about doing that here versus effectively calling the macros in desugaring?

So then the macros would run once for the first proram with the includes as is, then would also be run basically recursively on each include… Then also if a macro wanted to change how include works it could do that? Maybe it would be cleaner? Or maybe not IDK.

Like run it in:

```
        Command::Include(span, file) => {
            let s = std::fs::read_to_string(&file)
                .unwrap_or_else(|_| panic!("{span} Failed to read file {file}"));
            return desugar_program(
                parser.get_program_from_string(Some(file), &s)?,
                parser,
                seminaive_transform,
            );
```

src/lib.rs
oflatt: Oh you're right! I had a version that used the queue and now I don't use it. Fixing now.

src/lib.rs
oflatt: Yeah exactly, macros still run on included code. Macros can't change how include works as it is now.

oflatt: I could make it so we run macros on the include command

yihozhang (approved):

saulshanabrook (approved):
Just chatted with Oliver and said we could always let macros change include later if we want, but can do it this way for now

PR #747, opened 2025-11-22
make necessary items public to allow custom type constraints

marko-subotic:
In egglog-experimental, I was trying to make a change that added a custom type constraint, but was unable to because several constructs were not public. This change will make `AtomTerm`, `Atom`, and `TypeInfo` all public

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -32,7 +32,8 @@ use ast::*;
 #[cfg(feature = "bin")]
 pub use cli::*;
 use constraint::{Constraint, Problem, SimpleTypeConstraint, TypeConstraint};
-use core::{AtomTerm, ResolvedAtomTerm, ResolvedCall};
+pub use core::{Atom, AtomTerm};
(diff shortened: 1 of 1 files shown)


marko-subotic:
> This sounds reasonable to me. Could you point me to some example code that uses the items this PR exposes?

For example to implement a custom type constraint such as the following, we have to import `AtomTerm`, `Typeinfo`, and `Atom`

```rust
impl TypeConstraint for SomeCustomTypeConstraint {
    fn get(
        &self,
        mut arguments: &[AtomTerm],
        _typeinfo: &TypeInfo,
    ) -> Vec<Box<dyn Constraint<AtomTerm, ArcSort>>> {
        ... code here
        // say we want an arity mismatch type constraint:
        ImpossibleConstraint::ArityMismatch {
            atom: Atom {
                    span: ...,
                    head: ...,
                    args: ...,
            },
            expected: ...,
        },
    }
}
```

yihozhang (approved):
Makes a lot of sense, thanks!

PR #752, opened 2025-11-28
Helpers for getting table sizes in primitives

oflatt:
This PR adds
- New methods for rust rules to get the size of tables
- Sanitize names more consistently for resugaring tests (fixes egglog-experimental resugaring tests)

### DIFF
--- a/core-relations/src/action/mod.rs
+++ b/core-relations/src/action/mod.rs
@@ -432,6 +432,11 @@ impl<'a> ExecutionState<'a> {
         &self.db.table_info[table].table
     }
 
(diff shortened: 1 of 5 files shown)


yihozhang:
Yes, I agree this is a problem with the new backend that many endpoints require using the low-level operations in `core-relations` which break abstractions.

yihozhang (approved):
After a discussion, we agreed to merge this (so that we can support herbie usage). A future PR will revert the dependence on `table_info.name()` and provide a wrapper around ExecutionState that supports mappings from a function name to id/other stats

PR #754, opened 2025-12-01
Low-level optimization for rebuilding

yihozhang:
1. Avoid using a concurrent structure for TypeId -> ContainerId map
2. Avoid creating a new RowBuffer when dropping a UfBuffer
3. Avoid creating a new buffer during rebuilding when the table does not have updates (only in single-thread mode)

One optimization we can do is to replace the SegQueue in pending_rows/pending_removals with a bespoke queue implementation by noticing:

* Each producer only produces once (when the containing `Buffer` is dropped)
* There is only one consumer, the consumer only consumes after all producers have produced (i.e., it has exclusive access), and it always consumes everything.

I had an implementation in [yihozhang-sharded-update-queue-legacy](https://github.com/egraphs-good/egglog/tree/yihozhang-sharded-update-queue-legacy) that tries to replace the SegQueue by pre-allocating each producer a slot, so that each producer doesn't need a write lock to write to the slot. Ideally this should turn the required write locks to reads when producing and consuming. However, my implementation actually slows things down.. So I will skip this for now, as it does not seem to contribute to the performance critically.

### DIFF
--- a/core-relations/src/pool/mod.rs
+++ b/core-relations/src/pool/mod.rs
@@ -395,7 +395,7 @@ macro_rules! pool_set {
             }
 
             $vis fn get<T: InPoolSet<Self> + Default>(&self) -> Pooled<T> {
-                self.get_pool().get()
+                T::with_pool(self, |pool| pool.get())
             }
             $vis fn clear(&self) {
                 $( self.$ident.clear(); )*
--- a/core-relations/src/table/rebuild.rs
+++ b/core-relations/src/table/rebuild.rs
@@ -2,14 +2,12 @@
 
 use std::{cmp, mem};
 
-use crate::numeric_id::{IdVec, NumericId};
-use crossbeam_queue::SegQueue;
+use crate::numeric_id::NumericId;
 use rayon::prelude::*;
 
 use crate::{
     ColumnId, ExecutionState, Offset, RowId, Subset, Table, TableId, TaggedRowBuffer, Value,
     WrappedTable,
-    common::ShardId,
     hash_index::{ColumnIndex, Index},
     parallel_heuristics::parallelize_rebuild,
     table_spec::{Rebuilder, WrappedTableRef},
@@ -92,12 +90,6 @@ impl SortedWritesTable {
         );
 
         if parallelize_rebuild(to_scan.size()) {
-            // Iterate over `buf` in parallel and then fan out to a per-shard set of rows then
-            // process each shard in parallel.
-            let mut queues = IdVec::<ShardId, SegQueue<TaggedRowBuffer>>::default();
-            let shard_data = self.hash.shard_data();
-            queues.resize_with(shard_data.n_shards(), SegQueue::new);
-
             WrappedTableRef::with_wrapper(self, |wrapped| {
                 buf.par_iter()
                     .fold(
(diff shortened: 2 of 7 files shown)


core-relations/src/pool/mod.rs
```diff
@@ -395,7 +395,7 @@ macro_rules! pool_set {
             }
 
             $vis fn get<T: InPoolSet<Self> + Default>(&self) -> Pooled<T> {
-                self.get_pool().get()
+                T::with_pool(self, |pool| pool.get())
```
ezrosent, line 398: Why was this change needed?

core-relations/src/table/rebuild.rs
```diff
@@ -141,13 +132,15 @@ impl SortedWritesTable {
...
+            }
+            if !scratch.is_empty() {
+                let mut write_buf = self.new_buffer();
                 for (row_id, row) in scratch.non_stale_mut() {
                     if let Some(to_remove) = self.data.get_row(row_id).map(|x| &x[0..self.n_keys]) {
                         write_buf.stage_remove(to_remove);
                     }
                     insert_row!(self, write_buf, row, next_ts);
                 }
-                scratch.clear();
```
ezrosent, line 150: Seems like we still need this? Do we want to assume that scratch is always cleared by rebuild_subset?

core-relations/src/uf/mod.rs
```diff
@@ -221,17 +222,25 @@ impl Clone for DisplacedTable {
...
         let Some(buffered_writes) = self.buffered_writes.upgrade() else {
+            // SAFETY: If we can't write updates, manually drop to_insert
+            unsafe {
+                ManuallyDrop::drop(&mut self.to_insert);
+            }
             return;
         };
-        let arity = self.to_insert.arity();
-        buffered_writes.push(mem::replace(&mut self.to_insert, RowBuffer::new(arity)));
+        // SAFETY: self.to_insert will not be used again after this point.
```
ezrosent, line 238: Just want to confirm that you measured with and without this specific optimization on. I'm a little surprised that it helps! (I wonder if it points to an issue with RowBuffer being too complex... I can take a look at that later).

numeric-id/src/lib.rs
```diff
@@ -74,7 +74,9 @@ impl<K: NumericId, V> DenseIdMap<K, V> {
     /// Create an empty map with space for `n` entries pre-allocated.
     pub fn with_capacity(n: usize) -> Self {
         let mut res = Self::new();
-        res.reserve_space(K::from_usize(n));
+        if n > 0 {
+            res.reserve_space(K::from_usize(n - 1));
```
ezrosent: remove the `if` and instead res.reserve_space(K::from_usize(n.saturating_sub(1))) ?

ezrosent (approved):
Thanks! Nice speedups here. Minor comments.

core-relations/src/table/rebuild.rs
yihozhang: `rebuild_subset` is now its own loop, executed before any updates to the buffer. This iteration over scratch is now the last operation and the scratch is not reused.

core-relations/src/pool/mod.rs
yihozhang: The call to get_pool incurred an unnecessary clone I think.

core-relations/src/uf/mod.rs
yihozhang: Yeah the profiler showed a long flat bar for this for the tree/linked list traversal benchmark. This was the first optimization I found actually. The precise merge PR might make this optimization no longer relevant but I have not tested.

PR #758, opened 2025-12-10
Fix Extractor for DAG cost model

saulshanabrook:
This PR attempts to resolve https://github.com/egraphs-good/egglog-python/issues/387, which is where a panic is currently raised sometimes when using a DAG based cost model with the current extractor. It seems to resolve those cases and passes all tests here and in the Python library around extraction, so it seems correct… But @FTRobbin you would be the best one to understand if this seems like an OK change!


FTRobbin (commented):
- This breaks the correspondence between the computed cost and the reconstructed extraction. The reconstruction's actual cost may exceed the computed cost because this code only remembers the enode that leads to the lowest cost, not the full term. However, this does always produce a valid term if a term's cost is always greater than or equal to any of its subterms.
- This is also suboptimal in its implementation. It copies a table row at every update, and `topo` data structures are actually redundant if copies are made all the time. This hits the performance of extract-heavy workflows, which we don't have in the test suite right now but have observed in Herbie.

I recommend we reject this change. If any valid term is good enough for DAG cost, we should implement a separate reconstruction pass for that.

saulshanabrook:
Thanks @FTRobbin for the review! I updated the PR to keep the existing behavior if the edges exist, and just to fall back to retrieving an edge if needed during reconstruction. Alternatively, I could add a second pass before reconstruction to store the parent edges so that reconstruction is still cheaper…

saulshanabrook:
We chatted about how for DAG costs there are multiple ways to implement it, including a greedy algorithm that maintains a top n options for each e-class to not be totally greedy.

saulshanabrook:
Closing this and will upstream the slower but dag compatible extractor in the Python bindings for now to support dag costs. Later on, this could be moved to egglog experimental to add support for dag extraction there.

PR #759, opened 2025-12-10
Simplify extract option handling

saulshanabrook:
Removes a few lines from extract to help make the code slightly easier to understand. This shouldn't change behavior at all, it just changes some of the option handling to use ? and also only looks up the cost once instead of twice.


src/extract.rs
```diff
@@ -291,11 +272,7 @@ impl<C: Cost + Ord + Eq + Clone + Debug> Extractor<C> {
         //log::debug!("compute_cost_hyperedge head {} sorts {:?}", head, sorts);
```
ezrosent: Should we remove this line as well?

ezrosent (approved):
Nice cleanup. Thanks!

PR #760, opened 2025-12-11
Add longer extraction benchmark

saulshanabrook:
This PR adds a herbie example as a benchmark, which has a lot of time spent in extraction according to @FTRobbin 

It was copied from code @noahhuck generated in https://github.com/ajpal/poach/blob/1de0e5bf540130ea1e1520bf006907f33590a286/infra/nightly-resources/test-files/herbie-math-taylor/taylor7.egg

This was added to help benchmark a change in the extractor in https://github.com/egraphs-good/egglog/pull/758


FTRobbin (approved):
Looks good to me. We should keep in mind that this is not a stable version of Herbie's use of egglog and remove it from the test suite when a better substitute is found in the future.

PR #761, opened 2025-12-12
Support Different Extractors in Egglog

saulshanabrook:
~~This PR adds the ability to register different extractors with egglog in rust and then use them when calling `(extract expr :extractor <extractor name>)`.~~ This PR adds the ability to change the extractor from Rust used to during extraction. This can work with the builtin cost of `usize` but also can support other dynamic costs types.

This removes the need in experimental to register a custom extract command for the set cost cost model, and instead it can just register it and set it as default. (https://github.com/egraphs-good/egglog-experimental/pull/37)

It also [removes the need in Python](https://github.com/egraphs-good/egglog-python/blob/6a892ec50db3ef1b242ac97cdd8597e86d9f6e6b/python/egglog/egraph.py#L1027-L1034) to expose the extractor object and can instead use the normal code path of calling the extract command when we want to use a non standard cost model.

Originally, I just supported registering custom cost models, not arbitrary extractors… However if we want to support dynamic cost types, most of this code was still required, so I thought it made more sense to expose the general way of registering a custom extractor, not just a cost model. Also, this could support other extractor in the future without having to change our API, like if we wanted to hook into a SMT solver.

~~I know there was some questions if we wanted to settle at this point on whether we want this kind of API or if its too early. I made this PR as a proof of concept to understand the complexity involved, so I won't be offended if we decide not to merge it. I was also inspired to try this now since oliver I believe was blocked on some type checking on the current extract command in experimental. Eventually that could be resolved by reworking our macro system, but this would at least let his code work before then.~~ I originally exposed this in the egglog file as a way to change between registered extractors. I removed that feature, to keep this change more minimal and because I didn't see an immediate need for that flexibility. In Python, we can just change between extractors when users pass them in.

### DIFF
--- a/src/ast/desugar.rs
+++ b/src/ast/desugar.rs
@@ -111,9 +111,7 @@ pub(crate) fn desugar_command(
         Command::PrintOverallStatistics(span, file) => {
             vec![NCommand::PrintOverallStatistics(span, file.clone())]
         }
-        Command::Extract(span, expr, variants, extractor) => {
-            vec![NCommand::Extract(span, expr, variants, extractor)]
-        }
+        Command::Extract(span, expr, variants) => vec![NCommand::Extract(span, expr, variants)],
         Command::Check(span, facts) => vec![NCommand::Check(span, facts)],
         Command::PrintFunction(span, symbol, size, file, mode) => {
             vec![NCommand::PrintFunction(span, symbol, size, file, mode)]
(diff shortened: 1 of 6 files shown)


saulshanabrook:
Changing this to a draft for now, since I didn't realize it would add this much complexity with the dynamic costs.

Issue #749, opened 2025-11-25
CLI arguments set from egglog


oflatt:
It would be nice if all of the CLI arguments (proofs, serialization settings, ect) could be set in the egglog file itself. Something like a `set-option` command we used to have. 

I'm not sure what should happen if you pass a command line argument and it's set in the egglog file. Error? Override?

PR #765, opened 2025-12-20
Term encoding of egglog, which could make proofs easier

oflatt:
This PR adds an encoding of egglog using *term tables* and *view tables* behind a flag.
It uses an explicit union find table for each sort, and eliminates all `union` calls from the program.
At the end, queries use view tables (canonicalized enodes). Actions add to term and view tables. 

The term encoding doesn't support all of egglog yet. 

It starts to add support for proofs behind a flag but it's not done yet.

The term encoding doesn't support looking up tables, even at the global scope. So we do a different global desugaring using `union` instead of `set`. This means global primitives are not currently supported ):

It also
- Changes desugaring to get rid of include commands completely

### DIFF
--- a/src/typechecking.rs
+++ b/src/typechecking.rs
@@ -514,6 +514,8 @@ impl TypeInfo {
         constraints.extend(query.get_constraints(self)?);
 
         let mut binding = query.get_vars();
+        // We lower to core actions with `union_to_set_optimization`
+        // later in the pipeline. For typechecking we do not need it.
         let mut ctx = CoreActionContext::new(self, &mut binding, symbol_gen, false);
         let (actions, mapped_action) = head.to_core_actions(&mut ctx)?;
 
@@ -624,6 +626,8 @@ impl TypeInfo {
     ) -> Result<ResolvedActions, TypeError> {
         let mut binding_set: IndexSet<String> =
             binding.keys().copied().map(str::to_string).collect();
+        // We lower to core actions with `union_to_set_optimization`
+        // later in the pipeline. For typechecking we do not need it.
         let mut ctx = CoreActionContext::new(self, &mut binding_set, symbol_gen, false);
         let (actions, mapped_action) = actions.to_core_actions(&mut ctx)?;
         let mut problem = Problem::default();


src/core.rs
```diff
@@ -611,6 +656,7 @@ where
                 }
                 GenericAction::Union(span, e1, e2) => {
                     match (e1, e2) {
+                        /* TODO why did we have this case? And why is it using Set?
```
oflatt: This comment out what I believe is an optimization. The problem is that it uses `set` instead of `union` on constructors, causing problems for terms encoding. I think this optimization should come later in the pipeline?

egglog-ast/src/generic_ast_helpers.rs
```diff
@@ -357,6 +400,31 @@ where
             }
         }
     }
+
+    pub fn make_unresolved(self) -> GenericAction<String, String> {
```
ezrosent: document this, and any other new public methods.

ezrosent: also, should these make_unresolved impls be an instance of a generic 'map' method?

src/core.rs
```diff
@@ -75,13 +75,33 @@ pub enum ResolvedCall {
...
     pub fn output(&self) -> &ArcSort {
         match self {
             ResolvedCall::Func(func) => &func.output,
             ResolvedCall::Primitive(prim) => prim.output(),
         }
     }
 
+    /// Gives the types for a term's child with the given resolved call.
+    /// For functions this includes the output sort, for constructors it's just the inputs.
+    pub(crate) fn view_types(&self) -> Vec<ArcSort> {
```
ezrosent, line 95: where is this used?

src/core.rs
```diff
@@ -509,6 +529,7 @@ pub(crate) trait GenericActionsExt<Head, Leaf> {
         typeinfo: &TypeInfo,
         binding: &mut IndexSet<Leaf>,
         fresh_gen: &mut FG,
+        union_to_set_optimization: bool,
```
ezrosent: It might be easier / more readable to package some of these args up into a struct. That way each call is easier to type and we can pass arguments by name

src/lib.rs
```diff
@@ -348,6 +354,20 @@ impl Default for EGraph {
 pub struct NotFoundError(String);
 
 impl EGraph {
+    /// Create a new e-graph with term encoding enabled.
+    /// This instruments the egglog to use a explicit equality relation.
```
ezrosent: I think there should be a more detailed doc comment somewhere explaining what the term encoding is. Maybe we can just point to that?

src/lib.rs
```diff
@@ -348,6 +354,20 @@ impl Default for EGraph {
...
+    pub fn new_with_term_encoding() -> Self {
+        let mut egraph = EGraph::default();
+        egraph.proof_state.original_typechecking = Some(Box::new(egraph.clone()));
+        egraph
+    }
+
+    /// WARNING: this should be called before any commands are added to the e-graph.
+    pub fn with_term_encoding_enabled(mut self) -> Self {
```
ezrosent, line 378: Why do we need this? Seems like a bit of a footgun to allow enabling it after construction.

src/lib.rs
```diff
@@ -1349,14 +1380,56 @@ impl EGraph {
...
     fn resolve_command(&mut self, command: Command) -> Result<Vec<ResolvedNCommand>, Error> {
         let desugared = desugar_command(command, &mut self.parser)?;
-        let mut typechecked = self.typecheck_program(&desugared)?;
 
-        typechecked = remove_globals::remove_globals(typechecked, &mut self.parser.symbol_gen);
-        for command in &typechecked {
-            self.names.check_shadowing(command)?;
-        }
+        // Add term encoding when it is enabled
+        if self.proof_state.original_typechecking.is_some() {
```
ezrosent: use `if let` here? or is the borrow checker getting in the way of that somehow

src/lib.rs
```diff
@@ -1349,14 +1380,56 @@ impl EGraph {
...
+                .unwrap()
+                .typecheck_program(&desugared)?;
+
+            typechecked =
+                proof_global_remover::remove_globals(typechecked, &mut self.parser.symbol_gen);
+            for command in &typechecked {
+                self.names.check_shadowing(command)?;
+
+                if !command_supports_proof_encoding(&command.to_command()) {
+                    panic!(
```
ezrosent: I think this should return an Error, not panic?

src/typechecking.rs
```diff
@@ -623,7 +625,7 @@ impl TypeInfo {
         let mut binding_set: IndexSet<String> =
             binding.keys().copied().map(str::to_string).collect();
         let (actions, mapped_action) =
-            actions.to_core_actions(self, &mut binding_set, symbol_gen)?;
+            actions.to_core_actions(self, &mut binding_set, symbol_gen, false)?;
```
ezrosent: why always false here? (explain in a comment)

ezrosent (changes requested):
Some minor comments to start

src/core.rs
oflatt: The rebuilding rules use it to grab the types that need rebuilding

src/lib.rs
oflatt: Yeah I thought the same thing, but the way experimental works is that it builds an e-graph before command line arguments are interpreted. Suggestions on how to fix?

ezrosent: Any chance we can change experimental? Maybe as a TODO?

oflatt: Opened an issue here https://github.com/egraphs-good/egglog/issues/768
I also have this issue which is related:  https://github.com/egraphs-good/egglog/issues/749

egglog-ast/src/generic_ast_helpers.rs
```diff
@@ -194,6 +195,50 @@ where
             ruleset: self.ruleset.clone(),
         }
     }
+
+    /// Applies `f` to each action in the rule head, leaving the body unchanged.
+    pub fn visit_actions(
+        self,
+        f: &mut impl FnMut(GenericAction<Head, Leaf>) -> GenericAction<Head, Leaf>,
```
ezrosent, line 202: nit: for visitors like this I think having `f` be an `impl FnMut(...)->...` and then simply passing `&mut f` in recursive calls is cleaner. It amounts to the same thing, but it doesn't "leak" this fact about recursive calls into the interface and force all callers to do `&mut`.

src/term_encoding.rs
```diff
@@ -0,0 +1,1255 @@
...
+//!
+//! Lowering the program with term encoding expands it to the following commands:
+//!
+//! ```text
+//! (ruleset parent)
+//! (ruleset single_parent)
+//! (ruleset rebuilding)
+//! (ruleset rebuilding_cleanup)
+//! (ruleset delete_subsume_ruleset)
+//! (sort _view1)
```
ezrosent, line 30: `@view1` (sim for merge cleanup)

src/term_encoding.rs
```diff
@@ -0,0 +1,1255 @@
...
+//! (constructor add (i64 i64 i64) @view1)
+//! (constructor addView (i64 i64 i64) @view1)
+//! (constructor to_delete_add (i64 i64) @view1)
+//! (constructor to_subsume_add (i64 i64) @view1)
+//! (sort _mergecleanup8)
+//! (constructor _mergecleanup7 (i64 i64) @mergecleanup8)
+//! (rule ((addView c0_ c1_ old)
```
ezrosent, line 37: It'd probably help the example to include `union` as well, wdyt?

src/term_encoding.rs
```diff
@@ -0,0 +1,1255 @@
...
+            self.egraph.proof_state.uf_parent.insert(
+                sort.to_string(),
+                format!("{INTERNAL_SYMBOL_PREFIX}ufparent_{}", sort,),
+            );
+            self.egraph.proof_state.uf_parent[sort].clone()
+        }
+    }
+
+    pub(crate) fn uf_proof_name(&mut self, sort: &str) -> String {
+        format!("{}Proof", self.uf_name(sort),)
```
ezrosent, line 156: comma

src/term_encoding.rs
```diff
@@ -0,0 +1,1255 @@
...
+    }
+
+    fn single_parent_ruleset_name(&self) -> String {
+        "single_parent".to_string()
+    }
+
+    fn to_ast_name(&self, sort: &str) -> String {
+        format!("AST{}", sort,)
```
ezrosent, line 164: comma

src/term_encoding.rs
```diff
@@ -0,0 +1,1255 @@
...
+                   :name \"{fresh_name}\")
+             (rule (({pname} a b)
+                    ({pname} a c)
+                    (!= b c))
+                  ((delete ({pname} a (ordering-max b c)))
+                   ({pname} (ordering-max b c) (ordering-min b c)))
+                   :ruleset {}
+                   :name \"singleparent{fresh_name}\")
+                   ",
+            self.parent_direct_ruleset_name(),
```
ezrosent, line 252: nit: I think it'll make this a bit easier to read to name these positional args too


```rust
format!("
:ruleset {parent_ruleset_name}
...
",
parent_ruleset_name = self.parent_direct_ruleset_name(),
...
);
```

src/term_encoding.rs
```diff
@@ -0,0 +1,1255 @@
...
+                            ListDisplay(args, " ")
+                        ));
+                        fv
+                    }
+                }
+            }
+        }
+    }
+
+    // Returns a query, variable for the updated child, and proof
```
ezrosent, line 689: nit: I think this can be a doc comment

(also, please expand this comment it is a bit opaque)

egglog-ast/src/generic_ast_helpers.rs
oflatt: Wouldn't that make it so you can't re-use the same `f` for two different `visit_actions` or `visit_...` calls? I do wonder if that would ever come up

src/term_encoding.rs
oflatt: Will fix this to use underscore everywhere so it can be run again.

PR #766, opened 2025-12-23
Precise merge

ezrosent:
This PR adds support for "precise merging" to core-relations. Basically, instead of re-scanning all tables and checking if they've changed, this PR allows merging to only call `merge` on tables that have registered a change.

cc @yihozhang who has some private (?) profiling infra that may help shed light on what other areas there are to optimize.

### DIFF
--- a/core-relations/src/table_spec.rs
+++ b/core-relations/src/table_spec.rs
@@ -332,7 +332,9 @@ pub trait Table: Any + Send + Sync {
     /// the table.
     fn merge(&mut self, exec_state: &mut ExecutionState) -> TableChange;
 
-    /// Create a new buffer for staging mutations on this table.
+    /// Create a new buffer for staging mutations on this table. Mutations staged to a
+    /// MutationBuffer that is then dropped may not take effect until the next call to
+    /// [`Table::merge`].
     fn new_buffer(&self) -> Box<dyn MutationBuffer>;
 }
 


core-relations/src/table/rebuild.rs
```diff
@@ -126,23 +119,28 @@ impl SortedWritesTable {
...
             let mut write_buf = self.new_buffer();
+            let mut changed = false;
             for (_, id) in buf.iter() {
                 let Some(subset) = self.rebuild_index.get_subset(&id[0]) else {
                     continue;
                 };
                 WrappedTableRef::with_wrapper(self, |wrapped| {
                     rebuilder.rebuild_subset(wrapped, subset, &mut scratch, exec_state);
                 });
+                changed = subset.size() > 0;
```
yihozhang: Should this be `changed |= subset.size() > 0;`?

concurrency/src/tests.rs
```diff
@@ -147,6 +191,79 @@ fn basic_parallel_vec_push() {
...
+        v_inner.resize_with(128, || 2);
+    });
+
+    barrier.wait();
+    left.join().unwrap();
+    right.join().unwrap();
+
+    let slice = v.read();
+    assert_eq!(slice.len(), 128);
+    assert!(slice.iter().all(|&value| value == 1 || value == 2));
```
yihozhang, line 233: nit: Could you also test for the behavior that slice[i] == slice[i+1] for all i != 63?

core-relations/src/free_join/mod.rs
```diff
@@ -288,6 +289,7 @@ pub struct Database {
     pub(crate) counters: Counters,
     pub(crate) external_functions: ExternalFunctions,
     container_values: ContainerValues,
+    notification_list: NotificationList<TableId>,
```
yihozhang, line 294: Nit: the variable name itself is not descriptive of its intention. Maybe add a small comment here?

core-relations/src/free_join/mod.rs
```diff
@@ -600,20 +633,41 @@ impl Database {
     /// Get direct mutable access to the table.
     ///
     /// This method is useful for out-of-band access to databse state.
-    pub fn get_table(&self, id: TableId) -> &WrappedTable {
+    ///
+    // Not in doc comment since it mentions private items:
+    //
+    // **NOTE:** It is legal to call [`Table::new_buffer`] on the returned table handle, and use
```
yihozhang: Why is this "legal" and only "likely" a source of bugs? If a table is updated this way solely, are such updates always forgotten?

core-relations/src/free_join/mod.rs
```diff
@@ -529,21 +544,39 @@ impl Database {
         ever_changed
     }
 
+    /// A "fast path" merge method that is not optimized for parallelism and does not respect read
+    /// and write dependencies. This ends up being faster than the full "strata-aware" option in
+    /// the body of `merge_all`.
+    fn merge_simple(&mut self, mut to_merge: SmallVec<[TableId; 4]>) -> bool {
+        let mut changed = false;
+        while !to_merge.is_empty() {
+            for table_id in to_merge.iter().copied() {
```
yihozhang, line 555: A small number of tables to merge does not imply a small or a balanced workload, e.g., the AC workload, so parallelization can still help. Should we pass the parallelization flag from the call site and optionally do `par_iter` here?

core-relations/src/free_join/mod.rs
```diff
@@ -529,21 +544,39 @@ impl Database {
...
         let mut info = self.tables.unwrap_val(table);
         self.total_size_estimate = self.total_size_estimate.wrapping_sub(info.table.len());
-        let _table_changed = info.table.merge(&mut ExecutionState::new(
+        let table_changed = info.table.merge(&mut ExecutionState::new(
             self.read_only_view(),
             Default::default(),
         ));
         self.total_size_estimate = self.total_size_estimate.wrapping_add(info.table.len());
         self.tables.insert(table, info);
+        table_changed.added || table_changed.removed
```
yihozhang: One thing I'm curious about is that merge_all and merge_simple only count `added` as an update, while here we count both `added` and `removed`. Also, is the returned flag ever used currently?

yihozhang (commented):
Great PR! Super excited about the performance improvements.

core-relations/src/free_join/mod.rs
ezrosent: In that case, we would still parallelize merge for each table with a large set of merges (`Add` or `Mul` in the AC case).

I think we'll gain a decent amount from merge_simple but we can tweak the cutoff to equal 1 if you want to preserve parallelism?


The reason we can't do par_iter is that to_merge[0] may need access to to_merge[1], or similar, and the loop body requires that we remove `table_id` from table_info. The way to safely parallelize across merges is in the calling function, that iterates based on dependencies. (a good deal more complex / more constant factors).

core-relations/src/free_join/mod.rs
ezrosent: Yeah... I think just `added` is needed? I think this is a subtle question. You can definitely write rules that reach a different fixed point based on `removed` by writing an `ExternalFunction` that does something based on a row not being present in a table. But such a rule would not be legal egglog as it would allow for the encoding of non-stratified negation / break unique fixed points. I'll remove the `removed` check for now

core-relations/src/free_join/mod.rs
ezrosent: Done!

core-relations/src/free_join/mod.rs
ezrosent: Note that this is _just_ about core-relations. Doing this sort of thing is impossible for external users of this crate.

I think the reason it has to be allowed is that we need `WrappedTable` to implement `Table`, and `new_buffer` has to be a method on table, otherwise there wouldn't be a way for non-core-relations crates to implement their own tables. 

Open to refactor options here though! I think it's mostly a concern for tests in the core-relations crate.

core-relations/src/table/rebuild.rs
ezrosent: yup, thanks!

concurrency/src/tests.rs
ezrosent: added, thanks!

core-relations/src/free_join/mod.rs
yihozhang: I see! This makes sense. How about setting the threshold to 4, which is also the SmallVec threshold of the notification list?

Do you know which benchmark can be sped up by `merge_simple`? I tried one eggcc benchmark and the speedup for that is <5% (which is totally fine for a complex benchmark, but I wonder if some benchmarks can demonstrate this optimization).

core-relations/src/free_join/mod.rs
yihozhang: Thanks for the clarification! It would be great if we could incorporate part of your comments into the documentation (e.g., the assumption that downstream crates shouldn't expose `Database` (and `Table`?) to users).

core-relations/src/free_join/mod.rs
```diff
@@ -529,21 +546,39 @@ impl Database {
...
         let mut info = self.tables.unwrap_val(table);
         self.total_size_estimate = self.total_size_estimate.wrapping_sub(info.table.len());
-        let _table_changed = info.table.merge(&mut ExecutionState::new(
+        let table_changed = info.table.merge(&mut ExecutionState::new(
             self.read_only_view(),
             Default::default(),
         ));
         self.total_size_estimate = self.total_size_estimate.wrapping_add(info.table.len());
         self.tables.insert(table, info);
+        table_changed.added
```
yihozhang, line 581: Actually, `merge_all` not only checks the `added` flag, but also checks `ExecutionState::changed`, which is set to true whenever `stage_remove` is called. So we are already checking for removal in `merge_all`.

I was initially thinking that only counting `added` can cause a problem because rebuilding can shrink the database without introducing new tuples. But actually, the current implementation always removes all conflicting tuples and *inserts* the canonical tuple (with the updated timestamp), so this is fine. Then it seems the only places `stage_remove` is called without `stage_insert` are `delete` and user defined functions? Anyway, this is very subtle.

The code is also a bit complex right now, as there are multiple sources of truth. This is okay for now I think.

core-relations/src/free_join/mod.rs
ezrosent: on the python-array-optimize test, this command:

```
time target/release/egglog tests/python_array_optimize.egg > /tmp/x
```
Goes from 1.2s with merge_simple commented out to 0.9s with the merge_simple path enabled (on my m4 max laptop)

I'll change it to 4 though!

core-relations/src/free_join/mod.rs
ezrosent: > Then it seems the only places stage_remove is called without stage_insert are delete and user defined functions? Anyway, this is very subtle.

That is correct. I don't think we have a super clear notion of what it means to reach a fixed point at the core-relations level. It's possible that removal being used everywhere is the right heuristic, but I don't think it's super important for this PR one way or the other.

ezrosent:
So I think any external user of `core-relations` _can_ actually do this direct modification. I agree this isn't ideal, but the main ways around it introduce compromises.

perhaps the cleanest way is for `WrappedTable` to directly implement `Table` instead of just having a `deref`, and then having it trigger the change notification machinery. That would introduce quite a lot of extraneous code for not much benefit, but I could see it being better from an encapsulation perspective.

All other changes should be good to go though. Lmk what you think!

yihozhang:
This looks good! The only place new_buffer is currently used outside of `core-relations` is tests and benchmarks. Maybe also leave a sentence on `Table::new_buffer` saying that external users calling this should also call `Table::merge` after dropping the buffer? (We can also enforce this behavior with a new_buffer_safe that returns a wrapper over MutationBuffer; I'll leave it to you to decide if it's feasible and appropriate). 

Thanks for the PR again, and feel free to merge when you feel ready!

PR #770, opened 2025-12-28
Proof encoding in egglog itself

oflatt:
This PR

- Changes the TermDag API to use TermIds, simplifying lots of code. 
- Adds a new proof mode that can be enabled with the `--proofs` flag.
- Instruments egglog's term encoding with proofs that can be obtained with the `prove` command.
- Adds a proof format to the public API and a checker for the format.
- Adds primitive validators and a new macro for easily making checkable primitives. 
- Adds detection for egglog programs proofs currently do not support (containers, ect)
- Adds a notion of "proof normal form" with slight restrictions on the egglog query format.

In the future we'll want
- A way to bind an extraction to a variable so we can use then `prove` to prove equivalence. 
- An optimization that makes it possible to run math_microbenchmark
- Expanding the scope of benchmarks we support

### DIFF
--- a/tests/snapshots/files__proof_unsupported_files.snap
+++ b/tests/snapshots/files__proof_unsupported_files.snap
@@ -24,6 +24,7 @@ fail-typecheck/set-a-primitive.egg
 fail-typecheck/shadow-global-constructor.egg
 fail-typecheck/shadow-global.egg
 fail-typecheck/shadow-local-constructor.egg
+fail-typecheck/shadow-local-global-no-prefix.egg
 fail-typecheck/shadow-local-global.egg
 fail-typecheck/shadow-local.egg
 fail-typecheck/unbound.egg


src/sort/add_primitive/src/lib.rs
```diff
@@ -392,3 +463,138 @@ impl Parse for Arrow {
...
+/// It automatically generates validators by converting between Rust values and Literal types.
+///
+/// # Example
+/// ```rust,ignore
+/// add_literal_prim!(eg, "not" = |a: bool| -> bool { !a });
+/// add_literal_prim!(eg, "+" = |a: i64, b: i64| -> i64 { a + b });
+/// add_literal_prim!(eg, "<" = |a: i64, b: i64| -> bool { a < b });
+/// ```
+#[proc_macro]
+pub fn add_literal_prim(input: TokenStream) -> TokenStream {
```
oflatt, line 457: I'm bad a writing macros, this one is AI generated so take care. 
The rest of the code in this PR is mostly hand written

src/proofs/proof_encoding.md
```diff
@@ -0,0 +1,262 @@
...
+
+Similarly, the union-find table gets a proof table storing
+  proofs of equalities between terms.
+If term `a` has parent `b`, it stores a 
+  proof of `a = b`.
+
+
+
+```text
+(constructor AddView (i64 i64 Math) view)
```
oflatt: Need to give the actual view proof table here

src/proofs/proof_encoding.md
```diff
@@ -0,0 +1,262 @@
...
+```
+
+View tables are the trickiest.
+Recall that view tables store a term
+  along with the e-class representative.
+For a term `t` with representative `r`,
```
oflatt, line 281: Make this clearer that the view proof table is separate

src/proofs/proof_encoding_helpers.rs
```diff
@@ -0,0 +1,567 @@
...
+        let pnil = &self.proof_names().pnil;
+
+        let mut prooflist = format!("({pnil})");
+        for proof in proofs.iter().rev() {
+            prooflist = format!("({pcons} {proof} {prooflist})");
+        }
+        prooflist
+    }
+
+    pub(crate) fn parent_direct_ruleset_name(&self) -> String {
```
oflatt: add documentation for what this does

src/proofs/proof_encoding_helpers.rs
```diff
@@ -0,0 +1,567 @@
...
+(sort {proof_datatype})
+
+(constructor {pcons} ({proof_datatype} {proof_list_sort}) {proof_list_sort})
+(constructor {pnil} () {proof_list_sort})
+
+{to_ast_str}
+
+;; Fiat justification for globals and primitives, gives two terms t1 = t2 for the proposition being justified
+(constructor {fiat_constructor} ({ast_sort} {ast_sort}) {proof_datatype})
+;; name of rule and one proof per fact in the query
```
oflatt: proposition being proven is also there

src/proofs/proof_normal_form.rs
```diff
@@ -0,0 +1,181 @@
...
+/// Puts queries in "proof normal form". In proof form:
+/// 1. Function calls like (lower-bound a b) are always top level and look like this:
+///    (= (lower-bound a b) c)
+/// 2. Primitives can't have constructors or function calls as arguments.
+///    For example, (!= a (Const 0)) becomes:
+///    (= (Const 0) v), (!= a v)
+///
+/// Nested function calls like this are not allowed:
+/// (Add c (lower-bound a b))
+pub(crate) fn proof_form(
```
oflatt, line 15: Clarify that this function transforms the program, it doesn't check that it's in proof normal form

tests/subsume-relation.egg
```diff
@@ -6,4 +6,5 @@
 (R (A) (C))
 (union (B) (C))
 
-(check (= (R (A) (B)) ()))
+; TODO term encoding and egglog semantics don't agree- asking in slack
```
oflatt: Need to check on this again

src/proofs/proof_checker.rs
```diff
@@ -0,0 +1,907 @@
...
+) -> Result<(TermId, HashSet<Proposition>), ProofCheckError> {
+    let mut propositions = HashSet::default();
+
+    let term_id = match expr {
+        ResolvedExpr::Lit(_, lit) => dag.lit(lit.clone()),
+        ResolvedExpr::Var(_, var) => {
+            subst
+                .get(&var.name)
+                .copied()
+                .ok_or(ProofCheckError::RuleSubstitutionMismatch {
```
oflatt: Rename to proof unbound variable in substitution

src/proofs/proof_checker.rs
```diff
@@ -0,0 +1,907 @@
...
+                            "Fact {fact} does not match proven proposition under substitution {subst_str}.\nSubstituted: (= {fact_lhs_str} {fact_rhs_str})\nPremise proves: (= {} {})\nProof: {proof_str}",
+                            format_term(&self.term_dag, lhs),
+                            format_term(&self.term_dag, rhs),
+                        ),
+                    });
+                }
+
+                Ok(())
+            }
+            // For a plain expr, the proof should have the form t1 = t2 where t2 matches the expr under substitution
```
oflatt, line 900: I think in practice it's always t=t but should make sure that's true in this function

egglog-bridge/src/tests.rs
ezrosent, line 122: Need to rebase for these changes

src/ast/desugar.rs
```diff
@@ -139,11 +146,60 @@ pub(crate) fn desugar_command(
         Command::UserDefined(span, name, args) => {
             vec![NCommand::UserDefined(span, name, args)]
         }
+        Command::Prove(span, query) => desugar_prove(parser, span, query),
```
ezrosent, line 149: What happens if these prove commands are run but proof_testing is false? Should we return an error for that case?

src/ast/desugar.rs
```diff
@@ -139,11 +146,60 @@ pub(crate) fn desugar_command(
...
 
     Ok(res)
 }
 
+fn desugar_prove(parser: &mut Parser, span: Span, query: Vec<Fact>) -> Vec<NCommand> {
+    let fresh_sort = parser.symbol_gen.fresh("ExistsSort");
+    let constructor_name = parser.symbol_gen.fresh("ExistsConstructor");
+    let ruleset = parser.symbol_gen.fresh("exists");
+    let name = parser.symbol_gen.fresh("prove_exists_rule");
+    vec![
```
ezrosent, line 178: I think it'd help to provide "pseudo-egglog" in a comment here to explain what this is

src/ast/mod.rs
```diff
@@ -236,6 +284,62 @@ where
...
+                span,
+                generic_schedules
+                    .into_iter()
+                    .map(|schedule| schedule.visit_queries(f))
+                    .collect(),
+            ),
+        }
+    }
+
+    fn squash_sequences(self) -> Self {
```
ezrosent: nit: Document what this does (does it normalize sequences by flattening them all? Flatten would be better than Squash if that's the case. )

src/proofs/proof_checker.rs
```diff
@@ -0,0 +1,926 @@
...
+    pub var_bindings: HashMap<String, TermId>,
+    /// Propositions (equalities) implied by the actions
+    pub propositions: HashSet<Proposition>,
+}
+
+/// Gathers all global CoreActions from a program.
+/// This extracts all actions that occur at the top level, filtering out NormRule and other commands.
+pub(crate) fn gather_global_actions(
+    prog: &[ResolvedNCommand],
+) -> Vec<&GenericAction<ResolvedCall, crate::ast::ResolvedVar>> {
```
ezrosent: nit: can this return an iterator?

```rust
prog.filter_map(|cmd| {
if let GenericNCommand::CoreAction(action) = cmd { Some(action) } else { None }
})
```

src/proofs/proof_checker.rs
```diff
@@ -0,0 +1,926 @@
...
+        },
+    };
+
+    // Add reflexive equality for this term and all its subterms
+    add_subterm_reflexive_equalities(term_id, dag, &mut propositions);
+
+    Ok((term_id, propositions))
+}
+
+/// Add reflexive equalities for all subterms of a term
```
ezrosent: Add a sentence, like:

"If we have proved `(f (g 1)) = (f (g 1))` then we have also proved that `(g 1) = (g 1)`. " ?

src/proofs/proof_checker.rs
```diff
@@ -0,0 +1,926 @@
...
+/// without using globals (i.e., all global references are replaced with their definitions).
+///
+/// This expects the program to be in proof normalized form where globals appear as Let actions.
+pub(crate) fn gather_globals(
+    prog: &[ResolvedNCommand],
+    term_dag: &mut TermDag,
+) -> HashMap<String, TermId> {
+    let actions = gather_global_actions(prog);
+    let ctx = process_actions("global_action", HashMap::default(), &actions, term_dag)
+        .expect("Failed to process global actions");
```
ezrosent: should process_actions return a result as well? Or do we expect this to "never fail"

src/proofs/proof_checker.rs
```diff
@@ -0,0 +1,926 @@
...
+            Justification::MergeFn {
+                function,
+                old_proof,
+                new_proof,
+            } => {
+                // Check both sub-proofs - they should be reflexive proofs
+                let old_prop = self.check_proof_with_context(*old_proof, program, ctx)?;
+                let new_prop = self.check_proof_with_context(*new_proof, program, ctx)?;
+
+                let (_old_lhs, old_rhs) = (old_prop.lhs, old_prop.rhs);
```
ezrosent: Worth checking that lhs and rhs are the same here?

src/proofs/proof_encoding.md
```diff
@@ -0,0 +1,264 @@
...
+The job of the term encoding is to *remove all calls to union* in the egglog program.
+This makes proof production easier, since all equality reasoning is explicit and
+  can be instrumented with proof tracking.
+The term encoding adds an explicit union-find structure per sort, and maintains it via
+  rules that run during scheduled maintenance.
+For efficiency, every constructor becomes two tables:
+  a term table that stores the actual terms, and a view table storing representative terms along with their e-class (stored as the leader term).
+The term encoding enables proof tracking, done at the
+  same time in this file.
+The encoding keeps the operational semantics equivalent to the standard encoding (for the
```
ezrosent, line 14: Is this true? What operational semantics are you referring to here? I think what I want to say is that the _denotational_ model is the same. A lot of models of datalog are essentially "the set of facts in the database [after we reach a fixed point.]" I think what we want to say here is that if you take such a model of egglog with or without this encoding, there is a clear mapping that takes a fact in the 'native encoding' and maps it back to a fact in the 'term encoding.' Then within the 'term encoding' we can further say something about the corresponding proof. But that's very hand-wavy.

src/proofs/proof_encoding.md
```diff
@@ -0,0 +1,264 @@
...
+(sort view)
+(constructor Add (i64 i64) Math)
+(constructor AddView (i64 i64 Math) view)
+(constructor to_delete_Add (i64 i64) view)
+(constructor to_subsume_Add (i64 i64) view)
+```
+
+Each constructor in the original program is expanded to
+  a term table (`Add`), a view table (`AddView`), and helpers for deferred deletion/subsumption
+  (`to_delete_Add`, `to_subsume_Add`).
```
ezrosent, line 101: Needs an extra sentence introducing a concept of 'View'.

src/proofs/proof_encoding.md
```diff
@@ -0,0 +1,264 @@
...
+For each constructor, we add a congruence rule and a rebuild rule.
+The congruence rule adds equalities to the union-find table when two constructor applications
+  have equal arguments.
+The rebuild rule updates view tables so that views
+  point to representative terms for child e-classes.
+
+```text
+(function v2 () Math :no-merge)
+(set (v2) (Add 1 2))
+(AddView 1 2 (v2 ))
```
ezrosent: nit: `(v2)`

src/proofs/proof_encoding.md
```diff
@@ -0,0 +1,264 @@
...
+```
+
+Above is the desugaring for `(Add 1 2)`.
+We add to both view and term tables whenever we evaluate
+  a constructor or function application.
+Since global variables are not allowed after this pass,
```
ezrosent, line 138: Might be worth explaining in a separate short section what is going on with globals too.

src/proofs/proof_encoding.md
```diff
@@ -0,0 +1,264 @@
...
+We store the oldest proof currently.
+
+```text
+(function MathUFProof (Math Math) Proof :merge old)
+```
+
+Similarly, the union-find table gets a proof table storing
+  proofs of equalities between terms.
+If term `a` has parent `b`, it stores a 
+  proof of `a = b`.
```
ezrosent, line 267: Mention that the UF rules are instrumented to add 'Symmetry' and 'Transitivity' ?

src/proofs/proof_encoding.rs
ezrosent, line 1: Why rename from term encoding to proof encoding? Seems like term encoding is the more general concept and it includes both?

src/proofs/proof_encoding.rs
```diff
@@ -140,137 +44,153 @@ impl<'a> TermState<'a> {
...
-
-    fn subsumed_name(&self, name: &str) -> String {
-        format!("to_subsume_{}", name,)
-    }
-
-    fn delete_subsume_ruleset_name(&self) -> String {
-        "delete_subsume_ruleset".to_string()
-    }
-
+    /// Rules that execute deletion and subsumption based on the tables requesting the deletion/subsumption.
```
ezrosent, line 193: It'd be good to have a section in the markdown file for how `delete` and `subsume` work in the term encoding. I think if I was modifying this code it'd be pretty helpful to have that as a reference.

src/proofs/proof_encoding.rs
```diff
@@ -337,50 +261,57 @@ impl<'a> TermState<'a> {
...
                         {proof_query})
-                       ({merge_fn_code_str}
+                       (
+                        {merge_fn_code_str}
                         {rule_proof}
                         {term_and_proof}
-                        ({fresh_constructor} {merge_fn_var} old)
-                        ({fresh_constructor} {merge_fn_var} new)
+                        ({cleanup_constructor} {merge_fn_var} old)
+                        ({cleanup_constructor} {merge_fn_var} new)
```
ezrosent, line 304: Why are we cleaning up both old and new? What if the output of the merge fn is one of old or new? 

egglog currently runs all deletes before inserts if they land in the same iteration. Does deferring cleanups cause us to change that?

src/proofs/proof_encoding.rs
```diff
@@ -312,7 +234,7 @@ impl<'a> TermState<'a> {
             .map(|(i, _)| format!("c{i}_"))
             .collect::<Vec<_>>();
         let child_names_str = child_names.join(" ");
-        let rebuilding_ruleset = self.rebuilding_ruleset_name();
+        let rebuilding_ruleset = self.proof_names().rebuilding_ruleset_name.clone();
         let view_name = self.view_name(&fdecl.name);
         if fdecl.subtype == FunctionSubtype::Custom {
```
ezrosent: optional: This is a pretty big function... can we split it up?

src/proofs/proof_encoding.rs
```diff
@@ -392,13 +323,26 @@ impl<'a> TermState<'a> {
...
-            let uf_name = self.uf_name(&fdecl.schema.output);
+            let mut child_names_new = child_names.clone();
+            child_names_new.push("new".to_string());
+            let mut child_names_old = child_names.clone();
+            child_names_old.push("old".to_string());
+            let (query1, prf1) = self.query_view_and_get_proof(&fdecl.name, &child_names_new);
+            let (query2, prf2) = self.query_view_and_get_proof(&fdecl.name, &child_names_old);
+            let sym = &self.proof_names().eq_sym_constructor;
+            let trans = &self.proof_names().eq_trans_constructor;
+            let union_code = self.union(
```
ezrosent: Adding a comment on why this is transitivity would be helpful. It took me a while to internalize that a merge function basically involves two equalities

```
(f x) = y
(f x) = z
```

So the way we discover `y = z` is by glueing those two together (at opposite ends).

src/proofs/proof_encoding.rs
```diff
@@ -537,6 +415,9 @@ impl<'a> TermState<'a> {
         let view_proof_name = self.view_proof_name(&fdecl.name);
         let proof_type = self.proof_names().proof_datatype.clone();
 
+        // A view proof gives a proof that the representative term t_r equals the term in the view.
+        // Example: (AddView 2 3 t_r) proves the proposition that t_r = Add(2, 3) in that direction.
+        // This direction makes adding more proofs easier.
```
ezrosent: Why does it make it easier? (explain in the comment)

src/proofs/proof_encoding.rs
```diff
@@ -633,39 +515,171 @@ impl<'a> TermState<'a> {
...
+                            "
+                        );
+                    }
+
+                    proof
+                } else {
+                    "".to_string()
+                }
+            }
+            ResolvedFact::Eq(_span, generic_expr, generic_expr1) => {
```
ezrosent: nit: We can probably name these two exprs better

src/proofs/proof_encoding.rs
```diff
@@ -633,39 +515,171 @@ impl<'a> TermState<'a> {
...
+                        let (arg_str, proof) = self.instrument_fact_expr(arg, res);
+                        new_args.push(arg_str);
+                        arg_proofs.push(Some(proof));
+                    }
+                }
                 match resolved_call {
                     ResolvedCall::Func(func_type) => {
+                        assert!(
+                            func_type.subtype == FunctionSubtype::Constructor,
+                            "Only constructor function calls are allowed in fact expressions due to proof normal form. Got {:?}",
```
ezrosent: nit: Go {func_type:?}

src/proofs/proof_encoding.rs
```diff
@@ -633,39 +515,171 @@ impl<'a> TermState<'a> {
...
+                    } else {
+                        let (arg_str, proof) = self.instrument_fact_expr(arg, res);
+                        new_args.push(arg_str);
+                        arg_proofs.push(Some(proof));
+                    }
+                }
                 match resolved_call {
                     ResolvedCall::Func(func_type) => {
+                        assert!(
+                            func_type.subtype == FunctionSubtype::Constructor,
```
ezrosent, line 698: Worth making some of these Errors rather than asserts?

src/proofs/proof_encoding.rs
```diff
@@ -677,24 +691,37 @@ impl<'a> TermState<'a> {
...
+                        };
+
+                        (fv.clone(), proof)
                     }
                 }
             }
         }
     }
 
     // Returns a query, variable for the updated child, and proof
```
ezrosent: nit: `///` 

Give more context here. This returns three strings. What do those strings mean? I think they are mostly lines to splice into a query or action?


Same question for the next few methods returning strings.

src/proofs/proof_encoding.rs
```diff
@@ -785,103 +817,146 @@ impl<'a> TermState<'a> {
         res
     }
 
-    /// Return some code adding to the view and term tables,
-    /// returning a let-bound variable for the created term.
-    /// Expects the term arguments as `args`.
+    /// Update the view with  the given arguments.
```
ezrosent: extra space

src/proofs/proof_encoding.rs
```diff
@@ -785,103 +817,146 @@ impl<'a> TermState<'a> {
...
+            let proof_name = self.view_proof_name(fname);
+            res.push(format!(
+                "(set ({proof_name} {}) {proof})",
+                ListDisplay(args, " ")
+            ));
+        }
+        res.join("\n")
+    }
+
+    /// Return some code adding to the view and term tables.
```
ezrosent, line 862: What do .0 and .1 mean?

src/proofs/proof_encoding.rs
```diff
@@ -785,103 +817,146 @@ impl<'a> TermState<'a> {
...
+    /// For constructors, the arguments do not include the eclass of the resulting term (since it may not exist yet).
```
ezrosent: If this is true... what is `fv` doing? Isn't it the eclass?

src/proofs/proof_encoding_helpers.rs
```diff
@@ -0,0 +1,535 @@
...
+            view_name: HashMap::default(),
+            to_delete_name: HashMap::default(),
+            subsumed_name: HashMap::default(),
+            view_proof_name: HashMap::default(),
+            term_proof_name: HashMap::default(),
+        }
+    }
+}
+
+impl<'a> ProofInstrumentor<'a> {
```
ezrosent: we should be able to do `impl ProofInstrumentor<'_>` if we aren't using `'a` anywhere.

(This may be a `clippy` lint, I recommend running that if you aren't already).

src/proofs/proof_encoding_helpers.rs
```diff
@@ -0,0 +1,535 @@
...
+            fiat_constructor,
+            rule_constructor,
+            merge_fn_constructor,
+            eq_trans_constructor,
+            eq_sym_constructor,
+            congr_constructor,
+            pcons,
+            pnil,
+        ) = {
+            let names = self.proof_names();
```
ezrosent: I think we have a struct here with named fields. Can this be 

```
let EncodedNames { proof_list_sort, etc. } = self.proof_names(); 

```
?

src/proofs/proof_extraction.rs
```diff
@@ -0,0 +1,143 @@
...
+    PrimitivesUnsupported,
+    #[error("constructor {name} is marked :unextractable")]
+    ConstructorUnextractable { name: String },
+    #[error("prove requires proofs mode")]
+    ProofsDisabled,
+    #[error("Could not find a proof due to query not matching (constructor {constructor}).")]
+    QueryDidNotMatch { constructor: String },
+}
+
+impl<'a> ProofInstrumentor<'a> {
```
ezrosent: Same comment about `<'_>`

src/proofs/proof_format.rs
```diff
@@ -0,0 +1,749 @@
+use crate::{
+    ResolvedCall, Term, TermDag, TermId,
+    ast::{FunctionSubtype, ResolvedExpr, ResolvedFact, ResolvedNCommand},
+    proofs::proof_encoding_helpers::EncodingNames,
+    typechecking::FuncType,
+    util::{HEntry, HashMap, IndexSet, SymbolGen},
+};
+use egglog_ast::generic_ast::Literal;
+
+pub type ProofId = usize;
```
ezrosent: Can / should you use the `egglog-numeric-id` crate here? Provides more type-safety and helpful stuff like `DenseIdMap`

src/proofs/proof_format.rs
```diff
@@ -0,0 +1,749 @@
...
+/// leaves off the implicit rule substitution.
+/// Converting to a [`Proof`] with [`ProofStore::from_raw`] fills in these details.
+#[derive(Clone, PartialEq, Eq, Hash, Debug)]
+pub enum RawProof {
+    /// Equalities added at the top level are justified by fiat.
+    Fiat(TermId, TermId),
+    /// Given a rule name and proofs for each premise, produces a proof of a grounded equality t1 = t2 from the body of the rule.
+    /// The subsitution is implicit- in [`ProofTerm`] they are explicit.
+    Rule(String, Vec<ProofId>, TermId, TermId),
+    /// Given two proofs f(c1, c2, ..., old) = f(c1, c2, ..., old) and f(c1, c2, ..., new) = f(c1, c2, ..., new) and a term t produces a proof
```
ezrosent, line 58: (Feel free to resolve)

Seems like we're losing type safety here with having everything be an equality proof?

src/proofs/proof_format.rs
```diff
@@ -0,0 +1,749 @@
...
+    Congr(ProofId, usize, ProofId),
+}
+
+/// A [`ProofStore`] is similar to a [`TermDag`].
+/// It's a hash-consed arena enabling proofs to share sub-proofs.
+/// We refer to proofs with a [`ProofId`] which is an index into the store, used with [`ProofStore::get`] to retrieve the proof.
+#[derive(Clone, Debug)]
+pub struct ProofStore {
+    pub(super) term_dag: TermDag,
+    proof_id: HashMap<RawProof, ProofId>,
```
ezrosent, line 77: Are these ProofIds the same as the one used by RawProofStore?

src/proofs/proof_format.rs
```diff
@@ -0,0 +1,749 @@
...
+/// A proof shows that a [`Proposition`] is true, justified by a [`Justification`].
+#[derive(Clone, Debug)]
+pub struct Proof {
+    pub(super) proposition: Proposition,
+    pub(super) justification: Justification,
+}
+
+/// Justifies a [`Proposition`] using one of several proof rules.
+/// Some justifications are axioms of egglog, like Sym, Trans, and Congr.
+/// Other justifications are based on user input, like Fiat, Rule, and MergeFn.
```
ezrosent, line 118: Explain the relationship between Justification and RawProof (they are pretty similar)

src/proofs/proof_format.rs
```diff
@@ -0,0 +1,749 @@
...
+                ResolvedExpr::Call(
+                    _span2,
+                    head @ ResolvedCall::Func(FuncType {
+                        subtype: FunctionSubtype::Custom,
+                        ..
+                    }),
+                    args,
+                ),
+                // TODO this could actually be arbitrary pretty easily, it's just nested functions that are hard.
+                // We should also allow a function call with no bound output.
```
ezrosent: I think this TODO is repeated verbatim elsewhere... clean it up?

src/proofs/proof_format.rs
```diff
@@ -0,0 +1,749 @@
...
+                ResolvedExpr::Var(_span3, v),
+            ) => {
+                let term = proof.rhs();
+                let children = match self.term_dag.get(term) {
+                    Term::App(head_name, children) if head_name == head.name() => children.clone(),
+                    _ => panic!("expected function application term in proof rhs"),
```
ezrosent, line 518: Lots of comments along these lines... but what was the rationale for having this be a panic vs. a Error?

src/proofs/proof_normal_form.rs
```diff
@@ -0,0 +1,181 @@
...
+use egglog_ast::generic_ast::GenericExpr;
+
+/// Transforms queries so that they are in "proof normal form" by lifting subexpressions out, making them top level. In proof normal form:
+/// 1. Function calls like (lower-bound a b) are always top level and look like this:
+///    (= (lower-bound a b) c)
+/// 2. Primitives can't have constructors or function calls as arguments.
+///    For example, (!= a (Const 0)) becomes:
+///    (= (Const 0) v), (!= a v)
+///
+/// Nested function calls like this are not allowed:
```
ezrosent: let's say "flattened" not "not allowed". They _are_ allowed in `prog` they're just removed in this pass.

src/proofs/proof_simplification.rs
```diff
@@ -0,0 +1,349 @@
...
+                let sym_left_id = self.add_proof(sym_left);
+
+                // Create Trans(Sym(p2), Sym(p1)): c = a
+                let new_trans = Proof {
+                    proposition: Proposition::new(right_rhs, left_lhs),
+                    justification: Justification::Trans(sym_right_id, sym_left_id),
+                };
+
+                // Replace current proof
+                self.id_to_proof[proof_id] = new_trans;
```
ezrosent, line 216: Why do you need to explicitly remap here? Other optimizations don't seem to be doing it.

src/sort/add_primitive/src/lib.rs
```diff
@@ -4,6 +4,65 @@ use syn::parse::{Parse, ParseStream, Result};
 use syn::punctuated::Punctuated;
 use syn::{Expr, Ident, LitStr, Token, braced, bracketed, parenthesized, parse_macro_input};
 
+/// This macro lets the user declare custom egglog primitives with explicit validators.
+/// It combines primitive registration with a custom validator function.
```
ezrosent, line 8: Explain what a validator function is?

src/sort/add_primitive/src/lib.rs
```diff
@@ -4,6 +4,65 @@ use syn::parse::{Parse, ParseStream, Result};
...
+///     |termdag, lhs| {
+///         // Custom validation logic
+///         None
+///     }
+/// );
+/// ```
+#[proc_macro]
+pub fn add_primitive_with_validator(input: TokenStream) -> TokenStream {
+    // Convert to string to do manual parsing
+    let input_str = input.to_string();
```
ezrosent: Can you explain why we can't use `syn` to step through here, parse the arguments and then finally the custom validator if it's there?

src/sort/add_primitive/src/lib.rs
ezrosent: I guess I am curious what a validator really looks like... kinda hard to review without that.

ezrosent (changes requested):
First round of comments.

src/typechecking.rs
```diff
@@ -49,9 +49,17 @@ impl Hash for FuncType {
         }
     }
 }
+/// Primitive validators
+/// Validators take a termdag, arguments, and expected term id and return true if the computation is correct.
```
ezrosent: 'returns true' ? Not sure how this connects to the type signature. It seems like they return the expected output?

src/typechecking.rs
```diff
@@ -49,9 +49,17 @@ impl Hash for FuncType {
...
+pub type PrimitiveValidator = Arc<dyn Fn(&mut TermDag, &[TermId]) -> Option<TermId> + Send + Sync>;
 
 #[derive(Clone)]
-pub struct PrimitiveWithId(pub Arc<dyn Primitive + Send + Sync>, pub ExternalFunctionId);
+pub struct PrimitiveWithId {
+    pub(crate) primitive: Arc<dyn Primitive + Send + Sync>,
+    pub(crate) id: ExternalFunctionId,
+    // TODO is storing the validator here inefficient?
+    pub(crate) validator: Option<PrimitiveValidator>,
```
ezrosent, line 61: Should be okay? Worst-case we could stop generating validators when proofs aren't enabled.

tests/merge-saturates.egg
```diff
@@ -28,22 +28,25 @@
...
-      (baz x)
-      (< x 50)
-) (
-      (set (bar) (+ x 1))
-))
-
-(run 100)
-(check (baz 50))
\ No newline at end of file
+;; TODO now that relations are desugared, this trick also doesn't work:
```
ezrosent: What's the issue? (Could be a good topic for a regular meeting if this hasn't already come up?)

src/ast/desugar.rs
oflatt: Users should be able to write `prove` commands
What `proof_testing` does is turn `check` commands into `prove` commands.

src/sort/add_primitive/src/lib.rs
oflatt: See `validator_test` for an example (:

src/proofs/proof_checker.rs
oflatt: nice idea

src/proofs/proof_checker.rs
oflatt: good catch

src/proofs/proof_encoding.md
oflatt: Hmm I see what you are saying, but the goal is that from the user's perspective the term encoding and the non-term encoding are observationally equivalent (unless you use the prove command)

Maybe I just don't know what operational semantics means

src/proofs/proof_encoding.md
oflatt: good idea

src/proofs/proof_encoding.rs
oflatt: I thought it would be easier to find. Most people want to find the code related to proofs, which is the point of all this.

src/proofs/proof_encoding.rs
oflatt: Added to the running example in the markdown file

src/proofs/proof_encoding.rs
oflatt: The "cleanup_constructor" marks things as potentially needing to be cleaned up. This is because we don't actually know which one we need to cleanup.

I think this does the same thing as before but I could be wrong. During rebuilding we inspect this deferred cleanup constructor and delete things.

src/proofs/proof_encoding.rs
oflatt: good idea

src/proofs/proof_encoding.rs
oflatt: great idea

src/proofs/proof_encoding.rs
oflatt: I think this is an internal invariant deserving a panic because we enforce proof normal form right before this.

src/proofs/proof_encoding.rs
oflatt: Turns out inlining this makes the code much simpler, it was only used in one place.

tests/snapshots/files__towers_of_hanoi_term_encoding.snap
```diff
@@ -0,0 +1,17 @@
+---
+source: tests/files.rs
+expression: snapshot_content
+---
+(
+   (Config (Cons 1 (Cons 2 (Cons 3 (e)))) (e) (e) 0) -> (Config (Cons 1 (Cons 2 (Cons 3 (e)))) (e) (e) 0)
```
oflatt: See https://github.com/egraphs-good/egglog/issues/797

Now that relations are constructors, we get this

src/proofs/proof_encoding.rs
oflatt: I meant that the `args` passed in should not include `fv`, since we make it here. 
I'll clarify

src/proofs/proof_encoding_helpers.rs
oflatt: `add_term_encoding` uses `'a`

src/proofs/proof_encoding_helpers.rs
oflatt: Nice

src/proofs/proof_format.rs
oflatt: Yes!

src/proofs/proof_format.rs
oflatt: Yeah

src/proofs/proof_format.rs
oflatt: Good point, split into `RawProofId` and `ProofId`

src/proofs/proof_format.rs
oflatt: Good point
Inside the proof checker I've been not assuming that proofs are valid.

Inside proof computation I've been assuming that they are. This code computes rule substitutions and panics if it doesn't match up with the rule. 
Maybe we should start adding PANIC SAFETY comments like they do in Cedar, or expand the string errors to explain better.

src/proofs/proof_simplification.rs
oflatt: Other optimizations have an existing valid proof id, but this one has to compute a new `Proof` struct.

src/sort/add_primitive/src/lib.rs
oflatt: Actually that works, thanks!

src/typechecking.rs
oflatt: Sorry, old comment. Fixed.

tests/merge-saturates.egg
oflatt: Yeah, good topic for regular meeting. It doesn't work because the relation has a fresh sort, so you can't put it in the merge function position.

src/proofs/proof_encoding.md
ezrosent: I think OS + some observational equivalence relation would be fine here too... but then I think we should say something informal here that conveys that given that we don't have an "official" operational semantics for egglog. How about:


> The encoding does not change any user-observable behavior of an egglog program, except for commands that let you 'pop the hood' like `print-function`.

src/proofs/proof_extraction.rs
ezrosent: Did we need `'a` somethere here?

src/proofs/proof_format.rs
ezrosent: I think that'd be a good goal, though it would be a bit of a yak shave. Maybe this is an area where AI could speed us up, not sure.

src/sort/add_primitive/src/lib.rs
ezrosent: I still think a sentence like

> Validator functions "replay" a primitive without access to a full e-graph. Instead, they are given a TermDag with TermIds for each of the arguments. Validators are used when checking proofs of equivalences built during the execution of an egglog program.

would help

tests/files.rs
```diff
@@ -151,35 +180,33 @@ impl Run {
...
-    #[cfg(not(debug_assertions))]
     fn should_skip_snapshot(&self) -> bool {
-        // Skip tests with known non-deterministic output
-        let filename = self.path.file_stem().unwrap().to_string_lossy();
-        const SKIP_PATTERNS: [&str; 3] = [
-            "extract-vec-bench",
-            "python_array_optimize",
-            "stresstest_large_expr",
-        ];
+        // in parallel mode always skip
```
ezrosent, line 184: Instead, I'd recommend exposing a function from core-relations that checked rayon::current_num_threads and such. We can still run things in parallel when it's a release build. And if we ever changed the `debug_assertions` bit we won't have to hunt down every file that asked about it (most of the heuristics for parallelism are in one file I think).

ezrosent (approved):
I'd also see if you can track down the codspeed slowdowns... but they're small enough that they could be within the noise threshold.

tests/files.rs
oflatt: We already had parallel mode stuff in `files.rs` that checks this flag, seems like a separate issue

src/proofs/proof_extraction.rs
oflatt: Oh hmm, I guess not. Good catch, I thought we used it

src/proofs/proof_encoding.md
oflatt: I think we want print-function to also not change! It's just not true right now

PR #773, opened 2025-12-31
Avoid excessive cross-crate monomorphization

yihozhang:
This halves build time of egglog main crate from 28s to 12s on my m2 macbook under release mode. The artifact size also goes down from 10M to 7.8M.

### DIFF
--- a/core-relations/src/free_join/mod.rs
+++ b/core-relations/src/free_join/mod.rs
@@ -193,61 +193,55 @@ pub fn make_external_func<
     Wrapped(f)
 }
 
-pub(crate) trait ExternalFunctionExt: ExternalFunction {
-    /// A vectorized variant of `invoke` to avoid repeated dynamic dispatch.
-    ///
-    /// Implementors should not override this manually (in fact, they shouldn't
-    /// even be able to; some types are private); the default implementation
-    /// delegates core logic to `invoke`.
-    #[doc(hidden)]
-    fn invoke_batch(
-        &self,
-        state: &mut ExecutionState,
-        mask: &mut Mask,
-        bindings: &mut Bindings,
-        args: &[QueryEntry],
-        out_var: Variable,
-    ) {
-        let pool: Pool<Vec<Value>> = with_pool_set(|ps| ps.get_pool().clone());
-        let mut out = pool.get();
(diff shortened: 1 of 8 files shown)


core-relations/src/free_join/mod.rs
```diff
@@ -193,61 +193,55 @@ pub fn make_external_func<
...
+            this.invoke(state, args.as_slice())
         });
-        bindings.replace(out);
-    }
+    });
+    bindings.insert(out_var, &out);
 }
 
-impl<T: ExternalFunction> ExternalFunctionExt for T {}
+/// A variant of [`ExternalFunctionExt::invoke_batch`] that overwrites the output variable,
```
oflatt: Documentation needs updating. Shouldn't clippy docs have caught this invalid reference? I noticed we don't error on some doc warnings. We should turn that on.

oflatt:
Interesting, do you know why it's slow to compile the old version? Duplicates these methods for each external function maybe?

core-relations/src/free_join/mod.rs
```diff
@@ -193,61 +193,55 @@ pub fn make_external_func<
...
-        bindings: &mut Bindings,
-        args: &[QueryEntry],
-        out_var: Variable,
-    ) {
-        let mut out = bindings.take(out_var).expect("out_var must be bound");
-        for_each_binding_with_mask!(mask, args, bindings, |iter| {
-            iter.assign_vec_and_retain(&mut out.vals, |_, args| self.invoke(state, &args))
+/// A vectorized variant of `invoke` to avoid repeated dynamic dispatch.
+///
+/// Implementors should not override this manually (in fact, they shouldn't
```
oflatt: People can't override this anymore right?

oflatt (approved):
Nice!

ezrosent:
I _think_ this PR broke compilation for a bunch of unit tests. are we not running all tests in CI?

ezrosent:
Also FYI: this particular instance of cross-crate monomorphization was intentional: I wanted to get the tight loop compiled down for each external function. I guess none of our benchmarks actually benefit from this, but we may want to revisit in the future for the next big eval.

PR #775, opened 2026-01-02
Use bitset instead of Vec<Option<T>> for DenseIdMap

yihozhang:
Draft PR to see if it actually leads to speedup


yihozhang:
The typical map size is probably too small for this optimization to kick in. Closing for now...

yihozhang:
Okay I'm out of ideas. Closing again

PR #777, opened 2026-01-02
Remove duplicate variables using functional dependency

yihozhang:


src/core.rs
```diff
@@ -955,6 +969,65 @@ where
     }
 }
 
+impl<Head, Leaf> GenericCoreRule<Head, Head, Leaf>
+where
+    Leaf: Eq + Clone + Hash + Debug,
+    Head: Clone + Eq + Hash,
+{
+    pub(crate) fn remove_dup_vars(
```
oflatt, line 1007: Is this the kind of optimization that we should do in the backend crate instead of here?

oflatt, line 1007: Doc comment for this would be appreciated

src/core.rs
```diff
@@ -955,6 +969,65 @@ where
...
+        // Remove entries wit identical (head, inputs) pair and mark respective outputs to be merged.
+        self.body.atoms.retain(|atom| {
+            let (out, inp) = atom.args.split_last().unwrap();
+            let key = (atom.head.clone(), inp.to_owned());
+            let group = groups.entry(key).or_default();
+            group.push(out.clone());
+            group.len() == 1
+        });
+
+        let mut new_atoms = vec![];
```
oflatt: Could factor out into a generic helper that deduplicates variables

src/core.rs
```diff
@@ -955,6 +969,65 @@ where
...
+impl<Head, Leaf> GenericCoreRule<Head, Head, Leaf>
+where
+    Leaf: Eq + Clone + Hash + Debug,
+    Head: Clone + Eq + Hash,
+{
+    pub(crate) fn remove_dup_vars(
+        mut self,
+        value_eq: impl Fn(&GenericAtomTerm<Leaf>, &GenericAtomTerm<Leaf>) -> Head,
+    ) -> Self {
+        let mut groups: HashMap<(Head, Vec<GenericAtomTerm<Leaf>>), Vec<GenericAtomTerm<Leaf>>> =
```
oflatt, line 1012: Comment: maps function calls to sets of equivalent variables to be deduplicated

src/core.rs
```diff
@@ -1027,8 +1100,136 @@ impl ResolvedRuleExt for ResolvedRule {
         // `(rule ((= x y)) ((R x y)))`) but unboundedness is only checked during type checking.
         grounded_check(&rule)?;
 
-        let rule = rule.canonicalize(value_eq);
+        let rule = rule.canonicalize(&value_eq);
```
oflatt, line 1121: remove_dup_vars already calls canonicalize, maybe we should just call remove dup vars here? Or rename it to canon_fixpoint

oflatt (changes requested):
Nice, had some cleanup suggestions

oflatt:
This probably happens after resolving commands right? It should be fine. Proof encoding will happen just after typechecking

src/core.rs
yihozhang: I think this is easier here because the backend would have other columns like timestamp, subsumed etc.

src/core.rs
yihozhang: I'd like to keep them separate, since `canonicalize` is a necessary step in the compilation pipeline, while `remove_dup_vars` is an optional optimization.

oflatt (approved):

PR #778, opened 2026-01-04
Add snapshot tests

yihozhang:
Currently, our egglog program tests rely on `check`, and commands like `print-size` are not checked. This PR tries to add snapshots for the stdout of these programs.

`math-microbenchmark` under the term encoding is currently non-deterministically producing databases of different sizes. I don't know if this is a bug.

Some commands like `print-function` and `extract` can have different behaviors depending on e.g., which matches get found first. In the future we can consider an opt-in version of these commands that are deterministic (at the cost of being more expensive).


ezrosent:
I'm not sure about this... egglog doesn't really guarantee a deterministic database output depending on the external functions you use. And in the presence of parallelism the database is only equivalent "up to isomorphism" or some such (e.g. eclass ids may not be unique between two runs in parallel, even without any weird functions).

I think it'd be better to leverage the new proofs infrastructure to check proofs of assertions and ensure they're valid.  (we could even add validation that checks proofs for a random term in the database). This is a more end-to-end semantic guarantee compared to the alternative IMO, and won't break if we add more "benign" nondeterminism in the future.

yihozhang:
I am thinking having deterministic versions of commands like `print-function` printing alphabetically-sorted tuples, which should be the same even with different parallel setups. 

Also, currently, the snapshot tests are only enabled for release builds, so they are always run in a single thread.

ezrosent:
For alphabetically sorted tuples... either they're flat (in which case e-class ids won't be deterministic) or they're deep and contain a whole term. Seems like we're doing the latter. But then, can't the terms be exponential in size?

As for not running in parallel... I assume we'll want to change that at some point, right?

saulshanabrook (approved):
I think this PR makes sense. I had been thinking for a while about adding snapshots for the whole e-graph as an SVG, but this is probably a better start.

Even if it can't capture everything and some things are non-deterministic, most of the examples should be deterministic and this is a nice way to catch more behavior and also write tests more easily using extract than having to write them with rust code.

PR #780, opened 2026-01-05
Refactor query planning

yihozhang:
This PR

* Breaks down the planner into an immutable context and a mutable state, and makes code more modular at a few places.
* Removes unnecessary references
* Fixes a minor bug that can happen if directly using core-relations (when using free join to plan R(x, x), S(x, x), T(x, x), the JoinStage will contain only one variable but a `filter` can have two columns)

### DIFF
--- a/core-relations/src/free_join/plan.rs
+++ b/core-relations/src/free_join/plan.rs
@@ -432,7 +432,13 @@ fn plan_headers(
     ctx: &PlanningContext,
 ) -> (
     Vec<JoinHeader>,
-    DenseIdMap<AtomId, (usize, &Pooled<Vec<Constraint>>)>,
+    DenseIdMap<
+        AtomId,
+        (
+            usize, /* The approx size of the subset matching the constraints. */
+            &Pooled<Vec<Constraint>>,
+        ),
+    >,
 ) {
     let mut header = Vec::new();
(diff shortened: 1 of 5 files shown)


core-relations/src/free_join/plan.rs
```diff
@@ -669,3 +425,295 @@ impl<'a> BucketQueue<'a> {
...
+/// Build join headers from fast constraints and compute remaining constraints for planning.
+/// Returns (headers, remaining_constraints) tuple.
+fn plan_headers(
+    ctx: &PlanningContext,
+) -> (
+    Vec<JoinHeader>,
+    DenseIdMap<AtomId, (usize, &Pooled<Vec<Constraint>>)>,
+) {
+    let mut header = Vec::new();
+    let mut remaining_constraints: DenseIdMap<AtomId, (usize, &Pooled<Vec<Constraint>>)> =
```
ezrosent, line 444: nit: explain what the usize means

ezrosent (approved):
> Fixes a minor bug that can happen if directly using core-relations (when using free join to plan R(x, x), S(x, x), T(x, x), the JoinStage will contain only one variable but a filter can have two columns)

Thanks for fixing this! This shouldn't be too hard to write a test for, I think. Could you add one?

PR #783, opened 2026-01-06
Add new multiset primitives

saulshanabrook:
Adds new multiset primitives as used in https://uwplse.org/2026/02/24/egglog-containers.html and https://github.com/egraphs-good/egglog-python/pull/397 to try avoiding A/C blowup.

This also:

* Changes how higher order primitive functions are registered, so that either the container sort or the function sort can be registered first.
* Exposes some additional structures on the egraph to support "freezing" it, which is exposed in Python: https://github.com/egraphs-good/egglog-python/blob/bf94c2ec4d95ec9835be8826cb9d2e8d6e50765f/src/freeze.rs This is used for debugging to see what e-classes are together. Alternatively, I could keep these private and expose a frozen e-graph directly here.
* Add the `uf` table to be a merge function dep so that we can say use `vec-union` as a merge function (fix recommended by @ezrosent)
* Adds ability to lookup if a function is a let binding so that the Python bindings can expose this.
* Allow nested container types

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -987,8 +987,19 @@ impl EGraph {
 
     /// Read the contents of the given function.
     /// The callback f is called with each row and its subsumption status.
-    pub fn function_for_each(&self, func: &Function, f: impl FnMut(FunctionRow<'_>)) {
+    ///
+    /// Raises an error if the function does not exist.
+    pub fn function_for_each(
+        &self,
+        func_name: &str,
+        f: impl FnMut(FunctionRow<'_>),
+    ) -> Result<(), Error> {
+        let func = self
+            .functions
+            .get(func_name)
+            .ok_or_else(|| TypeError::UnboundFunction(func_name.to_string(), span!()))?;
         self.backend.for_each(func.backend_id, f);
+        Ok(())
     }
 
     /// Evaluates an expression, returns the sort of the expression and the evaluation result.


src/sort/multiset.rs
```diff
@@ -116,34 +128,73 @@ impl ContainerSort for MultiSetSort {
...
+                name: "multiset-sum-multisets".into(),
+                multiset: other_multiset_sort.clone(),
+                multiset_of_multisets: arc.clone(),
+            });
+        }
+        let all_ms_sorts = eg
+            .type_info
+            .get_arcsorts_by(|f| f.value_type() == Some(TypeId::of::<MultiSetContainer>()));
+        for fn_sort in eg.type_info.get_sorts::<FunctionSort>() {
+            for ms_sort in &all_ms_sorts {
```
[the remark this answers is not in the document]
saulshanabrook: We don't keep this index currently, so for now yeah it will be slow but hopefully manageable if the number of sorts isn't too large

src/sort/multiset.rs
```diff
@@ -162,18 +213,356 @@ impl ContainerSort for MultiSetSort {
...
+        for (v, c) in multiset.data.iter_counts() {
+            let mut row = vec![args[0], v];
+            // If we have already filled this multiset once, skip since it should still be accurate with the right
+            // merge function
+            if action.lookup(exec_state, &row).is_some() {
+                break;
+            }
+            row.push(exec_state.base_values().get::<i64>(c.try_into().unwrap()));
+            action.insert(exec_state, row.into_iter());
+        }
```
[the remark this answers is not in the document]
saulshanabrook: The multiset shouldn't ever be modified...

egglog-bridge/src/lib.rs
```diff
@@ -1216,6 +1216,7 @@ impl MergeFn {
             Primitive(_, args) => {
                 args.iter()
                     .for_each(|arg| arg.fill_deps(egraph, read_deps, write_deps));
+                write_deps.insert(egraph.uf_table);
             }
```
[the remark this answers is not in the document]
saulshanabrook: @ezrosent do you think its worth adding this fine grained handling of the dependency here as it suggests or just keep it as is?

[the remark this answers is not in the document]
ezrosent: I think we should leave it as-is and only make it fine-grained if it becomes a problem. Write dependencies are also more "flexible" than read dependencies.

src/lib.rs
```diff
@@ -975,6 +980,17 @@ impl EGraph {
         }
     }
 
+    /// Get the list of all functions in the e-graph.
+    pub fn get_function_names(&self) -> Vec<String> {
+        self.functions.keys().cloned().collect()
+    }
+
+    /// Read the contents of the given function.
+    /// The callback f is called with each row and its subsumption status.
```
oflatt, line 989: I don't think "row" is something egglog users really know about. Maybe we should say for each e-node in the case of a constructor, and each entry in the case of a function?

Also, would it be easier for users to just use a function id here or a string name for the function?

Finally I still wonder if we should make this a method on ExecutionState and give a helper to grab the ExecutionState. That way you can still use this helper in the `rust_rule` context.

saulshanabrook: > I don't think "row" is something egglog users really know about. Maybe we should say for each e-node in the case of a constructor, and each entry in the case of a function?


Just noting for future reference that the docs for `print_function` method say "Extract rows of a table using the default cost model with name sym". So I am not clear if we have a shared understanding of whether "row" is a word we should/shouldn't use publicly.

saulshanabrook: > Also, would it be easier for users to just use a function id here or a string name for the function?

Changed to take a function name. the `backend_id` is not currently public on the function.

saulshanabrook: > Finally I still wonder if we should make this a method on ExecutionState and give a helper to grab the ExecutionState. That way you can still use this helper in the rust_rule context.

I am open to trying to do this if this is the consensus on how we should best expose things. I think it would also lead to some other things to keep this consistant:

1. Move `for_each` and `for_each_while` from `egglog_bridge::EGraph` into `core_relations::ExecutionState`
2. Update all usages of those to now use `ExecutionState` (including serialization and extraction) instead of `egglog_bridge::EGraph`.

Makefile
```diff
@@ -28,7 +28,8 @@ nits:
 fixnits:
 	@rustup component add rustfmt
 	cargo fmt
-	cargo clippy --fix --workspace --allow-dirty
+	@rustup component add rustfmt
```
yihozhang, line 31:
```suggestion
	@rustup component add clippy
```

src/lib.rs
yihozhang: FWIW `FunctionRow` is also exposed in the cost model.

yihozhang (commented):
I think the reason it is okay to have nested containers currently is because we are rebuilding the containers in the order they are defined: https://github.com/egraphs-good/egglog/blob/388046734e525b54b04cc9391ba37846cb7c3e7f/core-relations/src/containers/mod.rs#L148 So naturally we rebuild children container before rebuilding parent.

However, I think this is no longer true in parallel mode, where all containers are built in parallel. Could you check if the tests are still passing when it's run in parallel mode and `parallelize_inter_container_op` is set to always be true?

src/lib.rs
saulshanabrook: After doing a bit more research, I am not sure exactly how to proceed here:

1. Adding `for_each` to `ExecutionState` would make it tied to bridge specific metadata it depends on bridge specific metadata for things like proof mode and subsumption and function IDs.
2. The TableAction in bridge does support these and we could add it to that instead and expose that at the top level. However, that is not supported in proof mode, since it also always mutations. So moving extraction, etc to use this would break proof mode.
3. We could add a another abstraction for view only tables in bridge and use that everywhere. However, I don't think that would get us to a goal of a unified goal inside actions and at the top level.

Overall, I think I am not quite sure if there is an easy local solution here that doesn't require a broader re-architecture of Rust APIs and safety. At the top level you can do full scans which break semi-naive and inside of actions/rules we might not want to allow this? Either way, it seems like rethinking about our internal/external APIs is tied up in large discussions arounds safety with semi naive (and potentially also proof support, for write ops).

saulshanabrook: Just chatted with @oflatt, came to a plan of merging this for now. Then when we figure out what we want the public API to be with execution state and table action, we can switch to that. If we don't want to break things, we can mark this method as deprecated and move people to new API.

PR #784, opened 2026-01-09
Try avoiding allocations during execution

yihozhang:
Just trying to see if this matters. Don't merge yet.


yihozhang:
Stale and most are already fixed

PR #785, opened 2026-01-12
Tree decomposition

yihozhang:
This PR does the following things

* Add tree decomposition to the query optimizer
* Use functional dependency information during query planning
* Add InPlaceMaterializer and ScopedMaterializer for materializing intermediate bags
* Add two benchmarks from hardboiled and a rectangle example


yihozhang:
@ezrosent This PR is ready for review! Some notes

* I added two tests from the hardboiled benchmarks. conv1d_32 goes from 0.7s to 1.1s, and conv1d_128 goes from 40s to 7s. Both the slowdown and the speedup are from the same rules, interestingly.
* I spot-checked the flamegraph of proof-testing-typecheck, and about 4% slowdown is because of the tree decomposition during query planning, and 2% is from rule evaluation.
* There are a few eggcc benchmarks that are slower with this PR because of a particular rule (not uploaded). I think I'll fix this in a future PR.
* Some diffs in this PR are due to the Rust toolchain upgrade. `cargo` refused to install nextest locally and complained that the toolchain is too old.
* There's a TODO left for generating a report for the visualizer for the decomposed plan. This does not affect the usability of this PR so I'll leave it to a future PR (maybe Claude can help me with that)

numeric-id/src/lib.rs
```diff
@@ -174,6 +174,14 @@ impl<K: NumericId, V> DenseIdMap<K, V> {
             .filter_map(|(i, v)| Some((K::from_usize(i), v.as_mut()?)))
     }
 
+    #[allow(clippy::should_implement_trait)]
```
yihozhang, line 177: Implementing IntoIterator requires me to explicitly type the expression, which I can't do because of the closure

yihozhang:
Okay I found my old heuristics for sort_plan_by_size is bad. I tweaked it, and now it's faster across the board, besides proofs.

core-relations/src/free_join/execute.rs
```diff
@@ -156,43 +161,155 @@ impl Database {
...
+                                            subset.as_ref(),
+                                            &with_pool_set(|ps| ps.get_pool()),
+                                        );
+                                        if cur.subset.is_empty() {
+                                            break 'eval;
+                                        }
+                                        binding_info.move_back_node(*atom, cur);
+                                    }
 
+                                    let mut materializations: DenseIdMap<
```
ezrosent, line 238: There may be some low hanging fruit here around representing these materializations and eliminating small allocations. 

Probably can wait to a follow-up?

core-relations/src/free_join/execute.rs
```diff
@@ -912,6 +1127,142 @@ impl<'a> JoinState<'a> {
...
+                            return false;
+                        }
+                    }
+                    true
+                };
+
+                match mode {
+                    MatScanMode::Full | MatScanMode::KeyOnly => {
+                        // enumerate keys
+                        // for group in cover_mat.iter() {
```
ezrosent: stale commented-out line?

core-relations/src/free_join/plan.rs
```diff
@@ -163,11 +182,71 @@ impl JoinStage {
...
+#[derive(Debug, Clone)]
+pub(crate) struct JoinStages {
+    pub instrs: Arc<Vec<JoinStage>>,
+}
+
+/// Specification of the materialization of the intermediate results, as required by tree decomposition.
+#[derive(Debug, Clone)]
+pub(crate) struct MatSpec {
+    pub msg_vars: Vec<Variable>,
+    pub val_vars: Vec<Variable>,
```
ezrosent: document what message vs. val variables mean

ezrosent: We should also consider SmallVec here, depending on the common sizes that we see.

core-relations/src/free_join/plan.rs
```diff
@@ -288,22 +369,668 @@ pub enum PlanStrategy {
     Gj,
 }
 
+// Computes the next variable to eliminate and the subquery of its neighborhood.
```
ezrosent: `///` ?

core-relations/src/free_join/plan.rs
```diff
@@ -288,22 +369,668 @@ pub enum PlanStrategy {
...
+    }
+
+    // Step 2: Sort bags topologically and merge leafy bags with their parents
+    let mut bags = topologically_sort_bags(bags);
+
+    if bags.len() <= 1 {
+        return fast_path!();
+    }
+
+    // Step 3: Count variable usage across bags. Used for deciding if a variable is public (i.e., messaege variables) or private.
```
ezrosent: typo: message

core-relations/src/free_join/plan.rs
```diff
@@ -288,22 +369,668 @@ pub enum PlanStrategy {
...
+        })
+        .min_by_key(|a| a.0)
+        .map(|a| (a.1, a.2))?;
+    Some(IndexSet::from_iter(
+        subquery_vars.into_iter().map(|(var, _)| var),
+    ))
+}
+
+/// It updates the hypergraph with the given bag of variables by:
+/// 1. Remove atoms that only contain subquery variables and remove variable occurrences of those atoms, and
```
ezrosent: What are 'subquery variables'

core-relations/src/free_join/plan.rs
ezrosent, line 1: This file needs a big doc comment. (It needed one before... but it definitely needs one now)

ezrosent (changes requested):
First pass, mostly superficial comments.Still trying to understand the decomposition stuff.

My initial reaction is that the materializations could potentially be a lot more efficient. I'm wondering if you've tried using a SortedWritesTable for them?

yihozhang:
@ezrosent This PR is ready for another look

ezrosent (approved):

Issue #671, opened 2025-09-10
match_limit and time_limit when running a query


yihozhang:
This somehow falls through the cracks, but one thing we talked about is to support early stopping of the matching step, when certain (time or match) limits are exceeded. This has uses in scheduling. Egg has this functionality, but supporting this in egglog is harder because our backend is parallel.

Assigning @ezrosent for now

PR #788, opened 2026-01-17
Add initial early stopping support and use it for panic functions

ezrosent:
This change adds a simple global "early stop" flag to ExecutionState that can be used to short-circuit rule execution. This doesn't let you get match limits or node limits on their own, but it is a useful primitive towards that feature.

Relevant to #671

### DIFF
--- a/core-relations/src/action/mod.rs
+++ b/core-relations/src/action/mod.rs
@@ -2,7 +2,13 @@
 //!
 //! This allows us to execute the "right-hand-side" of a rule. The
 //! implementation here is optimized to execute on a batch of rows at a time.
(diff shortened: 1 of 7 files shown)


saulshanabrook (approved):
Looks good, thanks!


As an extension of this, you could allow an arbitrary function to be called to return a boolean if we should stop or not... In Python for a while I have been wanting to support the user being able to stop egglog while its going, like handling CTRL C... To do this, you have to add a check signals call throughout any long running processes, to see if the user tried to cancel the process: https://docs.rs/pyo3/latest/pyo3/marker/struct.Python.html#method.check_signals

PR #789, opened 2026-01-18
Refactor TermDag API to use Ids and add pretty printer for proofs

oflatt:
This PR makes the termdag API use ids, simplifying code (especially incoming proof code)
It also adds a pretty printer which is good for printing terms with intermediate let bindings. It's used for proofs in the incoming PR.

### DIFF
--- a/src/termdag.rs
+++ b/src/termdag.rs
@@ -59,11 +59,20 @@ impl RenderedTerm {
     }
 }
 
+/// Context used during term rendering with let-binding support.
 struct TermRenderContext<'a> {
+    /// Generator for fresh variable names used in let bindings.
     fresh: &'a mut SymbolGen,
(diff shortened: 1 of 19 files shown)


src/termdag.rs
```diff
@@ -35,6 +40,61 @@ macro_rules! match_term_app {
...
+
+    fn is_multiline(&self) -> bool {
+        self.pretty.contains('\n')
+    }
+}
+
+struct TermRenderContext<'a> {
+    fresh: &'a mut SymbolGen,
+    ref_counts: &'a HashMap<TermId, usize>,
+    sizes: &'a HashMap<TermId, usize>,
```
ezrosent, line 71: document these fields (in particular what is the "unit" of size here and what are the refcounts counting?)

ezrosent (approved):
Looks good! Thanks for splitting this one out.

PR #790, opened 2026-01-19
Add support for code coverage

saulshanabrook:
This PR adds support for running the tests locally to produce code coverage and in CI uploading to Codecov.

This should allow us to see for each PR how it affects the overall code coverage and what added lines are covered by tests/examples. I have been finding myself wanting this, as I develop new features to make sure I covered all the cases with examples.

I don't suggest we require 100% coverage for new lines added. However, even just seeing the lines can help you understand if your tests are covering what you think they are and possibly encourage covering more lines.

It can also help see what parts of our code base are not used in any examples.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,9 +14,18 @@ jobs:
       - uses: actions/checkout@v3
       - uses: taiki-e/install-action@v2
         with:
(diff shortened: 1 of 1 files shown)


oflatt:
I'm in favor of coverage! We could do an 80% threshold or something like that

Issue #769, opened 2025-12-27
Rewrite tests to use globals with $ prefix


oflatt:
Our tests trigger the warning about `$` prefix on globals

PR #792, opened 2026-01-23
rename global vars in tests

marko-subotic:
Working on #769. I wrote a script to parse a file and add the `$` prefix to global variables that don't have it, another to use this to format all files, and third to verify that all files don't have compiler warnings.

I left the scripts off the commit to avoid cluttering the repo with one-off scripts, but if they might be used in the future to format computer generated files, I can add them.


tests/fail-typecheck/shadow-local-global.egg
```diff
@@ -1,8 +1,8 @@
...
   (Wrap i64))
 
-(let a (Wrap 1))
+(let $a (Wrap 1))
 
 
 (rule ((Wrap 1))
-      ((let a (Wrap 1))))
\ No newline at end of file
+      ((let $a (Wrap 1))))
```
oflatt, line 8: Probably we should also have a variant of this test where this line has `(let a (Wrap 1))` which should also fail.

oflatt: While we are at it, if the rule comes before the global it should be allowed

marko-subotic: When I enable strict mode, all of the `term_encoding` tests fail. One way to fix this (I think) is to change the `INTERNAL_PREFIX` variable from `@` to `$@`, that seems dirty because I have no idea if they're always supposed to be global.

oflatt: Oh I see. I'd say let's just disable strict mode for term encoding tests for now. 
In the future I'll make the term encoding work by adding the global prefix when it generates names.

global_var_formatter/batch_formatter.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3
```
oflatt, line 1: Maybe this folder should be in a `dev_utilities` folder for clarity?

Issue #793, opened 2026-01-23
Nondeterminism in math-microbenchmark with term encoding


oflatt:
#778 reveals that with the term encoding, math-microbenchmark has nondeterministic output in the database size.

I think that we intend to guarantee deterministic output for egglog in single-threaded mode. We should investigate this and just as importantly **document egglog's guarantees.**

oflatt:
My conclusion: If one day we find a way to deterministically iterate over an e-graph even though the engine is nondeterministic, that would be cool

Similarly (but a stronger property), it would be nice if the extraction algorithm was more stable and chose the same result even when you upgrade egglog 

For now we should make egglog deterministic in sequential mode

oflatt:
New nodeterminism problem:
https://github.com/egraphs-good/egglog/pull/824

The eggcc benchmark is getting different counts each time!

Issue #331, opened 2024-01-24
Desugaring global variables


oflatt:
In the past, we desugared global variables to instead use zero-argument functions as an encoding trick. However, it was ad-hoc and introduced shadowing between local variables and global functions.
After getting rid of this, we changed the back-end to support global variables directly, but this was also messy.

Yihong and I propose a new pass that takes care of this problem my doing the old encoding trick in a more principled way.

oflatt:
fixed by #338

PR #802, opened 2026-02-05
Fix snapshot tests

saulshanabrook:
Currently snapshots tests (added in #778) were sometimes failing depending on the ordering of print table (#793).

This PR fixes this by removing print table from the snapshots for now.

It also fixes some other failing tests that were added with the global/local change (#331), that for some reason were not caught before.

Finally, it updates the readme to mention snapshot tests and separates the updating of snapshots from running the tests.


saulshanabrook:
Closing this for now, since it seems like the print table tests are now stable

PR #805, opened 2026-02-10
Extraction for proof mode, improved snapshot testing

oflatt:
This PR adds:

- Working `extract` in proof mode
- A `:uf` annotation for sorts to specify the union-find for that sort. This is leveraged by extraction to do a single look-up before extracting an e-class. 
```
(sort UFSort)
(sort Expr :uf UF_Expr)
(constructor UF_Expr (Expr Expr) UFSort)

(constructor Foo () Expr)
(constructor Bar () Expr)

; Create two different terms
(let $t1 (Foo))
(let $t2 (Bar))

; Set up UF entries: $t2 points to $t1 as the canonical representative
; This makes $t1 the canonical rep for both
(UF_Expr $t2 $t1)

; Now extract $t2 - it should use UF_Expr to find that $t1 is the canonical rep
; So we expect to see (Foo), not (Bar)
(extract $t2)
```

- Better testing across different treatments. We test that we get the same output from proof mode and non-proof mode, and also test we get the right error message from fail-typechecking tests


ezrosent:
Why do we need the term annotation (i.e. why is it not sufficient to just not union these terms when we translate them to the term encoding?)

src/ast/desugar.rs
```diff
@@ -216,22 +242,27 @@ fn desugar_prove(parser: &mut Parser, span: Span, query: Vec<Fact>) -> Vec<NComm
...
-        }))
-        .collect()
+    vec![NCommand::Sort {
+        span: span.clone(),
+        name: name.clone(),
+        presort_and_args: None,
+        uf: None,
+        unionable: true,
+    }]
+    .into_iter()
```
ezrosent, line 252: nit: this can be iter::once I think

src/ast/mod.rs
```diff
@@ -49,11 +49,18 @@ where
...
+    Sort {
+        span: Span,
+        name: String,
+        presort_and_args: Option<(String, Vec<GenericExpr<String, String>>)>,
+        /// The name of the union-find function for this sort.
+        /// Used in term encoding to canonicalize values during extraction.
+        uf: Option<String>,
+        /// Whether values of this sort can be unioned.
+        /// Defaults to true for user-defined sorts.
+        /// Set to false for relations and term tables that should not allow union.
```
ezrosent, line 61: Still a little unsure why we need a new concept of `unionable` here... seems analogous to the `:term` annotation that got removed.

src/ast/parse.rs
```diff
@@ -311,11 +347,16 @@ impl Parser {
...
-                        [] => {}
-                        [(":unextractable", [])] => unextractable = true,
-                        [(":cost", [c])] => cost = Some(c.expect_uint("cost")?),
-                        _ => return error!(span, "could not parse constructor options"),
+                    let mut term_constructor = None;
+                    for (key, val) in self.parse_options(rest)? {
+                        match (key, val) {
+                            (":unextractable", []) => unextractable = true,
+                            (":cost", [c]) => cost = Some(c.expect_uint("cost")?),
+                            (":term-constructor", [tc]) => {
```
ezrosent: Should we have similar documentation for `:term-constructor` and `:unextractable` etc. in this code, like the one you added for "sort" above?

src/proofs/proof_encoding.rs
```diff
@@ -429,12 +429,25 @@ impl<'a> ProofInstrumentor<'a> {
         let merge_rule = self.handle_merge_or_congruence(fdecl);
         // the term table has child_sorts as inputs
         // the view table has child_sorts + the leader term for the eclass
+        // Propagate cost and unextractable flags from the original function
+        let mut term_flags = String::new();
```
ezrosent, line 433: optional: if you want to avoid the `&format!(...)` stuff you can make these byte vectors and then use `write!` which will run the formatting stuff in-place. Then you can convert to String using from_utf8

src/proofs/proof_encoding_helpers.rs
```diff
@@ -488,34 +538,76 @@ pub(crate) fn command_supports_proof_encoding(
...
-        return false;
+        return Err(ProofEncodingUnsupportedReason::PrimitiveWithoutValidator);
+    }
+
+    // Check actions (not queries) for function lookups
+    // Egglog supports lookups in actions at the global level, but not in proofs mode
+    // (global function calls are allowed - they get desugared to constructors)
+    let mut has_function_lookup_in_action = false;
+    command.clone().visit_actions(&mut |action| {
+        if action_has_function_lookup(&action, type_info) {
```
ezrosent: nit, optional: `has_function_lookup_in_action |= action_has_function_lookup(&action, type_info)`

src/extract.rs
```diff
@@ -149,24 +150,57 @@ impl<C: Cost + Ord + Eq + Clone + Debug> Extractor<C> {
...
+    ) -> Self {
+        Self::compute_costs_from_rootsorts_internal(rootsorts, egraph, cost_model, false, true)
+    }
+
+    fn compute_costs_from_rootsorts_internal(
+        rootsorts: Option<Vec<ArcSort>>,
+        egraph: &EGraph,
+        cost_model: impl CostModel<C> + 'static,
+        respect_unextractable: bool,
+        skip_view_tables: bool,
```
ezrosent: optional: I think code with multiple bool flags like this one can become more readable with structs, basically giving you named arguments:

```rust
struct ExtractionOptions<CM> {
cost_model: CM,
respect_unextractable: bool,
skip_view_tables: bool,
}
```

Or similar. Up to you

src/extract.rs
```diff
@@ -533,6 +582,34 @@ impl<C: Cost + Ord + Eq + Clone + Debug> Extractor<C> {
         )
     }
 
+    /// Find the canonical representative of a value using the union-find table.
+    /// If no UF is registered for this sort, returns the original value.
+    /// The UF table stores (value, canonical) pairs - one hop lookup.
+    fn find_canonical(&self, egraph: &EGraph, value: Value, sort: &ArcSort) -> Value {
```
ezrosent, line 617: Similar point to the copilot comment. I think we should just do one scan and cache a HashMap rather than scanning the UF.

src/typechecking.rs
```diff
@@ -779,6 +794,29 @@ impl TypeInfo {
...
     }
+
+    /// Check if an expression contains non-global function lookups (FunctionSubtype::Custom calls).
+    /// Global function calls are allowed since they get desugared to constructors.
+    /// Returns Some(span) if a lookup is found, None otherwise.
+    pub fn expr_has_function_lookup(&self, expr: &ResolvedExpr) -> Option<Span> {
+        use ast::GenericExpr;
+
+        let mut lookup_span = None;
+        expr.walk(
```
ezrosent: for stuff like this I wonder if we should support early stopping? e.g. pass a function from GenericExpr -> Option<T>, then stop the walk as soon as one is found?

tests/snapshots/files__points_to_proof_testing.snap
```diff
@@ -0,0 +1,179 @@
...
+  (Rule
+    prop0
+    (name
+      "(rule ((= x (New a b)))
+      ((VarPointsTo a b))
+         )")
+    (premises (Fiat (= t1 t1)))
+    (substitution (a "o1") (x t1) (b (Class "A"))))
+  (Fiat (= (Class "A") (Class "A")))
+  1)(let t0 (VarPointsTo "o2" (Class "B")))
```
ezrosent: nit: missing a newline somewhere?

tests/snapshots/files__print_function_desugar.snap
```diff
@@ -2,25 +2,4 @@
...
-   (G (F 2) "2") -> 2
-   (G (F 3) "3") -> 3
-   (G (F 4) "4") -> 4
-)
-(
-   (G (F 1) "1") -> 1
-   (G (F 2) "2") -> 2
-   (G (F 3) "3") -> 3
-)
+
```
ezrosent: can you explain why this and other ones are now empty?

tests/files.rs
```diff
@@ -25,18 +28,41 @@ impl Run {
...
                 other => Some(other.to_string()),
             })
             .collect::<Vec<_>>()
             .join("")
     }
+
+    /// Extraction results may differ slightly due to the proof encoding when multiple
+    /// solutions have the same cost. We filter the output to include only things that remain the same.
+    #[allow(dead_code)]
+    fn outputs_to_snapshot_preserved_across_treatments(&self, outputs: &[CommandOutput]) -> String {
```
ezrosent, line 26: why is this `[allow(dead_code)]` ? If it's only used in tests should it be `[cfg(test)]`?

ezrosent (approved):
Mostly minor comments!

src/ast/mod.rs
oflatt: Good question, we actually had `unionable` already for relations but we stored it on constructors which doesn't make sense, it should be on sorts. It's not exposed to users, just used in desugaring of relation.

yihozhang (changes requested):
The proofs seem very big and not yet merged to main line (unlike what we thought during the meeting). Should we not include them?

The new snapshot files have suffixes `proofs`, `term_encoding`, and `proof_testing`. It seems to me that `proofs` and `term_encoding` snapshots should be identical to the stock snapshot, and `proof_encoding` snapshots may oscillate across commits (like print-function), so we should have none of the snapshots?

src/extract.rs
oflatt: We just call this once per extraction so it's still one scan either way. I think if we cache a hashmap we have to keep track of if it is up to date

src/typechecking.rs
oflatt: Early stopping made it kinda complicated so I implemented a new `find` function on exprs instead.

tests/snapshots/files__print_function_desugar.snap
oflatt: Print function was showing the same nondeterminism bug, disabled for now. Will look into it.

tests/files.rs
oflatt: oops good catch

oflatt:
Deleted all the proof snapshots and all the non-shared ones

src/extract.rs
oflatt: good idea

src/ast/mod.rs
```diff
@@ -49,11 +49,18 @@ where
...
+        unionable: bool,
```
yihozhang, line 62: optional: it can't be the case that uf is Some while unionable is false. A cleaner approach is to represent both fields in an enum.

src/constraint.rs
```diff
@@ -667,6 +667,9 @@ impl Assignment<AtomTerm, ArcSort> {
                 if !sort.is_eq_sort() {
                     return Err(TypeError::NonEqsortUnion(sort, span.clone()));
                 }
+                if !typeinfo.is_sort_unionable(&sort) {
```
yihozhang, line 670: How about `union` in the form of a `set`? (e.g., instead of (union (a) (b)) do a (set (a) (b))).

src/lib.rs
```diff
@@ -287,6 +289,51 @@ impl Function {
     pub fn can_subsume(&self) -> bool {
         self.can_subsume
     }
+
+    /// For view tables (with term_constructor), the effective output sort is the last input column.
+    /// For regular tables, it's the output sort.
+    /// This is used by extraction to determine which sort a table produces values for.
+    pub(crate) fn extraction_output_sort(&self) -> &ArcSort {
```
yihozhang: nit: move these EGraph impl to the extraction file?

src/prelude.rs
```diff
@@ -306,16 +306,14 @@ pub fn rule(
     facts: Facts<String, String>,
     actions: Actions,
 ) -> Result<Vec<CommandOutput>, Error> {
-    let mut rule = Rule {
```
yihozhang, line 309: Is it an issue that rules can have the same names? `EGraph::add_rule` has a check that all rules of a ruleset must have distinct names.

tests/fail-typecheck/union_relation.egg
```diff
@@ -0,0 +1,10 @@
+; Test that union fails on relations
+; Relations are desugared with unionable: false, so they don't allow union
+
+(relation edge (i64 i64))
+
+(edge 1 2)
+(edge 3 4)
+
+; This should fail - cannot union values of a non-unionable sort (relation)
+(union (edge 1 2) (edge 3 4))
```
yihozhang, line 10: Could you also add a test checking for (set (edge 1 2) (edge 3 4))?

Also, we are still allowing rules like the following, right?

```
(rule
((= (edge 1 2) (edge 3 4)))
())
```

Is this an issue?

tests/fail-typecheck/union_term_constructor.egg
```diff
@@ -0,0 +1,11 @@
...
+; The :nounion annotation marks a sort as non-unionable
+
+(sort TermSort :nounion)
+(constructor term-fn (i64 i64) TermSort)
+
+(term-fn 1 2)
+(term-fn 3 4)
+
+; This should fail - cannot union values of a non-unionable sort
+(union (term-fn 1 2) (term-fn 3 4))
```
yihozhang: Same as the above comment.

tests/snapshots/files__fail-typecheck__arity_mismatch.snap
```diff
@@ -0,0 +1,15 @@
+---
```
yihozhang, line 1: nit: This failure message could become non-deterministic

tests/snapshots/files__shared_snapshot_primitive_args.snap
```diff
@@ -1,13 +1,7 @@
 ---
 source: tests/files.rs
-expression: snapshot
```
yihozhang, line 3: What kind of proofs are checked in currently?

tests/snapshots/files__shared_snapshot_prims.snap
```diff
@@ -0,0 +1,5 @@
+---
```
yihozhang: We don't need to generate the snapshot if the output is empty.

tests/files.rs
```diff
@@ -60,21 +70,42 @@ impl Run {
...
-                .iter()
-                .any(|o| !matches!(o, CommandOutput::RunSchedule(..)))
-        {
-            let snapshot_name = self.name().to_string();
-            let snapshot_content = self.outputs_to_snapshot(&_outputs);
-            insta::assert_snapshot!(snapshot_name, snapshot_content);
+        if !self.should_skip_snapshot() {
+            match &result {
+                Ok(outputs)
+                    if outputs
```
yihozhang: nit: We can simplify the logic here a bit by first converting the output to a string (outputs_to_snapshot_preserved_across_treatments), and then asserting the snapshot only when the string is not empty.

Currently, we are filtering both here and inside `outputs_to_snapshot_preserved_across_treatments`, and the filtering here at line 78 does not work, as empty snapshot files still get generated.

yihozhang (changes requested):

src/constraint.rs
oflatt: Remember that's disallowed by surface egglog, so we would get a type error if we tried to run the output

tests/fail-typecheck/union_relation.egg
oflatt: You're right, this is currently allowed but now has different semantics- the rule will never match

oflatt: Unrelated to this PR, since we introduced desugaring of relation separately. Will make new test in another PR

src/constraint.rs
oflatt: I guess it's not currently disallowed but should be

src/ast/mod.rs
oflatt: I think it could be, it would just be inefficient. In fact right now the proof generation does add a UF

tests/snapshots/files__shared_snapshot_primitive_args.snap
oflatt: Just the ones that require proofs in the proofs folder

tests/snapshots/files__fail-typecheck__arity_mismatch.snap
oflatt: Hmm how should I fix this? Shouldn't we just make sure it's deterministic?

src/prelude.rs
oflatt: I think desugaring already gives it a new name when the name is empty, so this code was redundant.

tests/files.rs
oflatt: good idea

tests/fail-typecheck/union_term_constructor.egg
oflatt: will fix in follow on

tests/snapshots/files__fail-typecheck__arity_mismatch.snap
yihozhang: Maybe we can enforce an order on the errors. Let's revisit if this becomes an issue.

yihozhang (approved):
Looks good. Thank you!

PR #807, opened 2026-02-12
Prepare for 2.0.0 release

yihozhang:
I cut the 2.0.0 release, which in particular does not include #770. This is the changelog from the 2.0.0 release.


oflatt (approved):
Thanks!

Issue #744, opened 2025-11-21
Crash seemingly due to unextractable terms.


pavpanchekha:
When running the following file, with no command-line flags:

[bug.txt](https://github.com/user-attachments/files/23668136/bug.txt)

```
$ egglog bug.txt
...
[INFO ] extracted variants:

thread 'main' panicked at src/actions.rs:360:29:
assertion failed: !terms.is_empty()
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Not sure what's going on and I might have generated the code wrong but I feel like in either case it probably shouldn't panic? I'm guessing that there's no nodes in the extracted eclass due to unextractable terms.

oflatt:
Yeah, the current extract panics on failure. We should probably have a variant that gives an result

oflatt:
Maybe something like
(extract (myterm) :onfail (mydefault))

Or maybe
(extract-fallible ...) 
And the text interface returns (Ok ...) or (Error ...)

saulshanabrook:
I think we could also just return an error from extraction and not panic? And still exit egglog. It seems like this would be a good behavior all the time.

saulshanabrook:
So are you saying that you want the program to continue after the extraction is unable to return a term but print a warning or error?

oflatt:
I'm 100% with Pavel on this one, it's been bugging me too

saulshanabrook:
I am trying to reproduce this bug, but on main and in the demo this does not panic. The script as written exits successfully and prints `()`. This makes sense to me, since extract with an integer will try to extract up to the max if available.

If the last line is changed to `(extract (const28))` then it will exit with an error code and log:

```
[ERROR] Unextractable root Value(12) with sort EqSort { name: "MTy" }
[ERROR] extraction failure: "Unable to find any valid extraction (likely due to subsume or delete)"
```

I am wondering if there are remaining items in this issue here that need to be dealt with.

PR #808, opened 2026-02-12
Print-size preserve behavior in desugaring and proof mode

oflatt:
This PR adds

- :internal-hidden, a new flag on constructors and functions for hiding them from print-size
- :internal-let, a new flag to mark things as let bindings so they show up in visualizations but are also hidden from print-size
- Disabling let bindings from appearing in print-size


tests/snapshots/files__taylor51_desugar.snap
saulshanabrook, line 1: Talked about adding this to the skip list along with any other files with huge files.

tests/snapshots/files__repro_new_backend_python_vec_desugar.snap
saulshanabrook, line 1: We also talked about removing the desugared snapshot

yihozhang (commented):
Could you make `print-size` prints S-expressions? I think we are all supportive of this idea in https://github.com/egraphs-good/egglog/issues/744.

tests/snapshots/files__taylor51_desugar.snap
oflatt: done

tests/snapshots/files__repro_new_backend_python_vec_desugar.snap
oflatt: done

src/ast/mod.rs
```diff
@@ -1181,9 +1218,12 @@ where
     pub merge: Option<GenericExpr<Head, Leaf>>,
     pub cost: Option<DefaultCost>,
     pub unextractable: bool,
+    /// Hidden functions are excluded from print-size output.
+    /// Used for internal tables generated by proof production.
+    pub hidden: bool,
     /// Globals are desugared to functions, with this flag set to true.
     /// This is used by visualization to handle globals differently.
-    pub let_binding: bool,
+    pub internal_let: bool,
```
saulshanabrook: hidden is passed in as `:internal-hidden` and let is passed in an `:internal-let`. Should either both of these in the struct be prefixed with `internal` or neither of them?

src/lib.rs
```diff
@@ -712,22 +719,38 @@ impl EGraph {
     }
 
     /// Print the size of a function. If no function name is provided,
-    /// print the size of all functions in "name: len" pairs.
+    /// print the size of all non-hidden functions in "name: len" pairs.
```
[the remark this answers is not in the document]
saulshanabrook: This is a good suggestion

src/lib.rs
```diff
@@ -148,10 +148,17 @@ impl std::fmt::Display for CommandOutput {
...
-                for name in names_and_sizes {
-                    writeln!(f, "{}: {}", name.0, name.1)?;
+                write!(f, "(")?;
+                for (i, (name, size)) in names_and_sizes.iter().enumerate() {
+                    if i > 0 {
+                        write!(f, " ")?;
+                    }
+                    write!(f, "({} {})", name, size)?;
+                    if i < names_and_sizes.len() - 1 {
+                        writeln!(f)?;
```
saulshanabrook, line 161: Could you add a comment here to describe the behavior. I can get a sense of what you are doing but its really hard to follow this code for some reason.

src/lib.rs
```diff
@@ -712,22 +719,38 @@ impl EGraph {
...
+    /// print the size of all non-hidden functions in "name: len" pairs.
     pub fn print_size(&self, sym: Option<&str>) -> Result<CommandOutput, Error> {
         if let Some(sym) = sym {
+            // First check if this is a term_constructor name that maps to a view table
             let f = self
                 .functions
-                .get(sym)
+                .values()
+                .find(|f| f.decl.term_constructor.as_deref() == Some(sym))
+                .or_else(|| self.functions.get(sym))
```
saulshanabrook, line 736: add comment about view tables, proof mode, when which branch fires and why we cant differentiate based on if we are in proof mode

src/lib.rs
```diff
@@ -712,22 +719,38 @@ impl EGraph {
...
                 .ok_or(TypeError::UnboundFunction(sym.to_owned(), span!()))?;
+            // Skip hidden and let_binding functions
+            if f.decl.hidden || f.decl.internal_let {
+                return Err(TypeError::UnboundFunction(sym.to_owned(), span!()).into());
+            }
```
saulshanabrook, line 741: We also now are disabling print-size on let bindings

src/typechecking.rs
```diff
@@ -212,7 +212,15 @@ impl EGraph {
 
         let command: ResolvedNCommand = match command {
             NCommand::Function(fdecl) => {
-                ResolvedNCommand::Function(self.type_info.typecheck_function(symbol_gen, fdecl)?)
+                let resolved = self.type_info.typecheck_function(symbol_gen, fdecl)?;
+                // If this is a let binding, add it to global_sorts
+                if resolved.internal_let {
+                    let output_sort = self.type_info.sorts.get(&fdecl.schema.output).unwrap();
```
saulshanabrook, line 219: Add comment this is for desugarring

tests/name-resolution.egg
```diff
@@ -22,4 +22,6 @@
 ))
       
       
-(run 10)
\ No newline at end of file
+(run 10)
+
+(print-size)
```
saulshanabrook:
```suggestion
```

saulshanabrook (approved):

src/lib.rs
```diff
@@ -712,22 +719,38 @@ impl EGraph {
...
     pub fn print_size(&self, sym: Option<&str>) -> Result<CommandOutput, Error> {
         if let Some(sym) = sym {
+            // First check if this is a term_constructor name that maps to a view table
             let f = self
                 .functions
-                .get(sym)
+                .values()
+                .find(|f| f.decl.term_constructor.as_deref() == Some(sym))
+                .or_else(|| self.functions.get(sym))
                 .ok_or(TypeError::UnboundFunction(sym.to_owned(), span!()))?;
```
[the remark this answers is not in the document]
oflatt: This is by design

PR #809, opened 2026-02-19
Disallow set on constructors

oflatt:
I thought we had this somewhere but we seem to have lost it. In the function/constructor split we decided `set` is for functions and `union` is for constructors.


tests/taylor51.egg
```diff
@@ -3367,3487 +3367,3487 @@
 (rule
  ()
  ((let ?sR (Var "R"))
-  (set (const1) ?sR)
+  (union (const1) ?sR)
```
saulshanabrook, line 3370: Do we know why herbie was omitting set here instead of union? Maybe we should check in before merging.

saulshanabrook (approved):

tests/taylor51.egg
oflatt: Unclear, will check with Noah

oflatt: no reason

PR #812, opened 2026-02-25
Roundtrip test for proofs, :proof-function annotation

oflatt:
This PR makes the roundtrip test for proofs work by making the proof encoding emit valid egglog.

Changes
- Add a :proof-function annotation so egglog knows which table stores the proof
- Fix a couple bugs in proof generation
- Tricky roundtrip test- we need to be able to check proofs with respect to the original program, not the desugared one. So there's a new endpoint which allows you to set the original program for checking.

### DIFF
--- a/src/proofs/proof_format.rs
+++ b/src/proofs/proof_format.rs
@@ -339,8 +339,10 @@ impl ProofStore {
             proof_id: HashMap::default(),
             id_to_proof: DenseIdMap::new(),
         };
+        let globals = gather_globals(prog, &mut store.term_dag)
+            .unwrap_or_else(|_| panic!("failed to gather globals from program"));
 
-        let proof_id = store.convert_raw_proof(prog, &raw_store, raw_proof_id);
+        let proof_id = store.convert_raw_proof(prog, &globals, &raw_store, raw_proof_id);
         (store, proof_id)
     }
 
@@ -351,12 +353,10 @@ impl ProofStore {
     fn convert_raw_proof(
         &mut self,
         prog: &Vec<ResolvedNCommand>,
+        globals: &HashMap<String, TermId>,
         raw_store: &RawProofStore,
         raw_proof_id: RawProofId,
     ) -> ProofId {
-        let globals = gather_globals(prog, &mut self.term_dag)
-            .unwrap_or_else(|_| panic!("failed to gather globals from program"));
-
         if let Some(&id) = self.proof_id.get(&raw_store.store[raw_proof_id.index()]) {
             return id;
         }
@@ -373,7 +373,7 @@ impl ProofStore {
             RawProof::Rule(name, premise_proofs, lhs, rhs) => {
                 let converted_premises: Vec<ProofId> = premise_proofs
                     .iter()
-                    .map(|pid| self.convert_raw_proof(prog, raw_store, *pid))
+                    .map(|pid| self.convert_raw_proof(prog, globals, raw_store, *pid))
                     .collect();
 
                 let mut substitution =
@@ -394,8 +394,8 @@ impl ProofStore {
                 }
             }
             RawProof::MergeFn(function, old_raw, new_raw, to_prove) => {
-                let old_proof_id = self.convert_raw_proof(prog, raw_store, *old_raw);
-                let new_proof_id = self.convert_raw_proof(prog, raw_store, *new_raw);
+                let old_proof_id = self.convert_raw_proof(prog, globals, raw_store, *old_raw);
+                let new_proof_id = self.convert_raw_proof(prog, globals, raw_store, *new_raw);
                 let to_prove = raw_store.unwrap_ast(*to_prove);
                 Proof {
                     proposition: Proposition::new(to_prove, to_prove),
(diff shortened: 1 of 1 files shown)


src/proofs/proof_format.rs
```diff
@@ -253,10 +251,10 @@ impl RawProofStore {
         let term = self.term_dag.get(list_term).clone();
         match term {
             Term::App(head, args) => {
-                if head == self.encoding_names.pnil {
+                if head.contains("PNil") {
                     assert!(args.is_empty(), "pnil should not have arguments");
                     Vec::new()
-                } else if head == self.encoding_names.pcons {
+                } else if head.contains("PCons") {
```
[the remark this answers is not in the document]
oflatt: this is intended so it's robust to desugaring

src/ast/mod.rs
```diff
@@ -937,9 +949,18 @@ where
...
+                proof_func,
                 ..
             } => {
-                write!(f, "(sort {name})")
+                write!(f, "(sort {name}")?;
+                if let Some(uf) = uf {
+                    write!(f, " :uf {uf}")?;
+                }
+                if let Some(pf) = proof_func {
+                    write!(f, " :proof-func {pf}")?;
```
saulshanabrook: Should these both be private like the other things we added? Do we for see a use case for users to directly add these or just during de-sugarring?

src/ast/mod.rs
```diff
@@ -1977,24 +2002,12 @@ where
     }
 }
 
-/// Sanitizes internal names so they do not contain any internal characters.
-/// This enables printing desugared egglog in a way that can be re-parsed.
```
saulshanabrook, line 1981: What are these sanitation changes for?

src/ast/parse.rs
```diff
@@ -278,32 +280,43 @@ impl Parser {
...
                             uf,
+                            proof_func,
                             unionable: true,
                         }]
                     }
                     _ => {
                         return error!(
                             span,
-                            "usages:\n(sort <name>)\n(sort <name> :uf <uf-function>)\n(sort <name> (<container sort> <argument sort>*))"
+                            "usages:\n(sort <name>)\n(sort <name> :uf <uf-function>)\n(sort <name> :proof-func <proof-func-name>)\n(sort <name> (<container sort> <argument sort>*))"
```
saulshanabrook: I would not include these in the docs if they are meant to just be private.

saulshanabrook (commented):
Will take another look soon, but just wanted to highlight if we should mark these attributes as private like the ones we added to functions.

src/ast/mod.rs
oflatt: Just refactoring, ended up not needing to extend sanitization

src/ast/mod.rs
oflatt: Yeah we can mark these as private!

saulshanabrook:
@coderabbitai full review

src/ast/mod.rs
```diff
@@ -2019,19 +2032,41 @@ where
...
+    Leaf: Clone + PartialEq + Eq + Display + Hash,
+{
+    // first convert to unresolved
+    let unresolved = program
+        .iter()
+        .map(|cmd| cmd.clone().make_unresolved())
+        .collect::<Vec<_>>();
+    let max_underscores = get_max_underscores(&unresolved);
+    let replacement = "_".repeat(max_underscores + 1);
+    replace_internal_symbol_with(&unresolved, &replacement)
```
saulshanabrook, line 2073: Could you add a little comment here to describe the algorithm of underscore replacing?

src/lib.rs
```diff
@@ -721,6 +744,27 @@ impl EGraph {
...
+        // make a new e-graph, desugar the program in proof mode
+        let mut proof_check_eg = EGraph::new_with_proofs();
+        if proof_testing {
+            proof_check_eg = proof_check_eg.with_proof_testing();
+        }
+        let resolved = proof_check_eg.process_program_internal(prog, false)?;
+
+        self.proof_check_program = resolved.resolved_before_proofs;
+        Ok(())
+    }
```
saulshanabrook, line 760: Could we just inline this into the tests instead of having it as a function? I think I just don't totally follow how you would use this.

saulshanabrook (commented):
I left a few more comments. I feel a little concerned about the overall complexity with de-sugarring now and proofs and testing, it was already a bit complicated. 

I don't think that should block this PR, just another item of debt that at some point might be worth re-thinking about it holistically to see if we can bring down complexity somehow...

src/lib.rs
oflatt: The problem is it uses process_program_internal which isn't public. Our tests use the public api in files.rs

Lmk if you have a nice way to fix this, this was my best idea. I already marked it as doc(hidden)

saulshanabrook: Sounds good for now! Thanks for the explanation

src/proofs/proof_format.rs
```diff
@@ -364,6 +362,9 @@ impl ProofStore {
         raw_store: &RawProofStore,
         raw_proof_id: RawProofId,
     ) -> ProofId {
+        let globals = gather_globals(prog, &mut self.term_dag)
+            .unwrap_or_else(|_| panic!("failed to gather globals from program"));
```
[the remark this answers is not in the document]
oflatt: this was a good find

PR #813, opened 2026-02-26
Proofs for extraction

oflatt:
- Adds a :internal-proof-function annotation so extraction knows which table is the proof function for a given constructor
- Adds a new `extract-with-proof` command in proof mode which gives a proof for an extraction.


src/proofs/prove_extraction.rs
```diff
@@ -0,0 +1,548 @@
...
+///
+/// See `proof_encoding.md` for full details on the term and proof encoding.
+pub(crate) struct ViewProofIndex {
+    /// For each original function/constructor name, maps the input columns of
+    /// its view table to the e-class representative (second-to-last value in
+    /// the row).
+    ///
+    /// Key: original function/constructor name (e.g. `"Add"`).
+    /// Value: `HashMap` from input values (`Vec<Value>`) to the e-class `Value`.
+    pub(crate) view_eclass: HashMap<String, HashMap<Vec<Value>, Value>>,
```
ezrosent, line 37: no change needed, jfyi: there are some tricks we have available in the backend to make this more efficient if you notice this becoming too expensive (lots of small allocations, etc)

src/proofs/prove_extraction.rs
```diff
@@ -0,0 +1,548 @@
...
+            // ----------------------------------------------------------
+            // Look for the corresponding proof table and build its map.
+            // ----------------------------------------------------------
+            let proof_table_name = match proof_names.view_proof_name.get(original_name) {
+                Some(name) => name,
+                None => continue,
+            };
+            let proof_function = match egraph.get_function(proof_table_name) {
+                Some(f) => f,
+                None => continue,
```
ezrosent: out of curiosity: do we ever expect this None branch to fire? (if not, worth a comment I think)

src/proofs/prove_extraction.rs
```diff
@@ -0,0 +1,548 @@
...
+            };
+            let proof_backend_id = proof_function.backend_id;
+
+            let mut proof_map: HashMap<Vec<Value>, Value> = HashMap::default();
+            egraph.backend.for_each(proof_backend_id, |row| {
+                let vals = row.vals;
+                if !vals.is_empty() {
+                    let key = vals[..vals.len() - 1].to_vec();
+                    let proof_value = vals[vals.len() - 1];
+                    proof_map.insert(key, proof_value);
```
ezrosent: maybe add a debug_assert! that the key was unique?

src/proofs/prove_extraction.rs
```diff
@@ -0,0 +1,548 @@
...
+                .egraph
+                .proof_state
+                .uf_parent
+                .get(&sort_name)
+                .ok_or_else(|| ExtractWithProofError::NoUfTable {
+                    sort_name: sort_name.clone(),
+                })?;
+            // The UF table is a constructor UF_Sort(child, parent) -> uf,
+            // so we scan for the row where vals[0] == value and read vals[1].
+            // If no entry exists the value IS the representative.
```
ezrosent: When does the case of "no entry exists" happen, in practice?

src/proofs/prove_extraction.rs
```diff
@@ -0,0 +1,548 @@
...
+            let uf_func = self.egraph.get_function(uf_table_name).ok_or_else(|| {
+                ExtractWithProofError::NoUfTable {
+                    sort_name: sort_name.clone(),
+                }
+            })?;
+            let mut canonical = value;
+            self.egraph
```
ezrosent: can we build an index for this mapping along with the other ones we build at the top?

ezrosent (approved):
Minor comments, a few not even requesting changes

src/proofs/prove_extraction.rs
oflatt: I changed it to a panic actually, that'll be less silent

src/proofs/prove_extraction.rs
oflatt: Actually the new invariant introduced in my other branch is that this never happens. I'll error out

src/proofs/prove_extraction.rs
oflatt: good idea

PR #817, opened 2026-03-04
Simplify rebuilding rule, proof runtime in CI with Codspeed

oflatt:
This PR generates a single rebuild rule that looks like this:

```
(rule ((AddView c0 c1 c2)
       (UF_Math c2 c2_leader)
       (filter
         (or (bool-!= c2 c2_leader))))
      ((AddView c0 c1 c2_leader)
       (delete (AddView c0 c1 c2)))
        :ruleset rebuilding :name "rebuild_rule")
```

Instead of the old rebuild rules generated which had one per child that needed to be rebuilt.

### DIFF
--- a/src/proofs/proof_encoding.md
+++ b/src/proofs/proof_encoding.md
@@ -94,6 +94,8 @@ This is because the rebuild rules query the union-find for every
   prevents the rule from firing even when other columns have changed.
 Self-loops are added in `add_term_and_view` whenever a constructor
   value is created.
+We may want to remove this invariant in the future if we move
+  to a different encoding, saving some space and time.
 
 
 ```text
@@ -121,7 +123,7 @@ The view tables are kept up to date during rebuilding.
        :ruleset rebuilding :name "congruence_rule")
 (rule ((AddView c0 c1 c2)
        (UF_Math c2 c2_leader)
-       (filter
+       (guard
          (or (bool-!= c2 c2_leader))))
       ((AddView c0 c1 c2_leader)
        (delete (AddView c0 c1 c2)))
--- a/src/proofs/proof_encoding.rs
+++ b/src/proofs/proof_encoding.rs
@@ -532,7 +532,7 @@ impl<'a> ProofInstrumentor<'a> {
 
         let uf_query_str = uf_queries.join("\n       ");
         let or_expr = format!("(or {})", bool_neq_exprs.join("\n             "));
-        let filter_query = format!("(filter {or_expr})");
+        let filter_query = format!("(guard {or_expr})");
 
         // Build the updated children: use leader_var for eq-sort columns, original for others.
         let children_updated: Vec<String> = leader_vars.clone();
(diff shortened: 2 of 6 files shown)


src/proofs/proof_encoding.md
```diff
@@ -87,6 +87,14 @@ We use the `ordering-max` and `ordering-min` egglog primitives
   so that we can deterministically choose which term becomes the parent
   in the union-find structure.
 
+**Important invariant:** every representative term must have a self-loop
+  entry in the union-find table (e.g., `(UF_Math v v)`).
+This is because the rebuild rules query the union-find for every
```
ezrosent, line 92: Let's make a note that if we ever move to a different encoding (e.g. based on functions, or similar) that we may want to remove the rules that add the self-loops

src/proofs/proof_encoding.rs
```diff
@@ -1203,6 +1231,8 @@ impl<'a> ProofInstrumentor<'a> {
                 for stmt in action_stmts {
                     res.extend(self.parse_program(&stmt));
                 }
+                // Rebuild before extract- we may have added new view rows that need canonicalization
```
ezrosent: mega-nit: semi-colon maybe?

src/sort/bool.rs
```diff
@@ -14,9 +14,23 @@ impl BaseSort for BoolSort {
...
+            }
+            Some(termdag.lit(Literal::Bool(result)))
+        };
+        add_primitive_with_validator!(eg, "or" = [xs: bool] -> bool {
+            xs.fold(false, |acc, x| acc || x)
+        }, or_validator);
         add_literal_prim!(eg, "xor" = |a: bool, b: bool| -> bool { a ^ b });
         add_literal_prim!(eg, "=>" = |a: bool, b: bool| -> bool { !a || b });
+        // A filter primitive that fails the query if the boolean is false
+        add_literal_prim!(eg, "filter" = |a: bool| -?> () { a.then_some(()) });
```
ezrosent: I don't think "filter" is super obvious syntax here... though maybe that's just because the difference between a predicate and a filter isn't super obvious.

Maybe we can call it "guard" or "require" ? No need to bikeshed, if you disagree feel free not to change.

src/lib.rs
```diff
@@ -381,6 +381,20 @@ impl Default for EGraph {
             neq_validator
         );
 
+        add_primitive_with_validator!(
+            &mut eg,
+            "bool-!=" = |a: #, b: #| -> bool {
+                (a != b)
+            },
+            |termdag: &mut TermDag, args: &[TermId]| -> Option<TermId> {
+                if args.len() == 2 {
```
ezrosent, line 390: nit, optional: (args.len() == 2).then(|| termdag.lit(...))

src/sort/bool.rs
oflatt: guard is better nice idea

PR #827, opened 2026-03-12
src/lib.rs: EGraph readonly pub. accessors

Muxianesty:
The following Pull Request introduces a number of read-only accessors to `EGraph`.
In order to preserve existing API no other function was changed.


oflatt:
Seems like this can be closed?

PR #829, opened 2026-03-16
Add index for union-find in proof mode

oflatt:
Optimize proof mode by adding a function index.

This makes bigger benchmarks dramatically faster, looks like it's slower for a couple small ones


ezrosent (approved):
Very nice!

Do we still need the __UF_X table or could we build everything on top of __UF_Xf?

PR #830, opened 2026-03-17
Rust rule benchmarking tableaction cache

MilkBlock:
cache TableAction for every RustRuleContext so that we could avoid copy every rule run.


MilkBlock:
@oflatt

src/prelude.rs
```diff
@@ -523,20 +620,28 @@ pub fn rust_rule(
 ) -> Result<Vec<CommandOutput>, Error> {
     let prim_name = egraph.parser.symbol_gen.fresh("rust_rule_prim");
     let panic_id = egraph.backend.new_panic(format!("{prim_name}_panic"));
+
+    let mut func_ids: HashMap<String, egglog_bridge::FunctionId> = Default::default();
```
yihozhang, line 624: `HashMap::with_capacity`?

src/prelude.rs
```diff
@@ -428,7 +523,9 @@ impl<F: Fn(&mut RustRuleContext, &[Value]) -> Option<()>> Primitive for RustRule
         let mut context = RustRuleContext {
             exec_state,
             union_action: self.union_action,
-            table_actions: self.table_actions.clone(),
+            func_ids: self.func_ids.as_ref(),
```
yihozhang, line 526: Do func_ids and table_actions_by_id need to be `Arc`-wrapped, if they are passed by reference?

src/prelude.rs
```diff
@@ -523,20 +620,28 @@ pub fn rust_rule(
 ) -> Result<Vec<CommandOutput>, Error> {
     let prim_name = egraph.parser.symbol_gen.fresh("rust_rule_prim");
     let panic_id = egraph.backend.new_panic(format!("{prim_name}_panic"));
+
+    let mut func_ids: HashMap<String, egglog_bridge::FunctionId> = Default::default();
+    func_ids.reserve(egraph.functions.len());
+    let mut table_actions_by_id: Vec<Option<egglog_bridge::TableAction>> = Vec::new();
```
yihozhang, line 626: nit: reserve the vec once (by scanning egraph.functions) instead of repeatedly resizing

src/prelude.rs
```diff
@@ -354,14 +360,59 @@ impl RustRuleContext<'_, '_> {
...
+        table_actions_by_id: &[Option<egglog_bridge::TableAction>],
+        table: egglog_bridge::FunctionId,
+    ) -> &egglog_bridge::TableAction {
+        table_actions_by_id
+            .get(table.rep() as usize)
+            .and_then(|x| x.as_ref())
+            .unwrap_or_else(|| panic!("missing table action for FunctionId({})", table.rep()))
+    }
+
+    fn get_table_action_cache_by_id<'a>(
```
yihozhang, line 380: I don't understand this function? Is this caching the lookups to a hashmap with a (Id, TableAction) vector, because supposedly looking up the vector is faster than a hashmap?

src/prelude.rs
yihozhang: Could you just keep the original `table_actions` but pass by reference here (instead of `self.table_actions.clone()`), and not have the indirections with `func_ids`, `table_actions_by_id`, and `table_actions_cache_by_id`? I'm not quite convinced that these new structs provide speedup

yihozhang (changes requested):

src/prelude.rs
```diff
@@ -373,20 +424,63 @@ impl RustRuleContext<'_, '_> {
     /// Insert a row into a table.
     /// For more information, see `egglog_bridge::TableAction::insert`.
     pub fn insert(&mut self, table: &str, row: impl Iterator<Item = Value>) {
-        self.get_table_action(table).insert(self.exec_state, row)
```
MilkBlock, line 376: This get_table_action call is expensive. Because it clones the TableAction every call. But this is redundant because though `insert` requires mutable borrow of TableAction but its so-called mutable borrow only change the capacity of scratch (because scratch will be cleared every call and other part will not change after first set).

So it should be cached to improve the performance.

```rust
pub struct TableAction {
    table: TableId,  // not change after first set
    table_math: SchemaMath,   // not change after first set
    default: Option<MergeVal>, // not change after first set
    timestamp: CounterId,  // not change after first set
    scratch: Vec<Value>,    // in fact scratch is a tmp variable will be `clear` in every TableOperation
}
```

There are only 4 kinds of TableOperations need TableAction : look_up, insert, delete, subsume. 

look_up & remove require & borrow
insert & delete require &mut borrow but it's fake(only touch capacity of scratch).

MilkBlock: @yihozhang  Basing on this observation, we can safely degrade the signature of `insert` and `subsume` from &mut TableAction to  & TableAction. So I can make this PR more readable.

PR #832, opened 2026-03-20
Primitive Semi-naive Matching Fix

### LINKED ISSUE #831: Rule Fails to Match With Containers and Functions
In https://github.com/egraphs-good/egglog-python/pull/397 I discovered a bug in Egglog around containers and higher order functions. Sometimes rules are not triggered when they should be. We can verify this by extracting a term, adding to a new e-graph, and then run the same rules and we will see we have new rewrites available.

This is what I am doing in the Python bindings to work around this for now, but it does make things slower.

Here's a small reproduction of the issue created by codex:

```
;; Thesis:
;; `unstable-vec-map` creates a vec element whose container child rewrites from
;; `(vec-of (w (k (b))))` to `(vec-of (k (b)))`, but saturation misses the new
;; match for `(l (vec-of (k x))) -> x` until the stuck term is re-added.

(sort E)
(sort VE (Vec E))
(sort VI (Vec i64))
(sort F (UnstableFn (i64) E))

(constructor b () E)
(constructor k (E) E)
(constructor w (E) E)
(constructor l (VE) E)
(constructor m (i64) E)

(rewrite (w x) x)
(rewrite (m i)
  (l (vec-of (w (k (b))))))
(rewrite (l (vec-of (k x)))
  x)

(push)
(let $stuck (unstable-vec-map (unstable-fn "m") (vec-of 0)))
(run-schedule (saturate (run)))
(check (= $stuck
  (vec-of (l (vec-of (k (b)))))))
(fail (check (= $stuck (vec-of (b)))))
(pop)

(let $resolved
  (vec-of (l (vec-of (k (b))))))
(run-schedule (saturate (run)))
(check (= $resolved (vec-of (b))))
```

saulshanabrook:
Fixes #831 by resolving how containers deal with updated contents and rebuildings.

Progress:

- [x] Narrow down bug to small reproduction to understand where it manifests (not actually related to containers, just primitives in general and semi-naive)
- [x] Try container specific fix. Too much code added and actualy core issue is not container specific (fast)
- [x] Try disabling semi naive for all primitives (too slow)
- [x] Try disabling semi naive only for primitives which are "stateful", i.e. their results can depend on state of union find and other containers (too slow)
- [x] Clean up initial work to just use that path for now
- [ ] Explore other tracking that is fast enough?
<details>
<summary>Implementation notes from LLM</summary>

## Background

The bug here is not that container primitives compute the wrong answer. The problem is that seminaive scheduling only revisits rule matches when it sees a fresh table delta, but container rebuild can change the semantics of an existing value *without* producing a fresh parent-row delta.

The important failing shape is:

- a row stores a container id
- the container’s contents rebuild from an unmatched shape to a matched shape
- the outer id stored in the parent row does not change
- a query-side matcher/primitive would now succeed
- seminaive never comes back to re-check that old row, so the match is missed

That is why the old repro behaved like:

- first egraph: saturates but gets stuck
- re-add the rebuilt term in a fresh context: now it matches

## Concrete example

A simplified version of the failure is:

```lisp
(rewrite (w x) x)
(rewrite (l (vec-of (k x))) x)

(let $stuck (l (vec-of (w (k (b))))))
(run-schedule (saturate (run)))
```

After rebuild, the inner container effectively becomes `vec-of(k(b))`, so the second rewrite should fire. But if the parent `l(...)` row still points at the same outer container id, seminaive sees no fresh row and never retries that query-side match.

## Why same-id rebuilds are the real problem

The subtle point is that not every container rebuild needs special handling.

If rebuild changes the outer container id, then ordinary table rebuild already rewrites parent rows to the new id, which creates a normal row delta. Seminaive sees that delta naturally.

The problematic case is narrower:

- the container’s *semantics* changed
- the parent row still stores the *same outer id*

That is the only time we need an explicit follow-up refresh.

## Chosen fix

The current fix is intentionally narrow and container-specific.

It does three things:

1. Container rebuild reports a summary of container ids whose semantics changed in a way that may require parent rows to be revisited.
2. Ordinary table rebuild still runs first.
3. After that, tables refresh rows whose rebuildable columns mention one of those changed container ids by re-inserting the same logical row with a fresh timestamp.

That refresh is only there to manufacture the missing delta so seminaive will revisit the newly enabled match.

In other words, the fix is:

- keep seminaive on
- rebuild containers
- do ordinary table rebuild
- if a container changed “in place” in a way that ordinary rebuild would not surface to seminaive, re-timestamp the affected parent rows

## What changed in the implementation

The implementation is small and follows the reasoning above:

- `ContainerRebuildSummary` records:
  - whether any container changed at all
  - which container ids may need parent-row refresh
- container rebuild now distinguishes:
  - changed-and-moved-to-a-different-id
  - changed-but-kept-the-same-id
- the refresh set only includes the second case
- `Table::refresh_rows_for_values` re-inserts matching logical rows with a fresh timestamp
- `EGraph::rebuild()` runs in this order:
  1. rebuild containers
  2. rebuild tables
  3. refresh parent rows for the same-id container changes

Two review-driven corrections were also important:

- the incremental container rebuild path now canonicalizes the outer id before reinsertion
- refresh ids are only recorded when the stored outer id stayed stable

Those details matter because refreshing by a post-rebuild id that already changed is both unnecessary and potentially over-broad.

## Why this approach was chosen

This approach was chosen because it preserves the performance characteristics of the original seminaive execution model and fixes the known container failure without broadening the scheduling policy for unrelated rules.

Compared with the alternatives we tried:

| Approach | Median on `python_array_optimize.egg` |
|---|---:|
| Original container-specific fix baseline | `926.1 ms` |
| Chosen dedicated refresh approach | `921.9 ms` |
| Persistent reverse index variant | `928.5 ms` |
| Disable seminaive for all query primitives | `1.721 s` |
| Stable/stateful primitive split | `1.800 s` |

These runs used the current dirty `tests/python_array_optimize.egg`, copied into detached worktrees for each revision so the workload stayed constant across comparisons.

So the chosen approach stays in the same performance band as the original fast container fix, while the broader primitive-level fallbacks were roughly 2x slower on this workload.

## Why not disable seminaive for query primitives?

That was the simplest soundness fix, and it does work, but it is too expensive in practice.

The broad primitive-level fallback makes many rules re-run much more often than necessary, even when the only real problem is a small number of container-backed matches becoming newly valid after rebuild.

For this bug, that was a large performance regression for a much broader policy change than we actually needed.

## Why not keep the stable/stateful primitive split?

That approach is more principled than “disable seminaive for everything,” but on this workload it did not recover the lost performance. It was still much slower than the container-specific fix, and it adds policy complexity around primitive classification that is not actually needed to solve the concrete failure here.

It also still does not provide a fully general invalidation story for all stateful query primitives.

## Why not use the persistent reverse index variant?

We also tried a deeper reverse-index version that maintained a dedicated per-table value-to-row index for refreshes.

It worked, but it added more code and did not produce a measurable win over the simpler dedicated-refresh version. Since the current refresh path is not hot enough for that extra maintenance cost to pay back, the simpler variant was preferred.

## Scope of the fix

This fix is intentionally narrow.

It fixes the container manifestation of the bug:

- a query-side match becomes newly true because container rebuild changed semantics without producing a fresh parent-row delta

It does **not** claim to solve the broader class of all stateful query primitives. In particular, a primitive whose answer changes because of evolving UF state, table state, or other runtime state would still need a more general invalidation strategy.

## Longer-term alternatives

The main longer-term directions are:

1. Keep this targeted container fix as-is.
2. Add coarser primitive metadata / domain invalidation for other stateful cases.
3. Build full dependency tracking for query primitives.

The third option is the most principled long-term design, but it is much more invasive because it requires:

- dependency reporting from query primitives
- tracking both successful and failed primitive evaluations
- reverse invalidation indexes
- scheduler support for waking matches even when no table row changed

For the current bug, that was too much machinery compared with the narrow container-specific repair.

## Why this is a reasonable compromise

This fix is somewhat specialized, but it is specialized in exactly the place where the missed-match behavior arises.

It keeps the existing seminaive execution model intact for normal rule evaluation, and only manufactures extra deltas when rebuild would otherwise hide a newly enabled match from the scheduler.

That tradeoff was preferable here because:

- it fixes the known bug
- it keeps the fast performance profile
- it is smaller and simpler than the reverse-index variant
- it avoids the large regression of disabling seminaive for query primitives more broadly

## Summary

The underlying issue is that seminaive only reacts to table deltas, while container rebuild can make an old query-side match newly true without creating one.

The chosen fix keeps seminaive fast and repairs exactly that missing delta by re-timestamping affected parent rows after container rebuild, but only for the same-id rebuild cases where ordinary table rebuild would not already have done the job.

</details>

### DIFF
--- a/core-relations/src/containers/mod.rs
+++ b/core-relations/src/containers/mod.rs
@@ -78,16 +78,19 @@ pub struct ContainerValues {
 /// `changed` means some container entry changed during rebuild, either because
 /// its contents changed or because its outer id canonicalized.
 ///
-/// `refresh_values` is narrower: it records container ids whose semantics
-/// changed while their stored outer id stayed stable. Ordinary table rebuild
-/// does not manufacture a fresh parent-row delta for that case, so parent rows
-/// mentioning those ids need a follow-up timestamp refresh.
+/// `dirty_ids` is narrower: it records container ids whose semantics changed
+/// while their stored outer id stayed stable. Ordinary table rebuild already
+/// handles changed-id cases; these ids need a follow-up parent-row refresh.
+///
+/// For example, `l(vec-of(w(k(b))))` can rebuild to `l(vec-of(k(b)))` without
+/// changing the `Vec` id. The row is now newly matchable, but seminaive will
+/// miss it unless the parent row is retimestamped.
 #[derive(Clone, Default)]
 pub struct ContainerRebuildSummary {
     changed: bool,
     // Container ids whose semantics changed in a way that may not produce a
     // fresh parent-row delta during ordinary table rebuild.
-    refresh_values: IndexSet<Value>,
+    dirty_ids: IndexSet<Value>,
 }
 
 impl ContainerRebuildSummary {
@@ -97,22 +100,22 @@ impl ContainerRebuildSummary {
     }
 
     /// Returns the container ids whose parent rows may need retimestamping.
-    pub fn refresh_values(&self) -> &IndexSet<Value> {
-        &self.refresh_values
+    pub fn dirty_ids(&self) -> &IndexSet<Value> {
+        &self.dirty_ids
     }
 
     fn note_change(&mut self) {
         self.changed = true;
     }
 
-    fn note_refresh_value(&mut self, value: Value) {
+    fn note_dirty_id(&mut self, value: Value) {
         self.changed = true;
-        self.refresh_values.insert(value);
+        self.dirty_ids.insert(value);
     }
 
     fn extend(&mut self, other: Self) {
         self.changed |= other.changed;
-        self.refresh_values.extend(other.refresh_values);
+        self.dirty_ids.extend(other.dirty_ids);
     }
 }
 
@@ -406,7 +409,7 @@ impl<C: ContainerValue> ContainerEnv<C> {
         }
         let actual = self.insert_owned(container, rebuilt_id, exec_state);
         if container_changed && rebuilt_id == old_id && actual == old_id {
-            summary.note_refresh_value(old_id);
+            summary.note_dirty_id(old_id);
         }
     }
 
@@ -513,7 +516,7 @@ impl<C: ContainerValue> ContainerEnv<C> {
             // If the outer id changed, ordinary table rebuild already creates a
             // fresh parent-row delta for seminaive to follow.
             if stable_id && actual == val {
-                summary.note_refresh_value(val);
+                summary.note_dirty_id(val);
             }
         }
         summary
@@ -576,7 +579,7 @@ impl<C: ContainerValue> ContainerEnv<C> {
             .max()
             .unwrap_or(false);
 
-        let refresh_values = SegQueue::new();
+        let dirty_ids = SegQueue::new();
         shards
             .iter_mut()
             .enumerate()
@@ -617,7 +620,7 @@ impl<C: ContainerValue> ContainerEnv<C> {
                             // As in the serial path, only same-id semantic
                             // changes need an explicit parent-row refresh.
                             if stable_id && result == val {
-                                refresh_values.push(val);
+                                dirty_ids.push(val);
                             }
                         }
                         Err(slot) => {
@@ -631,7 +634,7 @@ impl<C: ContainerValue> ContainerEnv<C> {
                                 shard.insert_in_slot(hc, slot, (container, SharedValue::new(val)));
                             }
                             if stable_id {
-                                refresh_values.push(val);
+                                dirty_ids.push(val);
                             }
                         }
                     }
@@ -641,8 +644,8 @@ impl<C: ContainerValue> ContainerEnv<C> {
         if changed {
             summary.note_change();
         }
-        while let Some(value) = refresh_values.pop() {
-            summary.note_refresh_value(value);
+        while let Some(value) = dirty_ids.pop() {
+            summary.note_dirty_id(value);
         }
         summary
     }
--- a/core-relations/src/free_join/mod.rs
+++ b/core-relations/src/free_join/mod.rs
@@ -371,17 +371,21 @@ impl Database {
     pub fn refresh_rows_for_values(
         &mut self,
         to_refresh: &[TableId],
-        values: &[Value],
+        dirty_ids: &[Value],
         next_ts: Value,
     ) -> bool {
-        if values.is_empty() {
+        if dirty_ids.is_empty() {
             return false;
         }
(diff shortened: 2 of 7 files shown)


core-relations/src/containers/mod.rs
```diff
@@ -73,6 +73,49 @@ pub struct ContainerValues {
...
+///
+/// `changed` means some container entry changed during rebuild, either because
+/// its contents changed or because its outer id canonicalized.
+///
+/// `refresh_values` is narrower: it records container ids whose semantics
+/// changed while their stored outer id stayed stable. Ordinary table rebuild
+/// does not manufacture a fresh parent-row delta for that case, so parent rows
+/// mentioning those ids need a follow-up timestamp refresh.
+#[derive(Clone, Default)]
+pub struct ContainerRebuildSummary {
```
ezrosent, line 89: I think it'll help to give a concrete scenario where this bug can happen

core-relations/src/containers/mod.rs
```diff
@@ -73,6 +73,49 @@ pub struct ContainerValues {
...
+    changed: bool,
+    // Container ids whose semantics changed in a way that may not produce a
+    // fresh parent-row delta during ordinary table rebuild.
+    refresh_values: IndexSet<Value>,
```
ezrosent: I think we can call this "dirty_id" or similar? I don't think refresh_vals is super descriptive.

core-relations/src/free_join/mod.rs
```diff
@@ -356,40 +356,68 @@ impl Database {
...
+                &func.table,
+                next_ts,
+                &mut ExecutionState::new(*view, Default::default()),
+            )
+        });
+        self.tables.insert(func_id, func);
+        self.merge_all()
+    }
+
+    pub fn refresh_rows_for_values(
```
ezrosent, line 371: Better documentation for what's happening here, with pointers back to the container code

core-relations/src/table/mod.rs
```diff
@@ -573,7 +577,7 @@ impl SortedWritesTable {
                         let (actual_shard, hc) = hash_code(shard_data, to_remove, self.n_keys);
                         assert_eq!(actual_shard, shard_id);
                         if let Ok(entry) = shard.find_entry(hc, |entry| {
-                            entry.hashcode == (hc as _)
+                            entry.hashcode == (hc as HashCode)
```
ezrosent: why is this change needed?

saulshanabrook: Removed change

PR #834, opened 2026-03-26
Add f64 exp, log, and sqrt primitives

saulshanabrook:
- add `exp`, `log`, and `sqrt` as f64 primitives
- expose them through the existing f64 sort implementation
- add egg tests for the new operations, including direct failure cases for invalid domain inputs


saulshanabrook:
This has already been merged in from another branch

PR #835, opened 2026-03-29
Move the thread pool field to core-relations

yihozhang:
* Made the index-building thread pool part of the e-graph 
* Addressed some review comments

One relatively invasive change is that now when we build an index or a SortedWritesTable we also need to pass in the thread pool for index building, which used to be a global variable.

### DIFF
--- a/core-relations/benches/writes.rs
+++ b/core-relations/benches/writes.rs
@@ -1,5 +1,7 @@
+use std::sync::Arc;
+
 use divan::{Bencher, counter::ItemsCount};
-use egglog_core_relations::{Database, SortedWritesTable, Table, Value};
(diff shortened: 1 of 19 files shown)


yihozhang:
I can confirm a significant slowdown from this PR

```
➜  egglog git:(oflatt-refactor-parallel-core-relations) ✗ time target/release/egglog tests/web-demo/typeinfer.egg
target/release/egglog tests/web-demo/typeinfer.egg  0.28s user 0.08s system 119% cpu 0.310 total
➜  egglog git:(oflatt-refactor-parallel-core-relations) ✗ time ~/egglog-main/target/release/egglog tests/web-demo/typeinfer.egg
~/egglog-main/target/release/egglog tests/web-demo/typeinfer.egg  0.11s user 0.01s system 99% cpu 0.120 total
```

PR #838, opened 2026-04-04
[codex] perf(rust_rule): share table actions across matches

MilkBlock:
This follow-up isolates a simpler version of the Rust-rule table action optimization on top of `main`.

Rust rules currently clone the full `table_actions` map on every match because `RustRuleContext` owns that map, and `TableAction::insert` / `TableAction::subsume` require `&mut self`. In practice, that mutable state was only the reusable `scratch` buffer inside `TableAction`; the semantic table metadata (`table`, `table_math`, `default`, `timestamp`) is immutable after construction.

This change removes `scratch` from `TableAction` and replaces its uses with per-call temporary vectors in `insert` and `subsume`. That lets both methods take `&self` instead of `&mut self`. With that API change, `RustRuleRhs` can hold a shared `Arc<HashMap<String, TableAction>>` and `RustRuleContext` can borrow it rather than cloning the entire map for every match.

Effectively, this makes the optimization much smaller and more direct:
- `TableAction` becomes immutable metadata plus behavior
- `RustRuleContext` borrows shared table actions instead of copying them
- the old per-clone scratch-buffer reset logic disappears

I also cleaned up now-unneeded `mut` bindings at the few call sites that were only mutable because `insert` previously required `&mut self`.

Validation performed locally:
- `cargo test -p egglog-bridge`
- `cargo test -p egglog --lib`
- `cargo bench --bench rust_api_benchmarking --no-run`

### DIFF
--- a/egglog-bridge/src/lib.rs
+++ b/egglog-bridge/src/lib.rs
@@ -1442,25 +1442,12 @@ impl ResolvedMergeFn {
 /// This is an intern-able struct that holds all the data needed
 /// to do table operations with an [`ExecutionState`], assuming
 /// that the [`FunctionId`] for the table is known ahead of time.
-#[derive(Debug, PartialEq, Eq, Hash)]
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
 pub struct TableAction {
     table: TableId,
     table_math: SchemaMath,
     default: Option<MergeVal>,
     timestamp: CounterId,
-    scratch: Vec<Value>,
-}
-
-impl Clone for TableAction {
-    fn clone(&self) -> Self {
-        Self {
-            table: self.table,
-            table_math: self.table_math,
-            default: self.default,
-            timestamp: self.timestamp,
-            scratch: Vec::new(),
-        }
-    }
 }
 
 impl TableAction {
@@ -1483,7 +1470,6 @@ impl TableAction {
                 DefaultVal::Const(val) => Some(MergeVal::Constant(*val)),
             },
             timestamp: egraph.timestamp_counter,
-            scratch: Vec::new(),
         }
     }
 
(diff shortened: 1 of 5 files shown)


MilkBlock (commented):
Please review this instead @yihozhang

egglog-bridge/src/lib.rs
```diff
@@ -1525,20 +1511,19 @@ impl TableAction {
     }
 
     /// Insert a row into this table.
-    pub fn insert(&mut self, state: &mut ExecutionState, row: impl Iterator<Item = Value>) {
+    pub fn insert(&self, state: &mut ExecutionState, row: impl Iterator<Item = Value>) {
         let ts = Value::from_usize(state.read_counter(self.timestamp));
-        self.scratch.clear();
-        self.scratch.extend(row);
+        let mut scratch = row.collect::<Vec<_>>();
```
```suggestion
        let mut scratch = row.collect::<SmallVec<[_; 8]>>();
```

egglog-bridge/src/lib.rs
```diff
@@ -1547,25 +1532,22 @@ impl TableAction {
     }
 
     /// Subsume a row in this table.
-    pub fn subsume(&mut self, state: &mut ExecutionState, key: impl Iterator<Item = Value>) {
+    pub fn subsume(&self, state: &mut ExecutionState, key: impl Iterator<Item = Value>) {
         let ts = Value::from_usize(state.read_counter(self.timestamp));
-        self.scratch.clear();
-        self.scratch.extend(key);
+        let mut scratch = key.collect::<Vec<_>>();
```
```suggestion
        let mut scratch = key.collect::<SmallVec<[_; 8]>>();
```

src/prelude.rs
```diff
@@ -404,7 +419,7 @@ struct RustRuleRhs<F: Fn(&mut RustRuleContext, &[Value]) -> Option<()>> {
     name: String,
     inputs: Vec<ArcSort>,
     union_action: egglog_bridge::UnionAction,
-    table_actions: HashMap<String, egglog_bridge::TableAction>,
+    table_actions: std::sync::Arc<HashMap<String, egglog_bridge::TableAction>>,
```
yihozhang: I think table_actions does not need to be wrapped in an `Arc`.

yihozhang (changes requested):
Thank you! Sorry I didn't get your previous PR but this PR makes sense to me.

src/prelude.rs
MilkBlock: Thank you for your advice!

yihozhang (approved):
Looks good! Thank you

PR #840, opened 2026-04-09
Index optimized for small data

yihozhang:
4/5 of the indices built for some large benchmarks only have one element. It is expensive to build a ColumnIndex for them.

### DIFF
--- a/core-relations/src/free_join/execute.rs
+++ b/core-relations/src/free_join/execute.rs
@@ -46,7 +46,7 @@ use super::{
 
 const SMALL_RESIDUAL: usize = 8;
 
-struct SparseIndex {
+struct SparseColumnIndex {
     n_keys: usize,
     n_subsets: usize,
     keys: [Value; SMALL_RESIDUAL],
@@ -54,7 +54,7 @@ struct SparseIndex {
     subset_ids: [RowId; SMALL_RESIDUAL],
 }
 
-impl SparseIndex {
+impl SparseColumnIndex {
     fn keys(&self) -> &[Value] {
         &self.keys[..self.n_keys]
     }
@@ -105,7 +105,7 @@ impl SparseIndex {
             subset_ids[i] = row_id;
         }
 
-        SparseIndex {
+        SparseColumnIndex {
             n_keys,
             n_subsets,
             keys,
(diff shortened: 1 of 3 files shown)


yihozhang:
pinging @ezrosent ! I'd like this PR merged to see how other performance optimizations interact with this one.

core-relations/src/free_join/execute.rs
```diff
@@ -55,6 +155,7 @@ enum DynamicIndex {
     },
     Dynamic(TupleIndex),
     DynamicColumn(Arc<ColumnIndex>),
+    Sparse(SparseIndex),
```
ezrosent: nit: Should we keep the Tuple vs. Column naming here? (i.e. name it SparseColumnIndex).

core-relations/src/row_buffer/mod.rs
```diff
@@ -17,15 +17,102 @@ use crate::{
 #[cfg(test)]
 mod tests;
 
+/// A trait for types that can store a vector of `Value`s.
+///
+/// Can either be backed by a pool or SmallVec.
+pub trait ValueVec {
+    fn as_values(&self) -> &[Cell<Value>];
```
ezrosent: I don't think we need Cell<Value>

core-relations/src/row_buffer/mod.rs
```diff
@@ -17,15 +17,102 @@ use crate::{
...
+        let v: &mut Vec<Cell<Value>> = self;
+        v.extend(values.iter().copied().map(Cell::new));
+    }
+
+    fn refresh(&mut self) {
+        Pooled::refresh(self);
+    }
+}
+
+pub(crate) type SmallValueVec = SmallVec<[Cell<Value>; 64]>;
```
ezrosent, line 67: will we ever have more than 64 values in this vector? If not, can we use tinyvec? https://docs.rs/tinyvec/latest/tinyvec/

ezrosent (approved):

core-relations/src/row_buffer/mod.rs
yihozhang: Thanks for the pointer to tinyvec! I think this PR's usage will never have more than 64 values, but in general it could, because the caller of `scan_project` may call with an arbitrarily big subset

PR #844, opened 2026-04-10
Reuse sub-tries during query execution

yihozhang:
The existing implementation of TrieNode only caches the column index, which is a mapping from key to a list of Row Ids. However, subtries are still created every time a key of the TrieNode is probed/iterated. So the existing implementation can repeatedly create TrieNodes which is very wasteful. This PR tries to also store subtries as part of the TrieNode.

### DIFF
--- a/core-relations/src/free_join/execute.rs
+++ b/core-relations/src/free_join/execute.rs
@@ -941,7 +941,7 @@ impl<'a> JoinState<'a> {
                             updates.push_binding(*var, val[0]);
 
                             let node = if x.size() <= 16 {
-                                let sub = refine_subset(x.to_owned(&ps.get_pool()), &[], &table);
+                                let sub = refine_subset(x.to_owned(&ps.get_pool()), &a.cs, &table);
                                 Arc::new(TrieNode::new(sub))
                             } else {
                                 prober
--- a/core-relations/src/free_join/execute.rs
+++ b/core-relations/src/free_join/execute.rs
@@ -3,11 +3,11 @@
 use std::{
     cmp, iter, mem,
     ops::Range,
-    sync::{Arc, Mutex, OnceLock, atomic::AtomicUsize},
+    sync::{Arc, OnceLock, atomic::AtomicUsize},
 };
 
 use crate::{
-    common::{HashMap, ShardData, ShardId},
+    common::HashMap,
     free_join::plan::{JoinStages, MatId, MatScanMode, MatSpec},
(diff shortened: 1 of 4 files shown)


core-relations/src/free_join/execute.rs
```diff
@@ -457,18 +499,21 @@ impl Database {
 
                 let search_and_apply_timer = Instant::now();
                 'eval: {
+                    for JoinHeader { atom, subset, .. } in plan.header() {
+                        if subset.is_empty() {
+                            break 'eval;
+                        }
+                        let mut cur = Arc::try_unwrap(binding_info.unwrap_val(*atom)).unwrap();
```
ezrosent, line 508: Explain why this won't crash (i.e. the Arc always has refcount==1).

core-relations/src/free_join/execute.rs
```diff
@@ -600,7 +631,14 @@ struct JoinState<'a> {
     exec_state: ExecutionState<'a>,
 }
 
+// TODO: use SmallVec might be better
```
ezrosent: clean up these comments and document how ColumnIndexes is used?

core-relations/src/free_join/execute.rs
```diff
@@ -600,7 +631,14 @@ struct JoinState<'a> {
...
 type ColumnIndexes = IdVec<ColumnId, OnceLock<Arc<ColumnIndex>>>;
+// pub(crate) type ChildrenMaps =
+//     IdVec<ColumnId, OnceLock<Arc<HashMap<Value, OnceLock<Arc<TrieNode>>>>>>;
+
+// pub(crate) type ChildrenMaps = IdVec<ColumnId, OnceLock<Arc<DashMap<Value, Arc<TrieNode>>>>>;
+// pub(crate) type ChildrenMaps = IdVec<ColumnId, Arc<DashMap<Value, Arc<TrieNode>>>>;
+pub(crate) type ChildrenMaps = IdVec<ColumnId, ReadOptimizedLock<HashMap<Value, Arc<TrieNode>>>>;
```
ezrosent, line 641: Should this be a OnceLock instead? I'd expect we could initialize this hashmap once only..

ezrosent (approved):
Minor comments.

core-relations/src/free_join/execute.rs
yihozhang: The idea is that we won't necessarily build a subtrie for every value in the subset, but only the ones probed. With OnceLock we would need to initialize the HashMap with all the values I think?

PR #846, opened 2026-04-12
[codex] fold math microbenchmark into rust_api_benchmarking

MilkBlock:
## Summary

This change folds the Rust-rule math microbenchmark into the existing `rust_api_benchmarking` bench suite so it runs alongside the other Rust API hot-path benchmarks instead of living in a separate bench target.

It also cleans up the shared benchmark support by making `run_math_microbenchmark_iters` honor its `iters` parameter and by failing fast when Rust-rule registration itself fails during setup.

## Root cause

The math microbenchmark had been added as a standalone `[[bench]]`, which meant it was disconnected from the upstream Rust API benchmarking entrypoint that now carries the table-action hot-path cases. That made it harder to compare and maintain these Rust API microbenchmarks together.

The support helper also hard-coded eight iterations regardless of the caller input, which made the API misleading and hid issues in tests that expected to control iteration count.

## Fix

The math support module is now imported directly by `benches/rust_api_benchmarking.rs`, and a new `rust_rule_math_microbenchmark` divan bench is registered there.

A small test file exercises the setup path and one benchmark iteration, including the case where the global Rayon pool has already been initialized.

## Validation

- `cargo test --test rust_api_math_microbenchmark -- --nocapture`
- `cargo bench --bench rust_api_benchmarking -- --list`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,6 +62,7 @@ jobs:
           - rust_rule_fib
           - rust_rule_match_overhead
           - rust_rule_match_with_serialize
+          - rust_rule_math_microbenchmark
           - rust_rule_insert_loop
           - rust_rule_tableaction_hot_path
     steps:
--- a/benches/math_microbenchmark_support.rs
+++ b/benches/math_microbenchmark_support.rs
@@ -1,0 +1,572 @@
+use egglog::prelude::*;
+use egglog::sort::S;
+use egglog::{ArcSort, Value};
+use std::sync::Once;
+
+static CONFIGURE_RAYON: Once = Once::new();
+
+pub struct MathMicroBenchInput {
+    pub egraph: egglog::EGraph,
+    pub ruleset: String,
+}
+
+fn configure_rayon_once() {
+    CONFIGURE_RAYON.call_once(|| {
+        let _ = rayon::ThreadPoolBuilder::new()
+            .num_threads(1)
+            .build_global();
+    });
+}
+
+fn math_vars(math: &ArcSort, names: &[&'static str]) -> Vec<(&'static str, ArcSort)> {
+    names.iter().map(|name| (*name, math.clone())).collect()
+}
+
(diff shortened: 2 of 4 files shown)


saulshanabrook:
It doesn't look like this benchmark is run in CI currently?

MilkBlock:
I've checked the num_tuples generated by .egg and rust api.
```
egg time: 16.07482225s
[egg] total num_tuples = 1047896

rust time: 17.833966666s
[rust] total num_tuples = 1047896
```

MilkBlock:
```
Run CodSpeedHQ/action@v4
Run RUNNER_VERSION=""
Run actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7
Cache hit for: codspeed-instruments-simulation-Linux-X64-6.17.0-1010-azure-4.13.0
Received 29076559 of 29076559 (100.0%), 44.7 MBs/sec
Cache Size: ~28 MB (29076559 B)
/usr/bin/tar -xf /home/runner/work/_temp/f23dc62d-fd50-468f-80d9-f787e82c75e3/cache.tzst -P -C /home/runner/work/egglog/egglog --use-compress-program unzstd
Cache restored successfully
Cache restored from key: codspeed-instruments-simulation-Linux-X64-6.17.0-1010-azure-4.13.0
Run actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd
Run RUNNER_VERSION="4.13.0"
Error: Installation of CodSpeed CLI with version 4.13.0 failed.
Reason: curl: (28) Failed to connect to codspeed.io port 443 after 135756 ms: Couldn't connect to server
Error: Process completed with exit code 1.
```
I don't quite understand what happens here. It seems like Codspeed connection is unstable now? 
I will have another try next day.

yihozhang:
I kicked a re-run and it works now. Looks good. Could you add the assertion that num_tuples = 1047896?

MilkBlock:
> I kicked a re-run and it works now. Looks good. Could you add the assertion that num_tuples = 1047896?

added

PR #847, opened 2026-04-14
Better subset representation with BitSet

yihozhang:


yihozhang:
I am closing this for now - although a smarter representation can make set intersection much faster, it is currently not a bottleneck.

> - **Set intersection is not a bottleneck.** Sorted-set intersection
>   (`Subset::intersect`, all variant pairs, plus `intersect_with_dense_ref`)
>   is ≤ ~2.3% of runtime everywhere, and most of that is the *result copy*,
>   not the intersection math. The classic bitvector win (word-parallel AND)
>   has nothing to bite on.
> - **Subset traffic is bimodal.** By *op count*, subsets are tiny: 60–85% of
>   probed/refined subsets are singletons, 95%+ are ≤ 16 rows. By *element
>   volume*, traffic is dominated by mid-size, half-dense subsets: 50–95% of
>   all elements copied/iterated/retained live in subsets of size 17–4K (some
>   to 64K) with density > 1/8, usually > 1/2.
> - **What actually costs money** is (a) *materializing* borrowed subsets into
>   owned vectors (`to_owned`, `Dense∩Sparse` subslice copies): ~5% on
>   whisper-class workloads; (b) *staleness refinement* on millions of tiny
>   subsets (paged_llama: ~13–15%, pure per-op overhead, representation can't
>   help); and (c) *per-trie-node sorted-index construction* (radix sort of
>   (value,row) pairs): 20–27% on whisper/qwen3_moe — representation-independent.
> - **Verdict:** a bitvector variant is justified only as a *materialization
>   format* for the mid-size half-dense class, where it cuts copy bytes 4–32×
>   and makes `retain`/repeated intersection in-place. Expected end-to-end win
>   ~2–4% on whisper-class benchmarks, ~0–1% elsewhere. The same sites can win
>   more (all of the ~5%) from O(1) *sharing* (Arc-backed views into
>   index-owned arrays), which composes with — and partly obsoletes — the
>   bitvector. Recommendation: do sharing first, add the `Bits` variant behind
>   the same materialization funnel if the remaining copy volume still shows up.

PR #848, opened 2026-04-14
Rebuild only when union find changes

oflatt:


saulshanabrook:
The included benchmark from luminal improves that script by 17% with this change. It also runs a bunch of rules, so seems significant that just improving let binding performance by referring rebuilding has such a big impact!

oflatt:
I don't think that screen shot above reflects the actual speedup- if you look at the diff it goes from 1.6 seconds to 0.3

That was for a different proof benchmark!

saulshanabrook:
Thanks! Yeah you are right, I was looking at overall speedup for *all* examples, not just that one.

saulshanabrook:
@coderabbitai full review

ezrosent (approved):
Nice improvement! I believe this is safe. If it ends up having an issue I think we should be able to tighten up some APIs and keep this optimizations

yihozhang:
Nice speedup!

PR #850, opened 2026-04-18
Proof optimization- consolidate proof and view/UF tables

oflatt:
This PR makes proofs twice as fast by consolidating tables, also resulting in a negative diff in the core proof instrumentation


src/proofs/proof_encoding.rs
```diff
@@ -446,11 +447,12 @@ impl<'a> ProofInstrumentor<'a> {
...
         // Propagate cost, unextractable, hidden, and internal_let flags from the original function
         let mut term_flags = String::new();
-        let mut view_flags = String::new();
         if let Some(cost) = fdecl.cost {
             term_flags.push_str(&format!(" :cost {cost}"));
-            view_flags.push_str(&format!(" :cost {cost}"));
         }
+        // View is always a function (returning Proof or Unit), with :merge old
+        let proof_type = self.proof_type_str().to_string();
+        let mut view_flags = String::new();
```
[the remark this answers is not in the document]
oflatt: ah this is a real problem

egglog-bridge/src/rule.rs
```diff
@@ -784,10 +794,20 @@ impl Query {
         let mut constraints: Vec<(core_relations::AtomId, Constraint)> =
             Vec::with_capacity(self.atoms.len());
         'outer: for focus_atom in 0..self.atoms.len() {
+            if self.skip_seminaive.contains(&focus_atom) {
+                continue;
+            }
             for (i, (_, _, schema_info)) in self.atoms.iter().enumerate() {
```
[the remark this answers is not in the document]
oflatt: hmm this shouldn't be here anymore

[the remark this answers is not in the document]
oflatt: we should rip out the changes in this file

src/proofs/snapshots/egglog__proofs__proof_tests__tests__doc_example_add_function1.snap
```diff
@@ -9,35 +9,34 @@ expression: snapshot
...
-       (__UF_Math b c))
+       (set (__UF_Math b c) ()))
         :ruleset __single_parent :name "singleparent__uf_update")
 (rule ((__UF_Math a b))
       ((set (__UF_Mathf a) b))
         :ruleset __uf_function_index :name "__uf_function_index1")
 (sort __view)
 (constructor Add (i64 i64) Math :unextractable :internal-hidden)
-(constructor __AddView (i64 i64 Math) __view :term-constructor Add)
+(function __AddView (i64 i64 Math) Unit :merge old :term-constructor Add)
```
saulshanabrook, line 32: Just curious, why do we need merge old here? Shouldn't any two units be the same value so the default merge function of assert equality should be fine?

oflatt: The function doesn't do much in term mode, in proof mode, the `Unit` is a `Proof` and we want to store only one `Proof`

src/proofs/proof_encoding.rs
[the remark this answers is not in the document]
oflatt: fixed

egglog-bridge/src/rule.rs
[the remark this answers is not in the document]
oflatt: fixed

src/sort/pair.rs
```diff
@@ -0,0 +1,149 @@
+use super::*;
```
saulshanabrook, line 1: This looks good! I have an in progress pair.rs file as well, looks about the same. I called them `left` and `right` but I think yours are better names and are closer to other languages like lean/Haskell. I added `unstable-pair-match`, `unstable-pair-map-left`, and `unstable-pair-map-right`, so a subsequent PR from me might add those if it ends up being useful.

src/ast/mod.rs
```diff
@@ -976,19 +980,27 @@ where
...
+                    write!(f, " :unextractable")?;
+                }
                 if *hidden {
                     write!(f, " :internal-hidden")?;
                 }
                 if *let_binding {
                     write!(f, " :internal-let")?;
                 }
+                if let Some(tc) = term_constructor {
+                    write!(f, " :term-constructor {tc}")?;
```
saulshanabrook: Sorry this is probably an unrelated change, but should we make this `internal`? Or is it meant to be used outside of proof mode?

src/sort/pair.rs
oflatt: okay thanks!

src/ast/mod.rs
oflatt: Yeah let's make it internal

yihozhang (approved):
Took a look at the non-proof part of the PR and looks good to me! I like the idea of `extraction-cost.`

PR #852, opened 2026-04-25
Claude experiment

yihozhang:
I have categorized the optimizations Claude did into the following categories.

- Low-level optimizations
    - fast paths
        - shard computation when shard_count == 0
        - don’t access thread-local pool if the subset is empty
        - skip staleness check if we know the table does not have stale rows (good for datalog only workload)
    - RefineAtomDense
        - avoids Arc<TrieNode> creation if we later find a candidate to join does not satisfy the given constraints
    - Replace get_row with unsafe versions get_row_unchecked
    - Replace debug_assert! with assert!
    - Inline the last level of GJ recursion
    - avoid drain_updates_parallel is we know we are in single thread
        - parallel draining may copy some intermediate state even if we are not creating new threads
- Good improvements
    - Avoid thread-local storage access, store the thread-local pointer in a struct
- Great improvements
    - Fixed a performance bug when intersecting two subsets
        - Old algorithm can take O(n log m) time in worse case when the ranges of the two subsets do not overlap
    - When interesting a slice of row ids with a range [l, r], the result does not need to be instantiated/allocated: the result is always a subslice of the original slice.

### DIFF
--- a/core-relations/src/free_join/execute.rs
+++ b/core-relations/src/free_join/execute.rs
@@ -269,8 +269,7 @@ impl Prober {
             }
             DynamicIndex::SparseColumn(tab) => {
                 debug_assert_eq!(key.len(), 1);
-                tab.get_subset(key[0])
-                    .map(PotentiallyStale::not_stale)
+                tab.get_subset(key[0]).map(PotentiallyStale::not_stale)
             }
         }
     }
--- a/core-relations/src/free_join/execute.rs
+++ b/core-relations/src/free_join/execute.rs
@@ -3,7 +3,7 @@
 use std::{
     cmp, iter, mem,
     ops::Range,
-    sync::{Arc, OnceLock, atomic::AtomicUsize},
+    sync::{Arc, OnceLock, RwLock, atomic::AtomicUsize},
 };
 
 use crate::{
@@ -16,7 +16,6 @@ use crate::{
 use crossbeam::utils::CachePadded;
 use dashmap::mapref::entry::Entry;
 use dashmap::mapref::one::RefMut;
-use egglog_concurrency::ReadOptimizedLock;
 use egglog_reports::{ReportLevel, RuleReport, RuleSetReport};
 use smallvec::SmallVec;
 use web_time::Instant;
@@ -147,11 +146,15 @@ impl SparseColumnIndex {
 
 enum DynamicIndex {
     Cached {
-        intersect_outer: bool,
+        /// When Some(range), intersect each subset from the index with this dense range.
+        /// The range is the Dense outer subset known at Prober construction time.
+        intersect_outer: Option<OffsetRange>,
         table: HashIndex,
     },
     CachedColumn {
-        intersect_outer: bool,
+        /// When Some(range), intersect each subset from the index with this dense range.
+        /// The range is the Dense outer subset known at Prober construction time.
+        intersect_outer: Option<OffsetRange>,
         table: HashColumnIndex,
     },
     Dynamic(TupleIndex),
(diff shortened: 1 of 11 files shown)


core-relations/src/free_join/execute.rs
```diff
@@ -519,16 +537,17 @@ impl Default for ActionState {
 struct JoinState<'a> {
     db: &'a Database,
     exec_state: ExecutionState<'a>,
+    /// Cached thread-local pool for SortedOffsetVector allocations.
+    /// Stored here to avoid a per-call `with_pool_set` TLS access in `get_index`.
+    pool: Pool<SortedOffsetVector>,
```
yihozhang, line 651: JoinState is always used within a state, so we can cache the thread-local pool here. I can confirm this improves the performance.

core-relations/src/offsets/mod.rs
```diff
@@ -43,7 +43,7 @@ impl Offsets for OffsetRange {
 
 impl OffsetRange {
     pub fn new(start: RowId, end: RowId) -> OffsetRange {
-        assert!(
+        debug_assert!(
```
yihozhang, line 46: these `assert`s seem to have a non-trivial impact on performance

core-relations/src/offsets/mod.rs
```diff
@@ -372,34 +373,62 @@ impl Subset {
...
             }
             (Subset::Sparse(cur), SubsetRef::Sparse(other)) => {
+                // Hybrid intersect: O(1) fast paths for dense matches and already-past
+                // cases, falling back to binary search when other needs to advance.
                 let mut other_off = 0;
-                cur.retain(|rowid| match other.scan_for_offset(other_off, rowid) {
-                    Ok(found) => {
-                        other_off = found + 1;
-                        true
+                cur.retain(|rowid| {
```
yihozhang: Existing implementation is m log n if the two subsets' bounds don't overlap, this fixes it.

core-relations/src/free_join/frame_update.rs
```diff
@@ -58,6 +61,12 @@ impl FrameUpdates {
         self.updates.push(UpdateInstr::RefineAtom(atom, node));
     }
 
+    /// Refine `atom` to consider only the given dense offset range, without
+    /// allocating an Arc<TrieNode> eagerly.
+    pub(super) fn refine_atom_dense(&mut self, atom: AtomId, range: OffsetRange) {
```
yihozhang, line 66: I can confirm using refine_atom_dense indeed improves performance (small but consistent). This is somewhat surprising.

oflatt:
Woah, some crazy optimizations
Probably worth the complexity? Hopefully we won't break these in the future

core-relations/src/table/mod.rs
```diff
@@ -970,7 +991,13 @@ impl SortedWritesTable {
     }
 
     fn get_if(&self, cs: &[Constraint], row: RowId) -> Option<&[Value]> {
-        let row = self.data.get_row(row)?;
+        // Fast path: when no stale rows, skip the stale check.
+        let row = if self.data.stale_rows == 0 {
+            // SAFETY: callers guarantee row is within bounds via subset construction.
+            unsafe { self.data.data.get_row_unchecked(row) }
```
oflatt: oof, idk about this one
maybe we mark this as inline and then rust may be able to optimize it away?

core-relations/src/table/mod.rs
```diff
@@ -357,19 +361,25 @@ impl Table for SortedWritesTable {
...
-        subset.offsets(|row| unsafe {
-            if let Some(vals) = self.data.get_row_unchecked(row) {
-                f(row, vals)
-            }
-        })
+        if self.data.stale_rows == 0 {
+            // Fast path: no stale rows, skip is_stale check per row.
+            // SAFETY: subsets are sorted, low must be at most hi, and hi is less
+            // than the length of the table.
+            subset.offsets(|row| unsafe { f(row, self.data.data.get_row_unchecked(row)) })
```
oflatt, line 374: more unsafe- worth it? I guess here I'm more okay with it

oflatt:
Just added a couple random comments, didn't do a full review

saulshanabrook:
@coderabbitai full review

yihozhang:
I have addressed the concerns with unsafe code and correctly implemented galloping. This PR is ready for another look!

core-relations/src/table/mod.rs
ezrosent: Looks like there were unsafe lines before as well, 

but we should actually be able to provide a safe API here:

`subset` is a sorted range of offsets, so we should be able to do a single bounds-check for the entire RowBuffer and package that up into a helper, I think.

core-relations/src/free_join/execute.rs
```diff
@@ -917,7 +957,7 @@ impl<'a> JoinState<'a> {
                 if self.exec_state.should_stop() {
                     return;
                 }
-                if cur == 0 || cur == 1 {
+                if (cur == 0 || cur == 1) && action_buf.supports_parallel_drain() {
                     drain_updates_parallel!($updates)
```
ezrosent, line 961: So... the intent for the action buffer trait is that the calls to `drain_updates_parallel!` wouldn't have any overhead for the in-place action buffer. Did you investigate what the actual overhead was?

yihozhang: drain_updates_parallel! would clone ExecutionState twice before doing any real work. This is probably where the slowdown is from.

```
let db = self.db;
let exec_state_for_factory = self.exec_state.clone();
let exec_state_for_work = self.exec_state.clone();
action_buf.recur( ... )
```

yihozhang: Getting rid of this yields
```
  hardboiled_conv1d_32.egg                        0.179       0.188  +  0.009  +   5.0%  ▲ slower
  hardboiled_conv1d_128.egg                       0.586       0.610  +  0.024  +   4.1%  ▲ slower
  luminal-llama.egg                               0.091       0.092  +  0.001  +   1.1%  ▲ slower
  python_array_optimize.egg                       0.402       0.413  +  0.011  +   2.7%  ▲ slower
  cykjson.egg                                     0.065       0.064   -0.001    -1.5%  ▼ faster
  eggcc-extraction.egg                            0.240       0.243  +  0.003  +   1.3%  ▲ slower

  Summary: 1 faster  ·  5 slower  ·  0 unchanged  ·  0 missing
```
so it in particular has an impact on hardboiled benchmarks.

core-relations/src/table/mod.rs
yihozhang: Aren't we already doing this for `scan_generic`, with the `assert` immediately before this? This isn't true for `scan_generic_bounded`, which does the bounds-check per invocation (instead of per RowBuffer)

ezrosent: Yeah we are already doing this technique elsewhere. I'm suggesting we add a method on RowBuffer that takes a subset and an `impl FnMut(RowId, &[Value])` (or similar).

core-relations/src/free_join/execute.rs
ezrosent: Hmm... I wonder if we can refactor action_buf to move these clones under the abstraction. Maybe add a TODO around `supports_parallel_drain` to look into this?

ezrosent (approved):

yihozhang:
Oops I forgot to add the TODO about supports_parallel_drain before clicking merge. Will do this in my other PRs

PR #854, opened 2026-04-26
Avoiding create semi-naive rule instantiations when unnecessary.

yihozhang:
`python_array_optimize.egg` has about 500 rules and 1500 iterations. During egglog execution, `add_rule_from_cached_plan` is called 721260 times, while the number of plans actually run is 34661, and the final database only has ~3500 tuples.

This PR attempts to refuse to create a rule when it is known to produce an empty result.


egglog-bridge/src/rule.rs
```diff
@@ -813,9 +811,18 @@ impl Query {
                     )),
                     Ordering::Greater => {}
                 };
+            };
+            // start with the focus atom since `add_rule_from_cached_plan` will apply the
+            // constraints in order, and the focus atom may have an empty delta, which
+            // will let it bail early.
+            work(focus_atom, &self.atoms[focus_atom].2);
```
oflatt: what is "work"? need better name?

egglog-bridge/src/rule.rs
```diff
@@ -766,14 +766,14 @@ impl Query {
...
-            rsb.add_rule_from_cached_plan(&cached_plan.plan, &[]);
+            let _ = rsb.add_rule_from_cached_plan(&cached_plan.plan, &[]);
             return Ok(());
         }
         if let Some(focus_atom) = self.sole_focus {
             // There is a single "focus" atom that we will constrain to look at new values.
             let (_, _, schema_info) = &self.atoms[focus_atom];
             let ts_col = ColumnId::from_usize(schema_info.ts_col());
-            rsb.add_rule_from_cached_plan(
+            let _ = rsb.add_rule_from_cached_plan(
```
oflatt, line 776: why _ here? Don't need the option results?

oflatt: not sure I see the use of the option here yet

core-relations/src/query.rs
```diff
@@ -134,128 +134,142 @@ impl<'outer> RuleSetBuilder<'outer> {
...
-        let plan = self.get_rule_with_extra_constraints(cached, action_id, extra_constraints);
-        self.rule_set
-            .plans
-            .push((plan, cached.desc.clone(), cached.symbol_map.clone()))
+    ) -> Option<RuleId> {
+        // Peek at the action_id without allocating it yet, so we don't break
+        // the rules<->actions bijection if the query turns out to be empty.
+        let action_id = self.rule_set.actions.next_id();
+        let plan = self.get_rule_with_extra_constraints(cached, action_id, extra_constraints)?;
+        // The query is non-empty: now commit the action and the plan.
```
oflatt, line 265: neat

core-relations/src/query.rs
```diff
@@ -134,128 +134,142 @@ impl<'outer> RuleSetBuilder<'outer> {
...
-                self.push_extra_constraints(&mut headers, &cached_plan.atoms, extra_constraints);
+                self.push_extra_constraints(&mut headers, &cached_plan.atoms, extra_constraints)?;
                 self.reprocess_existing_headers(
                     &mut headers,
                     &cached_plan.atoms,
                     &cached_plan.header,
-                );
-
-                // Process body blocks
+                )?;
```
oflatt, line 223: oh maybe option is used here

oflatt (approved):
lgtm, minor comment
great find!

egglog-bridge/src/rule.rs
yihozhang: Inlined `work` and wrote a description of how it does semi-naive here.

egglog-bridge/src/rule.rs
yihozhang: Yeah the returned option is nowhere used currently. But I'll keep it for now since it looks like it could be useful and this is a public facing function.

PR #856, opened 2026-04-27
ExectionState refactor for safety wrt semi-naive

oflatt:


tests/web-demo/multiset.egg
```diff
@@ -140,32 +140,26 @@
...
-    ((= outer (product inner)))
-    ((unstable-multiset-fill-index inner (unstable-fn "ms-count")))
-)
-(run 10)
-(check (= 2 (ms-count $xs (Num 1))))
-; Verify stays the same even if unioned and run again
-(union $zz $aa)
-(run 10)
+; Fill index table at top level (GlobalAction).
+(unstable-multiset-fill-index $xs (unstable-fn "ms-count"))
```
saulshanabrook: That makes sense!

I think if we wanted in this PR could just change fill index to not do the skip, and then it would work fine right as is? The skip was just a performance improvement but should have the right semantics without it.

oflatt: Unfortunately it seems like the skip is important to make it so repeated applications don't make it bigger. See my ai's response:

 I tried this — removed the skip, switched FillIndex back to a WritePrim so the rule pattern works again.
  But the existing test's assertions break in a way that's hard to fix without the skip:                     
                                                                                        
  (rule ((= outer (product inner)))                                                                          
        ((unstable-multiset-fill-index inner (unstable-fn "ms-count"))))                                     
  (run 10)                                                                                                   
  (check (= 2 (ms-count $xs (Num 1))))   ; ok: 2                                                             
                                                                                                             
  (union $zz $aa)                                                         
  (run 10)                                                                                                   
  (check (= 2 (ms-count $xs (Num 1))))   ; expects 2, gets 4              
                                                                                                             
  (union (Num 1) (Num 3))                                                                                    
  (check (= 3 (ms-count $xs (Num 1))))   ; expects 3, gets 6                                                 
                                                                                                             
  After (union $zz $aa), the rule re-fires on the unioned outer and inserts the same rows again; with :merge 
  (+ old new) that doubles the count to 4. Then the (Num 1) ↔ (Num 3) rebuild merges the duplicated entries
  instead of the original ones, so the third check goes to 6.                                                
                                                                          
  So the skip wasn't only a perf optimization — it was making FillIndex idempotent under repeated firings,   
  which the test's "stays the same after re-running" and "updates correctly on child union" assertions both
  rely on. Without it, the user has to either (a) use a merge function that's idempotent under re-insertion  
  (e.g. new rather than + old new) and rewrite the count math, or (b) gate the rule themselves so it only
  fires once per outer.

  Given that, I'd lean toward leaving FillIndex as a FullPrim (top-level only) for this PR and revisiting the
   rule-callable form separately — happy to do whichever you prefer though.

oflatt: Or maybe the problem is really the merge function with + old new?

oflatt: My opinion now is that seems like the skip part should be removed but the test case has an evil merge function

tests/factoring-multisets.egg
```diff
@@ -99,15 +99,15 @@
       ((union _n1 (polynomial _mss1))
        (delete (polynomial _mss))
        (set (get_sole_polynomial (multiset-of _n1)) _mss1))
-        :ruleset egglog.exp.array_api.to_polynomial_ruleset :name "unwrap monomial")
+        :ruleset egglog.exp.array_api.to_polynomial_ruleset :name "unwrap monomial" :naive)
```
saulshanabrook, line 102: Maybe easier to talk in person about this next week, but yeah I think I am still wondering if it's possible to infer this? Like without the `:naive` this will error right because it's unsafe?

oflatt: It is possible to infer this! I think it's a design choice we can consider.

I'm in favor of this because:
- makes performance more predicatable (doesn't get super slow when you change your primitive from pure to read-only)
- makes the code simpler

oflatt: changed to automatic plus warning

egglog-bridge/src/lib.rs
```diff
@@ -41,6 +41,55 @@ mod tests;
...
+///
+/// Maps table name to [`TableAction`] (plus the shared [`UnionAction`]
+/// and the default-panic external-function id) and is owned by the
+/// bridge `EGraph`. The state wrappers (`PureState`/`ReadState`/
+/// `WriteState`/`FullState`) live in the `egglog` crate; they read
+/// from this registry at invoke time to back name-indexed action
+/// methods. Held by the bridge `EGraph` inside an `Arc<RwLock<_>>` so
+/// `add_table` can mutate in place — full-registry clones on every
+/// declaration showed up as a hot path on programs that declare many
+/// constructors.
```
```suggestion
/// methods. Held by the bridge `EGraph` inside an `Arc<RwLock<_>>`.
```

egglog-bridge/src/lib.rs
```diff
@@ -452,9 +543,27 @@ impl EGraph {
         let info = &mut self.funcs[res];
         info.incremental_rebuild_rules = incremental_rebuild_rules;
         info.nonincremental_rebuild_rule = nonincremental_rebuild_rule;
+        let action = TableAction::new(self, res);
+        let table_name = self.funcs[res].name.to_string();
+        // Mutate in place — no full-registry clone. `add_table` is
+        // only called during single-threaded setup between top-level
+        // commands, so the `RwLock` write is uncontended.
```
```suggestion
```

core-relations/src/action/mod.rs
```diff
@@ -352,6 +352,47 @@ pub(crate) struct DbView<'a> {
 /// the next call to `merge` on the underlying table.
 ///
 /// ## Predicted Values
+/// The four contexts in which an external function may be invoked,
```
yihozhang: This new struct seems to be interjected in ExecutionState's documentation. Also, this is never referred in core-relations. Should it stay at a higher up crate?

src/prelude.rs
```diff
@@ -1,10 +1,27 @@
...
+//! - [`BaseSort`] / [`ContainerSort`] — declare custom sort types.
+//! - [`crate::PrimitiveCommon`] + one of four kind-specific traits
+//!   ([`crate::PurePrim`], [`crate::WritePrim`],
+//!   [`crate::ReadPrim`], [`crate::FullPrim`]) —
+//!   register custom primitives. Each kind names its state wrapper
+//!   directly; pick the trait matching what the body actually needs
+//!   and register via [`crate::EGraph::add_pure_primitive`]
+//!   etc. The Rust type checker enforces that the body only uses
+//!   methods the chosen state allows.
+//! - The [`add_primitive!`] / [`add_primitive_with_validator!`] /
```
yihozhang, line 16: add_primitive_with_validator seems gone

src/lib.rs
```diff
@@ -10,6 +10,40 @@
...
+//!   etc. The Rust type checker enforces that the body only uses
+//!   capabilities that wrapper exposes; the typechecker enforces that
+//!   the primitive is only invoked from contexts the wrapper is valid
+//!   in. See each kind trait's docs for the state→context details.
+//! - **Rust-bodied rule RHS**: register via [`prelude::rust_rule`]; the
+//!   closure receives an [`crate::WriteState`].
+//! - **Sorts and container types**: see [`Sort`], [`BaseSort`], and
+//!   [`ContainerSort`] (re-exported from the [`prelude`]).
+//!
+//! See issue #772 / `issue-772-proposal.md` in the repo for the
```
yihozhang:
```suggestion
//! See issue #772 in the repo for the
```

yihozhang (commented):
Did my first round of review and left some low-level comments. Will do another round later. I still need to digest the design around registering multiple copies of the same primitive and the changes to `ResolvedFunctionId`. I wish there is a better way to do this.

src/prelude.rs
oflatt: good catch, I got rid of it since I was refactoring anyways. I'll remove this part of the docs.

src/typechecking.rs
```diff
@@ -158,44 +313,181 @@ impl EGraph {
         }
     }
 
-    /// Add a user-defined primitive
-    pub fn add_primitive<T>(&mut self, x: T)
```
saulshanabrook, line 162: Could we leave this as an alias for `add_pure_primitive(..., None)` to reduce the diff?

saulshanabrook (commented):
Some notes from myself and @yihozhang. We also were saying if you are tired of dealing with this but do agree on these directions, we are open to helping:

- try removing four copy think, do lattice check during type checking. This requires having new type constraint
- Assume all code works with seminaive, but allow a manual flag to set naive=true for rules. If we get to calling an unstable fn in a query that does a read, then we panic if there is no naive flag set.  That way we don't have to infer whether to turn off seminaive for a rule but it will still always be safe.
- The same for reading in an action, if unstable fn is called with a read or full primitive at runtime in an action it should panic.


We were also thinking about what the minimal is to do before implementing better semi-naive support for unstable fn, so that it tracks all reads and only re-queries  when one of them changes. This assumes idempotent primitives and wouldn't be done in this PR but a follow up. At that point we could remove the need for the `naive` flag.

src/sort/vec.rs
```diff
@@ -285,7 +292,11 @@ struct Union {
     action: UnionAction,
 }
 
-impl Primitive for Union {
+// `Union` unions the corresponding entries of two vecs of equal length.
+// It writes to the union-find (via `UnionAction::union`), so it
+// implements `WritePrim` — valid in rule-action and global-action
+// contexts, rejected at rule-build time if used in a rule query.
+impl PrimitiveCommon for Union {
```
saulshanabrook: Is it possible to rename PrimitiveCommon to Primitive to reduce the diff again?

src/typechecking.rs
oflatt: we decided new name is clearer

src/sort/vec.rs
oflatt: samehere- new name is clearer, also users don't see this

tests/web-demo/multiset.egg
```diff
@@ -156,10 +155,6 @@
 )
 (run 10)
 (check (= 2 (ms-count $xs (Num 1))))
-; Verify stays the same even if unioned and run again
-(union $zz $aa)
-(run 10)
-(check (= 2 (ms-count $xs (Num 1))))
```
saulshanabrook, line 162: Does this need to be removed? I thought this would stil work?

oflatt: Oh right, I can revert this now that we handle reading and writing.

oflatt:
@copilot double check that there is no stale comments / docs in this pr and fix them. We refactored this pr a lot

saulshanabrook:
@coderabbitai full review

egglog-bridge/src/lib.rs
```diff
@@ -73,6 +119,14 @@ pub struct EGraph {
     /// bound.
     panic_funcs: HashMap<String, ExternalFunctionId>,
     report_level: ReportLevel,
+    /// Live registry of name-indexed action handles. Shared (via
+    /// `Arc<RwLock<_>>`) with state wrappers and primitive callbacks
+    /// in the egglog crate so name-indexed action methods on
+    /// [`WriteState`] / [`FullState`] can resolve table actions at
+    /// invoke time. Mutated in place from
+    /// [`add_table`](EGraph::add_table) to avoid cloning the entire
+    /// registry on every declaration.
```
```suggestion
    /// invoke time. Mutated in place from [`add_table`](EGraph::add_table).
```

egglog-bridge/src/lib.rs
```diff
@@ -189,6 +263,18 @@ impl EGraph {
...
+    ///
+    /// # Seminaive-safety trust boundary
+    ///
+    /// Like [`EGraph::with_execution_state`], this is a raw escape —
+    /// the registered function has unrestricted access and is not
+    /// tracked by the per-context validity system. Prefer building
+    /// primitives via the higher-level `egglog::Primitive` /
+    /// `egglog::EGraph::add_primitive` API (in the egglog crate, not
+    /// linkable from here), which enforces #772's seminaive-safety
+    /// contract.
```
```suggestion
    /// `egglog::EGraph::add_primitive` API, which enforces #772's
    /// seminaive-safety contract.
```

egglog-bridge/src/lib.rs
```diff
@@ -1228,6 +1357,13 @@ impl UnionAction {
         }
     }
 
+    /// Construct a `UnionAction` directly from its constituent ids.
+    /// Used by the bridge to build a default union action during
+    /// `EGraph::default()` (when `&self` does not yet exist).
+    pub(crate) fn from_parts(table: TableId, timestamp: CounterId) -> UnionAction {
```
saulshanabrook: Maybe more idiomatic to call it `new`?

egglog-bridge/src/lib.rs
```diff
@@ -452,9 +538,24 @@ impl EGraph {
...
+            .register_table(table_name, action);
         res
     }
 
+    /// A handle to the live [`ActionRegistry`] for this EGraph.
+    /// The handle is shared (`Arc<RwLock<_>>`); cloning the outer
+    /// `Arc` does not duplicate the underlying registry. Used by the
+    /// egglog crate's primitive machinery to thread the registry into
+    /// state wrappers at invoke time.
+    pub fn action_registry(&self) -> Arc<std::sync::RwLock<ActionRegistry>> {
```
yihozhang: I think you can return a reference to the registry here and the caller is responsible for cloning if needed.

src/sort/fn.rs
```diff
@@ -330,7 +347,12 @@ struct Ctor {
     function: Arc<FunctionSort>,
 }
 
-impl Primitive for Ctor {
+// `Ctor` (`unstable-fn "name" [...]`) builds a `FunctionContainer` and
+// interns it via `register_container`. Container interning is idempotent,
+// so it's safe in every context; declaring `State = PureState`
+// permits this primitive inside rule queries, actions, and global
+// contexts alike.
+impl PrimitiveCommon for Ctor {
```
saulshanabrook: Sorry if I already asked this, but can we keep this as `Primitive` just to reduce code change and keep things simpler? If not no worries

src/sort/fn.rs
```diff
@@ -409,18 +441,48 @@ impl BaseValue for ResolvedFunction {}
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub enum ResolvedFunctionId {
-    Lookup(egglog_bridge::TableAction),
-    Prim(ExternalFunctionId),
+    /// Wraps a constructor-table lookup. In a write-capable context
+    /// (`Write`/`Full`) `FunctionContainer::apply` mints a fresh
+    /// eclass via `lookup_or_insert`; in `Read` it does a read-only
+    /// `lookup`; in `Pure` it returns `None` (no DB ops).
```
yihozhang: Is `Pure` ever useful?

src/sort/fn.rs
```diff
@@ -409,18 +441,48 @@ impl BaseValue for ResolvedFunction {}
...
+    /// untracked seminaive read).
+    CustomLookup(egglog_bridge::TableAction),
+    /// Wraps a primitive. Carries every signature-matching
+    /// registration the typechecker found at build time, paired with
+    /// its `selection_ctx`. At dispatch time `FunctionContainer::apply`
+    /// picks the candidate whose `selection_ctx` matches the
+    /// application context — so the runtime selection is independent
+    /// of the build-site context, and an `unstable-fn` value may flow
+    /// freely from one context to another.
+    Prim {
```
saulshanabrook: Would renaming these to `Constructor`, `Function`, and `Primitive`, make sense? This is what they correspond to right? Any `(function ...)` is a `CustomLookup`? Or would a function with a custom `:merge` function not be `CustomLookup`?

src/sort/fn.rs
```diff
@@ -429,29 +491,82 @@ impl Primitive for Apply {
...
+                } else {
+                    mismatch(state)
+                }
+            }
+            ResolvedFunctionId::Prim { candidates } => {
+                // Pick the registration whose `selection_ctx` equals
+                // the application ctx. Each `add_*_primitive` commits
+                // disjoint singletons across the trait's valid ctxs,
+                // so exactly one candidate matches.
+                match candidates.iter().find(|(_, c)| *c == ctx) {
```
yihozhang: nit: add an assertion that it only matches one primitive

src/sort/fn.rs
```diff
@@ -429,29 +491,82 @@ impl Primitive for Apply {
...
-    /// Call function (primitive or table) `name` with value args `args` and return the value.
-    ///
-    /// Public so that other primitive sorts (external or internal) have access.
-    pub fn apply(&self, exec_state: &mut ExecutionState, args: &[Value]) -> Option<Value> {
+    /// Apply the wrapped function. `None` means the wrapped function
+    /// produced no result for this input; capability mismatch
+    /// triggers `panic_id` instead of silently returning `None`.
+    pub(crate) fn apply<'a, 'db>(
+        &self,
+        state: &mut crate::PureState<'a, 'db>,
```
saulshanabrook, line 514: Wait, I am confused, this is a `PureState`, but its context can be `Full`, `Write`, `Read`, or `Pure`?

src/lib.rs
```diff
@@ -10,6 +10,40 @@
...
+//!   capabilities that wrapper exposes; the typechecker enforces that
+//!   the primitive is only invoked from contexts the wrapper is valid
+//!   in. See each kind trait's docs for the state→context details.
+//! - **Rust-bodied rule RHS**: register via [`prelude::rust_rule`]; the
+//!   closure receives an [`crate::WriteState`].
+//! - **Sorts and container types**: see [`Sort`], [`BaseSort`], and
+//!   [`ContainerSort`] (re-exported from the [`prelude`]).
+//!
+//! See issue #772 in the repo for the seminaive-safety reasoning
+//! behind the typed primitive surface.
```
yihozhang:
```suggestion
```

src/lib.rs
```diff
@@ -95,21 +131,52 @@ pub const GLOBAL_NAME_PREFIX: &str = "$";
...
-/// A trait for implementing custom primitive operations in egglog.
+/// Methods shared by every kind-specific primitive trait.
 ///
-/// Primitives are built-in functions that can be called in both rule queries and actions.
-pub trait Primitive {
-    /// Returns the name of this primitive operation.
+/// `name` and `get_type_constraints` aren't capability-dependent, so
+/// the four kind-specific traits ([`PurePrim`], [`WritePrim`],
+/// [`ReadPrim`], [`FullPrim`]) require this as a
+/// supertrait rather than duplicating the methods.
```
yihozhang:
```suggestion
/// [`ReadPrim`], [`FullPrim`]) share this supertrait.
```

src/exec_state.rs
```diff
@@ -0,0 +1,479 @@
...
+    Read,
+    /// DB reads and writes both allowed. Top-level action-shaped
+    /// commands (`eval`, `let`, action-mode `run-schedule`) and the
+    /// head of a `:naive` rule.
+    Full,
+}
+
+impl Context {
+    pub const ALL: [Context; 4] = [Context::Pure, Context::Write, Context::Read, Context::Full];
+}
```
saulshanabrook: Not needed, but this pattern makes me wonder about a bitflags implementation to more easily compose and test against a set of contexts. It looks like we already depend on [bitflags](https://github.com/bitflags/bitflags) transitively if we wanna explore using it. [enumflags2](https://docs.rs/enumflags2/latest/enumflags2/) is another option that separates  a single flag from a set of flags at the type level.

src/exec_state.rs
saulshanabrook, line 1: Honestly this was a lot to review, kinda glazed over understanding all of it, but doesn't seem obviously wrong! Ofc if we could make it smaller that would be nice but OK to trust you that we need all this!

src/lib.rs
```diff
@@ -2154,15 +2311,21 @@ pub enum Error {
...
     struct InnerProduct {
         vec: ArcSort,
     }
 
-    impl Primitive for InnerProduct {
+    // Migrated to the new `Primitive` trait. `InnerProduct` is pure —
+    // it only reads base values and (idempotent) container contents — so it
+    // declares `State = PureState`, making it usable in all four
+    // contexts. The Rust type checker enforces that the body only uses
+    // methods available on `PureState`.
```
yihozhang:
```suggestion
    // `InnerProduct` is pure, so it declares
    // `State = PureState` and is usable in all
    // contexts. The Rust type checker enforces that the body only uses
    // methods available on `PureState`.
```

src/prelude.rs
yihozhang: It's still there

src/prelude.rs
```diff
@@ -1,10 +1,27 @@
...
+//!   [`add_literal_prim!`] macros (re-exported via `egglog::*`) cover
+//!   the "pure native function" case ergonomically — they generate
+//!   `PurePrim` impls and pull base-value conversions for you.
```
yihozhang:
```suggestion
//!   the simple "pure native function" case.
```

src/typechecking.rs
```diff
@@ -1,12 +1,114 @@
...
+// `T::apply(...)` is just one vtable hop plus a direct call — no
+// closure indirection that defeats inlining.
+#[derive(Clone)]
+struct PurePrimWrapper<T> {
+    prim: T,
+    /// The call-site [`Context`] this wrapper stamps onto the
+    /// `PureState` before dispatching. For non-HOF registrations the
+    /// natural choice is [`Context::Pure`] (fail-closed: HOF dispatch
+    /// via `apply_function` would refuse mint/read). For HOF
+    /// registrations, [`EGraph::add_higher_order_primitive`] registers
```
yihozhang: add_higher_order_primitive is gone? There are also two other occurrences of the reference.

src/typechecking.rs
```diff
@@ -1,12 +1,114 @@
...
+// `ExternalFunction` wrapper for primitives that need the
+// `ActionRegistry` (`ReadPrim`, `WritePrim`, `FullPrim`). One generic
+// over the `Wrap` strategy that knows how to construct the right
+// state type and dispatch to the primitive's `apply`.
+#[derive(Clone)]
+struct RegistryPrimWrapper<T, S> {
+    prim: T,
+    registry: Arc<RwLock<ActionRegistry>>,
+    /// Stamped onto the state wrapper. Same role as
+    /// [`PurePrimWrapper::ctx`].
```
yihozhang:
```suggestion
    /// Stamped onto the state wrapper.
```

src/sort/fn.rs
oflatt: Yeah sure

src/sort/fn.rs
oflatt: Yep- a purestate primitive is compatible with all these contexts

src/exec_state.rs
oflatt: good idea but this is a little clearer for me for now

yihozhang:
I haven't finished a full review, but I see there are also many repetitions about what each of the contexts does from a skim. Comments that contain en dashes likely have some repetitions.

saulshanabrook:
Bug from a local agent, global naive still raises error:

> 
> I think there is a concrete mismatch between global naive mode and rule-local `:naive`.
> 
> Minimal repro:
> 
> ```
> (datatype Math (Num i64))
> (sort Maths (MultiSet Math))
> (let $xs (multiset-of (Num 1) (Num 2)))
> 
> (constructor product (Maths) Math)
> (let $zz (product $xs))
> 
> (function ms-count (Maths Math) i64 :merge (+ old new))
> (sort MSIndexFn (UnstableFn (Maths Math) i64))
> 
> ;; No rule-local :naive.
> (rule
>     ((= outer (product inner)))
>     ((unstable-multiset-fill-index inner (unstable-fn "ms-count"))))
> 
> (run 1)
> (check (= 1 (ms-count $xs (Num 1))))
> ```
> 
> Run with global naive mode:
> 
> ```
> cargo run --features bin -- --naive /tmp/egg-smol-global-naive-repro.egg
> ```
> 
> Observed:
> 
> ```
> [ERROR] In 15:6-66 of egg-smol-global-naive-repro.egg: (unstable-multiset-fill-index inner (unstable-fn "ms-count"))
>     Unbound function unstable-multiset-fill-index
> ```
> 
> But the same rule succeeds if I remove --naive and add rule-local :naive:
> 
> ```
> (rule
>     ((= outer (product inner)))
>     ((unstable-multiset-fill-index inner (unstable-fn "ms-count")))
>     :naive)
> ```
> 
> So it looks like backend lowering respects global EGraph::seminaive = false, but typechecking still chooses the rule context only from local :naive.
>

src/sort/fn.rs
oflatt: What do you mean? Unstable fn can be used in pure contexts

src/sort/fn.rs
```diff
@@ -409,18 +441,48 @@ impl BaseValue for ResolvedFunction {}
...
+    /// registration the typechecker found at build time, paired with
+    /// its `selection_ctx`. At dispatch time `FunctionContainer::apply`
+    /// picks the candidate whose `selection_ctx` matches the
+    /// application context — so the runtime selection is independent
+    /// of the build-site context, and an `unstable-fn` value may flow
+    /// freely from one context to another.
+    Prim {
+        candidates: Vec<(ExternalFunctionId, crate::Context)>,
+    },
 }
```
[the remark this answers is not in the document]
oflatt: done

src/prelude.rs
oflatt: Actually we still have that as a macro

src/sort/fn.rs
oflatt: yep, renamed

src/exec_state.rs
oflatt: ok!

oflatt:
Nice, fixed bug

src/sort/fn.rs
```diff
@@ -192,14 +200,24 @@ impl Sort for FunctionSort {
...
-            function: self.clone(),
-        });
+        eg.add_pure_primitive(
+            Ctor {
+                name: "unstable-fn".into(),
+                function: self.clone(),
+            },
+            None,
+        );
+        // Higher-order primitive: registers four `ExternalFunctionId`s
```
yihozhang: stale comment

src/typechecking.rs
```diff
@@ -158,44 +261,114 @@ impl EGraph {
...
+            Box::new(RegistryPrimWrapper::<T, S> {
+                prim: x,
+                registry: registry.clone(),
+                ctx,
+                _wrap: std::marker::PhantomData,
+            })
+        });
+    }
+
+    /// Shared registration engine. Stores one primitive definition, plus
```
yihozhang, line 332: stale?

yihozhang (approved):
Looks good to me! Let's merge this once it also looks good to @saulshanabrook

saulshanabrook (approved):

src/typechecking.rs
oflatt: I think this one is right

PR #857, opened 2026-04-28
Factorize binding representations

yihozhang:
For benchmarks like cykjson, the join has the following patterns

```
R(a, b) x S(b, c)
```

where for the same b, there are many `a`s and `c`s. The existing join implementation is not efficient because it will call run_plan on the cartesian product of `a ' and `c`, which can waste time in index building and create quadratically many Arc<Trie>.

I think with this optimization we may be able to revert several other optimizations, or they may not be as important, such as #840 and some ugly ones in #852 or #855 .

### DIFF
--- a/core-relations/src/free_join/execute.rs
+++ b/core-relations/src/free_join/execute.rs
@@ -960,7 +960,6 @@ impl<'a> JoinState<'a> {
         if cur >= instr_order.len() {
             action_buf.push_bindings_factorized(
                 action,
-                self.db,
                 &mut binding_info.bindings,
                 &binding_info.binding_sets,
                 || self.exec_state.clone(),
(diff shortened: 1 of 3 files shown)


ezrosent:
These numbers are awesome and this looks like a big win, but I'm wondering if this is evidence that `run_plan` is just way too expensive? It looks like we are still expanding the cartesian product, just doing it at a much lower level in the stack: should we be looking for ways to remove this overhead in non-"leaf" plan nodes?

yihozhang:
> if this is evidence that run_plan is just way too expensive

Definitely! We are essentially creating one Arc<TrieNode> per match currently, and this is not good. There are other overheads due to run_plan being run over a singleton subset `Dense(row, row.inc())`. I think it makes sense to have dedicated code paths for the last layer of the trie, but am interested in learning more about what's your thoughts on removing this overhead in non-leaf nodes

ezrosent (approved):

PR #858, opened 2026-04-28
add shared_arena module

ezrosent:
We may want to use an abstraction like this in query execution to cut down on `Arc` allocations.

### DIFF
--- a/concurrency/src/shared_arena.rs
+++ b/concurrency/src/shared_arena.rs
@@ -1,0 +1,298 @@
+//! Scoped thread-safe arena allocation.
+//!
+//! This module provides [`SharedArena`], a bump-allocation arena that can be
+//! shared by reference across scoped worker threads. Each worker creates a
+//! non-`Send`, non-`Sync` [`Handle`] and performs unsynchronized allocations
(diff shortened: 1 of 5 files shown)


concurrency/src/shared_arena.rs
```diff
@@ -0,0 +1,298 @@
...
+        }
+    }
+}
+
+struct DropEntry {
+    ptr: NonNull<()>,
+    drop: unsafe fn(*mut ()),
+}
+
+unsafe fn drop_value<T>(ptr: *mut ()) {
```
yihozhang, line 293: Does T need to be 'static? Not sure

yihozhang (approved):

oflatt:
I haven't read the code closely but seems fine to me, will review any follow up that uses it

concurrency/src/shared_arena.rs
ezrosent: `drop_value` isn't safe to call in general, but as for the places we call it in this module, `T` need only live as long as `'arena`, which is guaranteed by the bound on the `alloc` method.

PR #859, opened 2026-04-28
Claude experiment day3

yihozhang:


core-relations/src/hash_index/mod.rs
```diff
@@ -163,7 +163,7 @@ pub(crate) trait IndexBase {
 }
 
 struct ColumnIndexShard {
-    table: Pooled<IndexMap<Value, BufferedSubset>>,
+    table: Pooled<HashMap<Value, BufferedSubset>>,
```
yihozhang, line 166: Reverting this change causes regression, but one concern may be if this will affect determinism.

```

  hardboiled_conv1d_32.egg                        0.160       0.166  +  0.006  +   3.8%  ▲ slower
  hardboiled_conv1d_128.egg                       0.444       0.465  +  0.021  +   4.7%  ▲ slower
  luminal-llama.egg                               0.089       0.090  +  0.001  +   1.1%  ▲ slower
  python_array_optimize.egg                       0.260       0.264  +  0.004  +   1.5%  ▲ slower
  cykjson.egg                                     0.054       0.056  +  0.002  +   3.7%  ▲ slower
  eggcc-extraction.egg                            0.229       0.236  +  0.007  +   3.1%  ▲ slower
```

ezrosent (approved):

PR #861, opened 2026-04-30
Add more container/fn tests

saulshanabrook:
add test and benchmark for graphics example from https://uwplse.org/2026/02/24/egglog-containers.html and update Python example to match newest python source, that includes more intentional container use.

Useful to test higher order functions and matching on containers


oflatt (approved):
lgtm!

PR #862, opened 2026-04-30
Claude experiment day 4

yihozhang:


core-relations/src/free_join/execute.rs
```diff
@@ -54,6 +54,18 @@ struct SparseColumnIndex {
     subset_ids: [RowId; SMALL_RESIDUAL],
 }
 
+/// Return a SubsetRef for the given range of rows in a SparseColumnIndex.
+/// Single-row ranges become Dense to skip pool allocation in to_owned.
+#[inline]
+fn sparse_subset_ref(ids: &[RowId], range: Range<usize>) -> SubsetRef<'_> {
```
ezrosent: Looks like we don't have a guarantee at the callsite that `ids` is sorted. I'd suggest either making this a method on `SubsetRef`, or make this method `unsafe` (with explicit safety comments at every callsite).

core-relations/src/hash_index/mod.rs
```diff
@@ -344,6 +416,76 @@ fn run_in_thread_pool_and_block<'a>(pool: &rayon::ThreadPool, f: impl FnMut() +
...
 
+/// Adaptive LSB radix sort for (Value, RowId) pairs, sorting by the Value field.
+///
+/// Chooses 1–4 passes of 8-bit radix sort based on the observed maximum Value, so that
+/// only the actually-used bit range is processed.  Within each Value group the original
+/// order (scan order = RowId-ascending) is preserved by the LSB stability guarantee,
+/// satisfying the add_row_sorted invariant without an explicit RowId sort.
+///
+/// Falls back to `sort_unstable()` for n < 64 where radix setup overhead dominates.
+#[allow(clippy::needless_range_loop)]
```
ezrosent: why are we allowing this lint?

core-relations/src/hash_index/mod.rs
```diff
@@ -344,6 +416,76 @@ fn run_in_thread_pool_and_block<'a>(pool: &rayon::ThreadPool, f: impl FnMut() +
...
+    } else if max_val < 65_536 {
+        2
+    } else if max_val < (1 << 24) {
+        3
+    } else {
+        4
+    };
+
+    let null_pair = (Value::new_const(0), RowId::new_const(0));
+    let mut buf: Vec<(Value, RowId)> = vec![null_pair; n];
```
ezrosent, line 446: Why aren't we doing in-place radix sort?

core-relations/src/table_spec.rs
```diff
@@ -405,11 +405,21 @@ impl<T: Table> TableWrapper for WrapperImpl<T> {
         })
     }
     fn group_by_col(&self, table: &dyn Table, subset: SubsetRef, col: ColumnId) -> ColumnIndex {
-        let table = table.as_any().downcast_ref::<T>().unwrap();
         let mut res = ColumnIndex::new();
-        table.scan_generic(subset, |row_id, row| {
-            res.add_row(&[row[col.index()]], row_id);
-        });
+        if subset.size() >= 512 {
```
ezrosent: Let's have this size-gating heuristic in the index module

ezrosent (changes requested):

core-relations/src/hash_index/mod.rs
yihozhang: I tried a version of in-place radix sort (https://en.wikipedia.org/wiki/American_flag_sort) and it caused a 10-20% degration on HardBoiled benchmarks.

ezrosent (approved):

PR #867, opened 2026-05-06
Return simplified proofs

saulshanabrook:
I believe there is a bug currently where it simplifies the proof and checks that simplification but still returns the un simplified version.

This fixes it and updates the snapshots which should now show the simplified version.


oflatt (approved):
nice lol

PR #868, opened 2026-05-07
Use region-based allocator for TrieNodes

yihozhang:


yihozhang:
Below is what the agent found. It's worth noting this PR does not have the change from #857, so it creates many Dense singleton subsets that would otherwise be eliminated in #857. Still, it's slightly slower than #857 after incorporating its change.

<details>
<summary>Agent's finding</summary>

```markdown

Investigation summary — why arena is slower

  Repro (with tests/cykjson.egg switched to the medium token, as you had locally):

  ┌──────────────────┬───────┬───────┬───────┬─────────┐
  │                  │ wall  │ user  │  sys  │   RSS   │
  ├──────────────────┼───────┼───────┼───────┼─────────┤
  │ baseline 23dd731 │ 5.7 s │ 5.8 s │ 0.0 s │ 20 MB   │
  ├──────────────────┼───────┼───────┼───────┼─────────┤
  │ arena HEAD       │ 8.4 s │ 5.8 s │ 2.5 s │ 2.28 GB │
  └──────────────────┴───────┴───────┴───────┴─────────┘
  
  CPU work (user time) is essentially unchanged. Almost all of the regression is from:
  - +2.5 s system time from page-faulting / mmap as the arena grows
  - ~100× memory bloat (20 MB → 2.28 GB) causing TLB/cache pressure
  
  Root cause: transient TrieNodes leak for the entire plan lifetime

  I instrumented every handle.alloc call site. For a single run of cykjson (medium):

  plans=528
  header                =          0
  insert_subset         = 116,272,833   <-- 99.997% of allocations
  refine_nondense       =      3,100
  get_cached_trie_node  =      2,330   (216 cache hits)
  new_handle_calls      =          0   (single-thread, parallel-drain path unused)

  Almost every TrieNode that lives in the arena is a transient wrapper around a Subset::Dense(range) produced by RefineAtomDense in
  the drain at execute.rs:1015-1016. In the Arc-based baseline these were Arc::new(TrieNode::new(...)) and their refcount dropped to
   zero the moment the next refinement / next frame overwrote binding_info.subsets[atom], so memory was reused continuously. With
  the arena, all 116 million of them stay alive until the per-plan SharedArena is dropped at end of plan.

  The cached_children work the design was trying to optimize (get_cached_trie_node) is fewer than 3,000 allocs in this benchmark —
  that's not the hot path.

  Secondary findings (perf profile)

  In the arena run, ~10% of cycles is direct arena machinery, plus the ~6 % kernel time:
  the drain at execute.rs:1015-1016. In the Arc-based baseline these were Arc::new(TrieNode::new(...)) and their refcount dropped to
   zero the moment the next refinement / next frame overwrote binding_info.subsets[atom], so memory was reused continuously. With
  the arena, all 116 million of them stay alive until the per-plan SharedArena is dropped at end of plan.

  The cached_children work the design was trying to optimize (get_cached_trie_node) is fewer than 3,000 allocs in this benchmark —
  that's not the hot path.

  Secondary findings (perf profile)

  In the arena run, ~10% of cycles is direct arena machinery, plus the ~6 % kernel time:

    5.99%  egglog_concurrency::shared_arena::Handle::alloc
    2.07%  <egglog_concurrency::shared_arena::LocalArena as Drop>::drop
    1.90%  core::ptr::drop_in_place<...::TrieNode>
    ~6%    kernel (page faults / mmap)

  For comparison, the Arc baseline spent ~6.6% in mimalloc + Arc::drop_slow doing the same allocation work plus immediate frees. The
   bumpalo arena isn't beating mimalloc on per-alloc cost here, and it pays the deferred-drop tax on top.

  Also worth flagging (latent bug, not hit by the single-thread cykjson repro): in drain_updates_parallel at execute.rs:1076,
  &arena.new_handle() is called per RefineAtomDense update. Each new_handle() takes a Mutex lock and Box::pin(LocalArena::new()). In
   multi-threaded runs this will both contend and accumulate one LocalArena per update.

  Suggested fixes (in priority order)

  1. Don't allocate a TrieNode in insert_subset for the dense fast path. This is what's blowing up. Options:
    - Change BindingInfo.subsets to a DenseIdMap<AtomId, NodeSlot<'arena>> enum with Subset(Subset) and Node(SharedRef<...>)
  variants. Lazily promote to Node only when get_cached_trie_node actually needs a stable parent (which happened only ~2,500 times
  vs 116 M).
    - Or: switch BindingInfo.subsets to store the TrieNode by value (it's a single-occupancy slot anyway). Cached-child storage in
  cached_children still uses SharedRef.
  2. In drain_updates_parallel, hoist arena.new_handle() out of the per-update loop — create one handle per recur task and reuse it.
   The current code mutex-locks once per dense refinement.
  3. If the arena lifetime is going to stay "per plan," the perf model assumed by Pooled<ChildrenMaps> in pool/mod.rs:445 is broken
  too: those maps are returned to the pool only on TrieNode drop, which now never happens mid-plan, so the pool can't recycle them
  within a plan. Fix #1 makes this moot for transients; for the genuinely-cached ones it just means the pool is effectively unused.

  The arena design is sound for the cached TrieNodes (the ones in cached_children), but it's currently being used for the wrong set
  of nodes. Once fix #1 lands, the arena's job is to hold ~thousands of long-lived shared TrieNodes per plan instead of ~hundreds of
   millions of single-use ones, and the Arc-based atomic refcounting it eliminates should actually become measurable savings.
```

</details>

PR #874, opened 2026-05-14
Clean up ExecutionState Refactor PR

saulshanabrook:
Changes to #856 to:

* Register only one primitive instead of four, one per context
* Change to use EnumMap instead of Vec<(Enum, T))> for clearer semantics
* Clean up the resolution for unstable function to make the errors more explicit

I pushed this branch by accident to Oliver's branch so the old PR https://github.com/egraphs-good/egglog/pull/873 got closed. This is a replacement for that one.

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -160,40 +160,23 @@ impl ResolvedCall {
             }
         }
 
-        // For a given (signature, context) pair, exactly one primitive
-        // definition must match. Each definition carries the runtime ids
-        // for the contexts it permits. If two definitions accept the
-        // same signature in the same context, registration was
-        // ambiguous (e.g. the same primitive added twice); panic here
-        // rather than silently picking one.
-        if let Some(primitives) = typeinfo.get_prims(head) {
-            let matches: Vec<_> = primitives
-                .iter()
-                .filter(|p| p.context_ids[ctx].is_some() && p.accept(types, typeinfo))
-                .collect();
-            match matches.as_slice() {
-                [] => {}
-                [picked] => {
-                    let (out, inp) = types.split_last().unwrap();
-                    return ResolvedCall::Primitive(SpecializedPrimitive {
-                        prim_with_id: (*picked).clone(),
-                        input: inp.to_vec(),
-                        output: out.clone(),
-                    });
-                }
-                multiple => {
-                    let sig: Vec<_> = types.iter().map(|s| s.name().to_string()).collect();
-                    panic!(
-                        "Ambiguous primitive resolution for {head:?} in context {ctx:?} \
-                         with signature [{}]: {} registrations accept this call. \
-                         Each (name, signature, context) tuple must map to a unique \
-                         primitive — was {head:?} registered more than once with the \
-                         same type?",
-                        sig.join(", "),
-                        multiple.len(),
-                    );
-                }
+        let mut primitives = typeinfo
+            .get_prims(head)
+            .into_iter()
+            .flatten()
+            .filter(|p| p.context_ids[ctx].is_some() && p.accept(types, typeinfo));
+        if let Some(picked) = primitives.next() {
+            if primitives.next().is_some() {
+                panic!(
+                    "Ambiguous primitive resolution for {head:?} in direct call context {ctx:?}"
+                );
             }
+            let (out, inp) = types.split_last().unwrap();
+            return ResolvedCall::Primitive(SpecializedPrimitive {
+                prim_with_id: picked.clone(),
+                input: inp.to_vec(),
+                output: out.clone(),
+            });
         }
 
         panic!("No resolution for {head:?} in context {ctx:?}");
--- a/src/core.rs
+++ b/src/core.rs
@@ -81,7 +81,20 @@ impl SpecializedPrimitive {
 
 impl PartialEq for SpecializedPrimitive {
     fn eq(&self, other: &Self) -> bool {
+        // This is the key used when resolved atoms are deduplicated by
+        // `(head, inputs)`. The context-id map identifies the primitive
+        // registration, while the concrete input/output sorts identify the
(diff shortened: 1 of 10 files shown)


src/sort/fn.rs
```diff
@@ -450,20 +451,18 @@ pub enum ResolvedFunctionId {
...
-    /// freely from one context to another.
+    /// Wraps a primitive. Carries the unique exact-signature runtime
+    /// id found for each context at build time. At dispatch time
+    /// `FunctionContainer::apply` picks the id for the application
+    /// context — so the runtime selection is independent of the
+    /// build-site context, and an `unstable-fn` value may flow freely
+    /// from one context to another.
     Primitive {
-        candidates: Vec<(ExternalFunctionId, crate::Context)>,
+        context_ids: EnumMap<crate::Context, Option<ExternalFunctionId>>,
```
oflatt, line 465: why do you need an option here? I'm confused about that

oflatt: I know the doc comment was confusing before, it's still confusing haha

src/core.rs
```diff
@@ -76,15 +81,15 @@ impl SpecializedPrimitive {
 
 impl PartialEq for SpecializedPrimitive {
     fn eq(&self, other: &Self) -> bool {
-        self.prim_with_id.id == other.prim_with_id.id
+        self.prim_with_id.context_ids == other.prim_with_id.context_ids
```
oflatt, line 90: why did you end up doing all context ids instead of first one here? needs docs?

src/core.rs
```diff
@@ -155,41 +160,23 @@ impl ResolvedCall {
...
-                        "Ambiguous primitive resolution for {head:?} in context {ctx:?} \
-                         with signature [{}]: {} registrations accept this call. \
-                         Each (name, signature, context) tuple must map to a unique \
-                         primitive — was {head:?} registered more than once with the \
-                         same type?",
-                        sig.join(", "),
-                        multiple.len(),
-                    );
-                }
+        let mut primitives = typeinfo
```
oflatt, line 181: nice

src/lib.rs
```diff
@@ -702,7 +702,7 @@ impl EGraph {
                     .map(|arg| self.translate_expr_to_mergefn(arg))
                     .collect::<Result<Vec<_>, _>>()?;
                 Ok(egglog_bridge::MergeFn::Primitive(
-                    p.external_id(),
+                    p.external_id(crate::Context::Full),
```
oflatt: err hmm, I wonder if merge fns should be write only

src/typechecking.rs
```diff
@@ -155,14 +156,12 @@ pub type PrimitiveValidator = Arc<dyn Fn(&mut TermDag, &[TermId]) -> Option<Term
...
-    /// `add_*_primitive` call commits one registration per context
-    /// the trait permits, so the constraint solver filters by
-    /// `selection_ctx == ambient_ctx` to pick exactly one id per
-    /// call site.
-    pub(crate) selection_ctx: Context,
+    /// Runtime entrypoints for the contexts this primitive is valid in.
+    /// The primitive definition is stored once, while each context keeps
+    /// its own backend id so higher-order dispatch can still recover the
+    /// application context at runtime.
+    pub(crate) context_ids: EnumMap<Context, Option<ExternalFunctionId>>,
```
oflatt, line 164: nice

oflatt (changes requested):
looking real good, some lower level concerns 
the merge fn one I spotted is worth some consideration

src/sort/fn.rs
saulshanabrook: An enun map is total, which is why we need the option! (i.e. this is the same as the old vec? Every context may not have an external function id?)

src/lib.rs
saulshanabrook: Would it make sense to have the merge function context be the same as the context under which the merge was triggered under? Like which could only be Write or Full, right? Or just make it always Write for simplicity?

src/core.rs
saulshanabrook: Thanks for this! I actually added a bit more data and a comment to justify it after looking into it further.

src/sort/fn.rs
oflatt: Oh I see, that makes sense

src/lib.rs
oflatt: I think that sounds sketchy, since the same merge function could be triggered by actions or global actions. Write is my vote

saulshanabrook: Will do! Refactoring this with the type checking, I am also realizing that other top level actions have been like Full I believe, like extract? But Extract should probably be read or write or should it be full?

oflatt: Hmm I guess extract can be Full? Right now it's just a global action like `(let ...)` at the top level

src/core.rs
oflatt: neat- I wonder if there's a library to exclude certain things from the eq implementation instead of listing everything that should be included

src/lib.rs
saulshanabrook: Fixed!

src/typechecking.rs
```diff
@@ -451,35 +425,44 @@ impl EGraph {
...
-            NCommand::CoreAction(Action::Let(span, var, expr)) => {
-                let expr = self
-                    .type_info
-                    .typecheck_expr(symbol_gen, expr, &Default::default())?;
-                let output_type = expr.output_type();
+            NCommand::CoreAction(action @ Action::Let(span, var, _)) => {
+                let action = self.type_info.typecheck_standalone_action(
+                    symbol_gen,
+                    action,
+                    &Default::default(),
```
saulshanabrook, line 432: This was an unrelated change to fix how let actions at the top level are typechecked, to go through typecheck action instead of expr

PR #877, opened 2026-05-16
Cache source line starts for span formatting

oflatt:
## Summary
- Cache line-start offsets in SrcFile so repeated Span formatting no longer scans from the start of large source files.
- Route parser source construction through SrcFile::new.

## Motivation
Profiling --term-encoding on the reduced eggcc fixture showed startup time dominated by formatting generated rules: Span::Display repeatedly called SrcFile::get_location, which rescanned the whole file prefix for each span.

## Testing
- cargo test -p egglog-ast

### DIFF
--- a/egglog-ast/src/span.rs
+++ b/egglog-ast/src/span.rs
@@ -1,5 +1,6 @@
 use std::fmt::{self, Debug, Display};
-use std::sync::Arc;
+use std::hash::{Hash, Hasher};
+use std::sync::{Arc, OnceLock};
 
 #[derive(Clone, PartialEq, Eq, Hash)]
 pub enum Span {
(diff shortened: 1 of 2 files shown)


saulshanabrook:
What if we just change the `EgglogSpan` to store the line and column number for start and stop values, instead of storing the offset into the file regardless of line number? Then we don't have to recompute this each time to display?

oflatt:
Closing this.

Digging into *why* `Span::Display` was hot under `--term-encoding`: it traces back to a regression in my downstream fork, not to `main`. The fork added a backend-trait abstraction whose `lookup` / `call_external_func` took an eager `panic_msg: String`, and the bridge impl wrapped it in `move || msg` — so `format!("{span}: …")` (which scans the source via `get_location`) ran eagerly for every RHS lookup/call at rule-build time. With term/proof encoding generating thousands of rules, that's O(rules × file).

`main` doesn't have that layer: `BackendRule::actions` passes `move || format!(…)` straight to `RuleBuilder` (already lazy), so the rule-building span-format cost this PR was reacting to doesn't exist here. I've fixed it downstream by restoring the lazy `FnOnce` through the trait.

The `line_starts` cache is still a reasonable standalone speedup for genuine `Span::Display` on large files, so happy to revive it if that path proves to matter — but closing for now since the driving motivation was the fork-side regression.

PR #880, opened 2026-05-17
fix bag merging during topological sort for tree decomposition

### LINKED ISSUE #866: Query performance depends on when rule is added
Attached an egglog file from herbie that times out: [rewrite24.txt](https://github.com/user-attachments/files/27385727/rewrite24.txt).
The rule that causes this is a large lowering rule (below). For this benchmark, the rule never matches anything, but it spends too long trying to match and times out. After running some experiments:
- When running one iter of `lower` at a time:
  - Iters 1-3 take <2s.
  - Iter 4 takes ~1m.
  - Iter 5 times out completely.
- However, if I remove this rule and saturate, it finishes in <1s.
  - If I then add the rule to `lower` and saturate again, it again finishes in <1s.

So, the performance of this rule seems to depend on when it is added to the ruleset. One hypothesis is that when doing query planning, the optimizer assumes that cardinalities of tables will remain approximately the same. In Herbie, before lowering, the `MTy` sort is almost empty, with many constructors with empty tables, and these tables are filled in by the lower ruleset, so this hypothesis would explain the timeout - query optimizer could see some empty tables and join them in a way that becomes bad after they grow; but, I'm not very familiar with how egglog does query planning.

```
(rule
  ((=
      ?root
      (Mul
        (Mul
          (Mul
            (Div
              (Div
                (Mul
                  (Div (Mul sgn J2) (Num (bigrat (from-string "2") (from-string "1"))))
                  (Pow (Div req a) (Num (bigrat (from-string "2") (from-string "1")))))
                (Pow
                  (Sqrt
                    (Sub
                      (Num (bigrat (from-string "1") (from-string "1")))
                      (Pow e (Num (bigrat (from-string "2") (from-string "1"))))))
                  (Num (bigrat (from-string "4") (from-string "1")))))
              (Num (bigrat (from-string "8") (from-string "1"))))
            e)
          (Pow
            (Sqrt
              (Sub
                (Num (bigrat (from-string "1") (from-string "1")))
                (Pow e (Num (bigrat (from-string "2") (from-string "1"))))))
            (Num (bigrat (from-string "2") (from-string "1")))))
        (Mul
          (Add
            (Sub
              (Num (bigrat (from-string "1") (from-string "1")))
              (Mul
                (Num (bigrat (from-string "11") (from-string "1")))
                (Pow (Cos i) (Num (bigrat (from-string "2") (from-string "1"))))))
            (Neg
              (Div
                (Mul
                  (Num (bigrat (from-string "40") (from-string "1")))
                  (Pow (Cos i) (Num (bigrat (from-string "4") (from-string "1")))))
                (Sub
                  (Num (bigrat (from-string "1") (from-string "1")))
                  (Mul
                    (Num (bigrat (from-string "5") (from-string "1")))
                    (Pow (Cos i) (Num (bigrat (from-string "2") (from-string "1")))))))))
          (Cos (Mul (Num (bigrat (from-string "2") (from-string "1"))) omega)))))
    (= te (do-lower e "sym_binary64"))
    (= ti (do-lower i "sym_binary64"))
    (= tomega (do-lower omega "sym_binary64"))
    (= tsgn (do-lower sgn "sym_binary64"))
    (= tJ2 (do-lower J2 "sym_binary64"))
    (= treq (do-lower req "sym_binary64"))
    (= ta (do-lower a "sym_binary64")))
  ((union
      (do-lower ?root "sym_binary64")
      (Clmeanoscmap-De1Ty te ti tomega tsgn tJ2 treq ta)))
  :ruleset
  lower)
```

yihozhang:
A bug in the current topological sort after tree decomposition only merged immediate non-first child bags, ignoring their descendants. This left some bags with multiple children, causing the Cartesian product of bags, which leads to #866 . This PR ensures all children are fully merged into the parent, so the tree decomposition always produces a chain.

However, this makes the hardboiled_conv1d_128 benchmark significantly slower. This benchmark favors less collapsing between bags, because one of the bags has cardinality 0 during execution, but can be very slow if that bag is merged with others to be a super big bag. The proper tree decomposition -> path decomposition conversion makes this query run very slow because it collapses more bags.

Therefore, this PR also removed the heuristics before that tried to pick a child that minimizes # of overlapping variables. Instead, it does a tree DP that allows it to always pick the child that minimizes the overall pathwidth of the resulting path decomposition. This makes it produce the path decomposition with the smallest pathwidth corresponding to the tree decomposition.

Fixes #866


yihozhang:
This PR makes the existing benchmarks slightly slower, but makes the new herbie-repro's runtime go from minutes to 0.2s

```
  hardboiled_conv1d_32.egg                        0.179       0.191  +  0.012  +   6.5%  ▲ slower
  hardboiled_conv1d_128.egg                       0.515       0.496   -0.019    -3.7%  ▼ faster
  luminal-llama.egg                               0.103       0.104  +  0.001  +   0.9%  ▲ slower
  python_array_optimize.egg                       0.316       0.324  +  0.008  +   2.5%  ▲ slower
  cykjson.egg                                     0.066       0.063   -0.003    -4.6%  ▼ faster
  eggcc-extraction.egg                            0.269       0.274  +  0.005  +   1.8%  ▲ slower

  Summary: 2 faster  ·  4 slower  ·  0 unchanged  ·  0 missing
  Overall average Δ: +0.57%
```

core-relations/src/free_join/plan.rs
```diff
@@ -699,18 +699,67 @@ fn decompose_into_bags(original_ctx: &PlanningContext) -> Vec<PlanningContext> {
...
+    let mut all_children_list: Vec<Vec<usize>> = vec![vec![]; bags.len()];
+    // best_pathwidth[i] = the best pathwidth of the chain if we pick bag i
+    // to be the chain child.
+    let mut best_pathwidth = vec![usize::MAX; bags.len()];
+    let mut full = vec![HashSet::default(); bags.len()];
+    let mut choice = vec![usize::MAX; bags.len()];
+    for i in 0..bags.len() {
+        full[i] = bags[i].atoms.iter().map(|(atom_id, _)| atom_id).collect();
+        for child in all_children_list[i].iter() {
+            full[i] = full[i].union(&full[*child]).copied().collect();
```
ezrosent: is this the same as `full[i].extend(&full[child])` (or similar?)

core-relations/src/free_join/plan.rs
```diff
@@ -699,18 +699,67 @@ fn decompose_into_bags(original_ctx: &PlanningContext) -> Vec<PlanningContext> {
...
+        }
+        best_pathwidth[i] = full[i].len();
+        for chain_child in all_children_list[i].iter() {
+            let mut chain_score: HashSet<_> =
+                bags[i].atoms.iter().map(|(atom_id, _)| atom_id).collect();
+            for child in all_children_list[i].iter() {
```
ezrosent: similar: can this be an extend(all_children_list[i].flat_map(...) ) ?

core-relations/src/free_join/plan.rs
```diff
@@ -699,18 +699,67 @@ fn decompose_into_bags(original_ctx: &PlanningContext) -> Vec<PlanningContext> {
...
-/// This ensures that we evaluate bags in order.
-///
-/// This method also merges bags. The case where a bag has multiple children in the tree decomposition
-/// has terrible performance currently, because all children but one has to be an innermost lookup.
-/// So in this function, we merge all the non-first children into the parent. This improves HardBoiled benchmarks
-/// significantly.
+/// At every DFS node we pick one child as the chain continuation. Every other reachable bag —
+/// siblings *and* their entire sub-trees — gets absorbed into the current chain node. The
+/// continuation is picked in a way that minimizes the maximum number of atoms in a bag, i.e.,
+/// the pathwidth.
```
ezrosent, line 701: I think explaining the pathwidth is would improve clarity here.

ezrosent (approved):

PR #881, opened 2026-05-18
Add primitive body typecheck helpers

saulshanabrook:
## Context

This is the egglog-side support needed by https://github.com/egraphs-good/egglog-experimental/pull/46 for body-defined `(primitive ...)` commands. It intentionally contains reusable core APIs only; the experimental command implementation stays in egglog-experimental.

## Current shape

- Adds `EGraph::typecheck_expr_with_bindings_and_output` for expression typechecking with explicit local bindings, an expected output sort, and a runtime primitive context.
- Ensures the output-constrained helper rejects resolved expressions whose final output sort still differs from the requested output sort, including literal bodies.
- Rewrites global references returned by that helper into the same zero-argument function calls used by normal command execution, so direct runtime evaluation can handle globals consistently.
- Adds `Core::eval_resolved_expr` so primitive bodies can evaluate already-typechecked and prepared expressions through the current execution state.
- Adds `Core::apply_primitive` for evaluating an already-specialized primitive against runtime values.
- Exposes `PrimitiveWithId::is_valid_in_context` so downstream code can infer the minimum runtime context for a resolved primitive target without reaching into the internal context-id map.
- Adds `EGraph::prepare_unstable_fn_targets_for_eval`, a narrow bridge that rewrites literal `(unstable-fn "...")` targets into hidden bindings containing the same `ResolvedFunction` value that backend rule lowering injects.
- Keeps top-level `EGraph::eval_resolved_expr` on the existing backend-rule execution path.
- Adds focused tests and an unreleased changelog entry.

## Tried and decided against

- Tried a general `PreparedResolvedExpr` / `ResolvedCoreActions` bridge for direct runtime expression evaluation. Dropped it because it introduced a broader public concept and action-interpreter surface just to support one hidden `unstable-fn` runtime value.
- Considered migrating top-level `EGraph::eval_resolved_expr` to direct `ExecutionState` evaluation. Kept the backend-rule path instead, since top-level eval already gets the full lowering behavior there and does not need to share the primitive-body direct-eval path.
- Explored representing the resolved function target directly in the lowered IR / atom term layer. Parked that as a larger semantic cleanup so this PR stays isolated and minimal for the experimental primitive-body API.
- Does not include the egglog-experimental `(primitive ...)` command implementation and does not include the unrelated core-relations ordering fix from PR #899.

## Validation

- Latest after merging `upstream/main`:
  - `cargo test --lib typecheck_expr_with --quiet`
  - `cargo test --test typed_primitive --quiet`
  - `cargo check --lib`
  - `git diff --check`
- Earlier on this PR:
  - `make nits`
  - `cargo check --locked --lib`
  - `cargo test --lib --quiet`
  - `cargo test --test naive_unstable_fn_via_global --quiet`
  - `cargo test --test files --features bin -- unstable_fn --quiet`
  - `cargo test --test files --features bin -- typed_primitive_unstable_app --quiet`
  - Separate review agent over this PR plus the dependent experimental PR: no remaining findings after the dependency pin was updated.

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -65,16 +65,12 @@ impl SpecializedPrimitive {
 
     /// Get the external function ID of this primitive
     pub(crate) fn external_id(&self, ctx: crate::Context) -> ExternalFunctionId {
-        self.external_id_in(ctx).unwrap_or_else(|| {
+        self.prim_with_id.context_ids[ctx].unwrap_or_else(|| {
             panic!(
                 "primitive {:?} is not valid in context {ctx:?}",
                 self.prim_with_id.primitive.name()
             )
         })
-    }
-
-    pub(crate) fn external_id_in(&self, ctx: crate::Context) -> Option<ExternalFunctionId> {
-        self.prim_with_id.context_ids[ctx]
     }
 
     /// Get the validator function of this primitive, if any
--- a/src/core.rs
+++ b/src/core.rs
@@ -53,6 +53,11 @@ impl SpecializedPrimitive {
         self.prim_with_id.primitive.name()
     }
 
+    /// Apply this primitive to already-evaluated arguments.
+    pub fn apply(&self, exec_state: &mut ExecutionState<'_>, args: &[Value]) -> Option<Value> {
+        exec_state.call_external_func(self.external_id(crate::Context::Full), args)
+    }
+
     /// Get the output sort of this primitive
     pub fn output(&self) -> &ArcSort {
         &self.output
--- a/src/core.rs
+++ b/src/core.rs
@@ -53,11 +53,6 @@ impl SpecializedPrimitive {
         self.prim_with_id.primitive.name()
     }
 
-    /// Apply this primitive to already-evaluated arguments.
-    pub fn apply(&self, exec_state: &mut ExecutionState<'_>, args: &[Value]) -> Option<Value> {
-        exec_state.call_external_func(self.external_id(crate::Context::Full), args)
-    }
-
     /// Get the output sort of this primitive
     pub fn output(&self) -> &ArcSort {
         &self.output
--- a/src/core.rs
+++ b/src/core.rs
@@ -65,12 +65,16 @@ impl SpecializedPrimitive {
 
     /// Get the external function ID of this primitive
     pub(crate) fn external_id(&self, ctx: crate::Context) -> ExternalFunctionId {
-        self.prim_with_id.context_ids[ctx].unwrap_or_else(|| {
+        self.external_id_in(ctx).unwrap_or_else(|| {
             panic!(
                 "primitive {:?} is not valid in context {ctx:?}",
                 self.prim_with_id.primitive.name()
             )
         })
+    }
+
+    pub(crate) fn external_id_in(&self, ctx: crate::Context) -> Option<ExternalFunctionId> {
+        self.prim_with_id.context_ids[ctx]
     }
 
     /// Get the validator function of this primitive, if any
--- a/src/exec_state.rs
+++ b/src/exec_state.rs
@@ -30,11 +30,17 @@
 //! [`ReadPrim`]: crate::ReadPrim
 //! [`FullPrim`]: crate::FullPrim
 
-use std::ops::Deref;
+use std::{fmt, ops::Deref};
 
 use crate::core_relations::{
     BaseValue, BaseValues, ContainerValue, ContainerValues, ExecutionState, ExternalFunctionId,
     TableId, Value,
+};
+use crate::{
+    ast::{FunctionSubtype, Literal, ResolvedExpr},
+    core::ResolvedCall,
+    sort::{F, S},
+    typechecking::FuncType,
 };
 use egglog_bridge::{ActionRegistry, TableAction};
 
@@ -100,6 +106,26 @@ pub(crate) trait Internal<'a, 'db: 'a>: 'a {
     }
     fn raw_exec_state(&mut self) -> &mut ExecutionState<'db> {
         self.es_mut()
+    }
+    fn apply_resolved_function(&mut self, _func: &FuncType, _args: &[Value]) -> Option<Value> {
+        None
+    }
+
+    fn apply_table_function(
+        &mut self,
+        subtype: FunctionSubtype,
+        action: &TableAction,
+        args: &[Value],
+    ) -> Option<Value> {
+        match (subtype, self.ctx()) {
+            (FunctionSubtype::Constructor, Context::Write | Context::Full) => {
+                action.lookup_or_insert(self.es_mut(), args)
+            }
+            (FunctionSubtype::Custom, Context::Read | Context::Full) => {
+                action.lookup(self.es(), args)
+            }
+            _ => None,
+        }
     }
 }
 
@@ -199,7 +225,9 @@ pub trait Core<'a, 'db: 'a>: Internal<'a, 'db> {
     }
 
     /// Dispatch an already type-specialized primitive in the current
-    /// call-site context.
+    /// call-site context. The primitive should have been resolved for
+    /// this same context, for example by typechecking the expression
+    /// under the context where it will run.
     fn apply_primitive(
         &mut self,
         primitive: &crate::core::SpecializedPrimitive,
@@ -207,6 +235,72 @@ pub trait Core<'a, 'db: 'a>: Internal<'a, 'db> {
     ) -> Option<Value> {
         let id = primitive.external_id(self.ctx());
         self.es_mut().call_external_func(id, args)
+    }
+
+    /// Evaluate an already typechecked expression in this primitive call
+    /// context, using `bindings` for local variables.
+    ///
+    /// Primitive calls dispatch through the current call-site context.
+    /// Table-backed function calls follow the same capability split as
+    /// `unstable-app`: custom functions require a read-capable state,
+    /// and constructors require a write-capable state that can mint on miss.
+    fn eval_resolved_expr(
+        &mut self,
+        expr: &ResolvedExpr,
+        bindings: &[(&str, Value)],
+    ) -> Option<Value> {
+        eval_resolved_expr_result(self, expr, bindings).ok()
+    }
+}
+
+#[derive(Debug)]
+pub(crate) enum EvalError {
+    UnboundVariable(String),
+    FunctionFailed(String),
+    PrimitiveFailed(String),
+}
+
+impl fmt::Display for EvalError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            EvalError::UnboundVariable(name) => write!(f, "unbound variable {name}"),
+            EvalError::FunctionFailed(name) => write!(f, "lookup of function {name} failed"),
+            EvalError::PrimitiveFailed(name) => write!(f, "call of primitive {name} failed"),
+        }
+    }
+}
+
+pub(crate) fn eval_resolved_expr_result<'a, 'db: 'a>(
+    state: &mut (impl Core<'a, 'db> + ?Sized),
+    expr: &ResolvedExpr,
+    bindings: &[(&str, Value)],
+) -> Result<Value, EvalError> {
+    match expr {
+        ResolvedExpr::Lit(_, literal) => Ok(match literal {
+            Literal::Int(x) => state.base_to_value(*x),
+            Literal::Float(x) => state.base_to_value(F::from(*x)),
+            Literal::String(x) => state.base_to_value(S::new(x.clone())),
+            Literal::Bool(x) => state.base_to_value(*x),
+            Literal::Unit => state.base_to_value(()),
+        }),
+        ResolvedExpr::Var(_, resolved_var) => bindings
+            .iter()
+            .find_map(|(name, value)| (*name == resolved_var.name).then_some(*value))
+            .ok_or_else(|| EvalError::UnboundVariable(resolved_var.name.clone())),
+        ResolvedExpr::Call(_, resolved_call, children) => {
+            let mut values = Vec::with_capacity(children.len());
+            for child in children {
+                values.push(eval_resolved_expr_result(state, child, bindings)?);
+            }
+            match resolved_call {
+                ResolvedCall::Primitive(primitive) => state
+                    .apply_primitive(primitive, &values)
+                    .ok_or_else(|| EvalError::PrimitiveFailed(primitive.name().to_owned())),
+                ResolvedCall::Func(func) => state
+                    .apply_resolved_function(func, &values)
+                    .ok_or_else(|| EvalError::FunctionFailed(func.name.clone())),
+            }
+        }
     }
 }
 
@@ -279,6 +373,15 @@ fn lookup_action<'r>(registry: &'r ActionRegistry, name: &str) -> &'r TableActio
     registry
         .lookup_table(name)
         .unwrap_or_else(|| panic!("missing table action for table: {name}"))
+}
+
+fn apply_registered_function<'a, 'db: 'a>(
+    state: &mut impl RegistrySealed<'a, 'db>,
+    func: &FuncType,
+    args: &[Value],
+) -> Option<Value> {
+    let action = lookup_action(state.registry(), &func.name).clone();
+    state.apply_table_function(func.subtype, &action, args)
 }
 
 // =====================================================================
@@ -433,6 +536,9 @@ impl<'a, 'db: 'a> Internal<'a, 'db> for ReadState<'a, 'db> {
     fn ctx(&self) -> Context {
         self.ctx
     }
+    fn apply_resolved_function(&mut self, func: &FuncType, args: &[Value]) -> Option<Value> {
+        apply_registered_function(self, func, args)
+    }
 }
 impl<'a, 'db: 'a> RegistrySealed<'a, 'db> for ReadState<'a, 'db> {
     fn registry(&self) -> &ActionRegistry {
@@ -452,6 +558,9 @@ impl<'a, 'db: 'a> Internal<'a, 'db> for WriteState<'a, 'db> {
     fn ctx(&self) -> Context {
         self.ctx
     }
+    fn apply_resolved_function(&mut self, func: &FuncType, args: &[Value]) -> Option<Value> {
+        apply_registered_function(self, func, args)
+    }
 }
 impl<'a, 'db: 'a> RegistrySealed<'a, 'db> for WriteState<'a, 'db> {
     fn registry(&self) -> &ActionRegistry {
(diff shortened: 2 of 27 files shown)


src/lib.rs
```diff
@@ -1103,6 +1103,48 @@ impl EGraph {
...
+    ///
+    /// `output_sort` constrains overload resolution and output-type inference
+    /// for `expr`.
+    pub fn typecheck_expr_with_output(
+        &mut self,
+        expr: &Expr,
+        output_sort: ArcSort,
+    ) -> Result<ResolvedExpr, TypeError> {
+        self.typecheck_expr_with_bindings_and_output(expr, &[], output_sort)
+    }
```
saulshanabrook: Can you remove this top level command? We dont need it in the downstream API and the lower one covers this case.

src/core.rs
```diff
@@ -53,6 +53,11 @@ impl SpecializedPrimitive {
         self.prim_with_id.primitive.name()
     }
 
+    /// Apply this primitive to already-evaluated arguments.
+    pub fn apply(&self, exec_state: &mut ExecutionState<'_>, args: &[Value]) -> Option<Value> {
+        exec_state.call_external_func(self.external_id(crate::Context::Full), args)
+    }
+
```
saulshanabrook: Do we need this function? I dont see it used in this PR or in the egglog experimental PR that builds on it?

src/typechecking.rs
```diff
@@ -979,6 +981,43 @@ impl TypeInfo {
         }
     }
 
+    pub(crate) fn typecheck_expr_with_output(
+        &self,
+        symbol_gen: &mut SymbolGen,
+        expr: &Expr,
+        binding: &IndexMap<&str, (Span, ArcSort)>,
+        output_sort: ArcSort,
```
saulshanabrook, line 1007: Could this be used anywhere else in the codebase besides in exposing this at the top level to to the e-graph? Like aren't there other places where we typecheck expressions with bindings? Or is it not worth it to generalize between them? Can you give an overview of where else this logic lives?

saulshanabrook:
@coderabbitai full review

src/lib.rs
```diff
@@ -2038,82 +2233,23 @@ impl<'a> BackendRule<'a> {
...
+            let resolved_function = resolved_function_for_unstable_fn(
+                self.rb.egraph(),
+                self.functions,
+                self.type_info,
+                prim,
+                name,
                 panic_id,
-            });
+            )
+            .unwrap_or_else(|err| panic!("{err}"));
```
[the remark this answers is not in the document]
saulshanabrook: Deferring for now, added to https://github.com/egraphs-good/egglog/issues/898, to not change existing signatures

src/exec_state.rs
```diff
@@ -197,6 +223,85 @@ pub trait Core<'a, 'db: 'a>: Internal<'a, 'db> {
...
+            Literal::Int(x) => state.base_to_value(*x),
+            Literal::Float(x) => state.base_to_value(F::from(*x)),
+            Literal::String(x) => state.base_to_value(S::new(x.clone())),
+            Literal::Bool(x) => state.base_to_value(*x),
+            Literal::Unit => state.base_to_value(()),
+        }),
+        ResolvedExpr::Var(_, resolved_var) => bindings
+            .iter()
+            .find_map(|(name, value)| (*name == resolved_var.name).then_some(*value))
+            .ok_or_else(|| EvalError::UnboundVariable(resolved_var.name.clone())),
```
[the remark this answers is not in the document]
saulshanabrook: Just raising an error for now on this.

src/exec_state.rs
```diff
@@ -101,6 +107,26 @@ pub(crate) trait Internal<'a, 'db: 'a>: 'a {
...
+        match (subtype, self.ctx()) {
+            (FunctionSubtype::Constructor, Context::Write | Context::Full) => {
+                action.lookup_or_insert(self.es_mut(), args)
+            }
+            (FunctionSubtype::Custom, Context::Read | Context::Full) => {
+                action.lookup(self.es(), args)
+            }
+            _ => None,
+        }
+    }
```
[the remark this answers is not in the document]
saulshanabrook: Adding documentation to state this isn't allowed. This can never happen from the experimental call site.

src/lib.rs
```diff
@@ -1125,6 +1125,139 @@ impl EGraph {
...
+        }
+
+        let resolved = self.type_info.typecheck_expr_with_output(
+            &mut self.parser.symbol_gen,
+            expr,
+            &binding_map,
+            output_sort,
+            context,
+        )?;
+        if let Some((name, span)) = global_ref(&resolved) {
```
yihozhang: Why are globals bad? Is it also bad for other usages of this function?

src/lib.rs
```diff
@@ -1944,6 +2077,84 @@ impl EGraph {
     }
 }
 
+fn resolved_function_for_unstable_fn(
```
yihozhang: Could you document this function?

src/lib.rs
```diff
@@ -1125,6 +1125,139 @@ impl EGraph {
...
+        Ok(resolved)
+    }
+
+    /// Replace literal `(unstable-fn "...")` targets with hidden evaluator bindings.
+    ///
+    /// The returned expression must be evaluated with the returned bindings in
+    /// scope before any caller-supplied local bindings. This lets direct
+    /// execution-state evaluation use the same hidden `ResolvedFunction` value
+    /// that backend rule lowering injects for `unstable-fn`.
+    pub fn prepare_unstable_fn_targets_for_eval(
```
yihozhang, line 1221: Maybe give an example in the doc

src/lib.rs
```diff
@@ -1944,6 +2077,84 @@ impl EGraph {
...
+        let ps: Vec<_> = type_info
+            .get_prims(name)
+            .into_iter()
+            .flatten()
+            .filter(|p| p.accept(&types, type_info))
+            .collect();
+        let mut context_ids = enum_map::EnumMap::from_fn(|_| None);
+        for runtime_ctx in Context::ALL {
+            let mut ids = ps.iter().filter_map(|p| p.context_ids[runtime_ctx]);
+            match (ids.next(), ids.next()) {
```
yihozhang, line 2256: Add a comment explaining ids.next(), id.next()

yihozhang (commented):
Some small questions. Overall looks good

src/lib.rs
saulshanabrook: Thanks for this one! I fixed this, globals now just panic. However, the included typecheck expression will properly transform globals into let bindings like the normal path does, so using these together should support globals now. I also added a test in the primitive in experimental for globals.

src/lib.rs
```diff
@@ -2077,53 +2166,93 @@ impl EGraph {
...
 }
 
-fn resolved_function_for_unstable_fn(
+/// Build the runtime value backing a resolved `(unstable-fn name)` target.
+///
+/// For table-backed functions, this captures the table action that
+/// `unstable-app` will call later. For primitive targets, this bakes one
+/// dispatch id per runtime context so application can choose the entrypoint
+/// matching the primitive body's current call-site context.
+fn resolve_function_container_target_with_context(
```
yihozhang, line 2175: Why did this function get refactored? What changed?

yihozhang (approved):
Had a small question. Otherwise looks good. Thanks!

oflatt:
Hmm, can we chat about this PR before we merge it? I'd like some context

src/lib.rs
```diff
@@ -1170,6 +1170,134 @@ impl EGraph {
...
+    /// The returned expression must be evaluated with the returned bindings in
+    /// scope before any caller-supplied local bindings. This lets direct
+    /// execution-state evaluation use the same hidden `ResolvedFunction` value
+    /// that backend rule lowering injects for `unstable-fn`.
+    ///
+    /// For example, a resolved body like `(unstable-app (unstable-fn "f") _0)`
+    /// cannot evaluate the string literal `"f"` directly. This helper replaces
+    /// the `(unstable-fn "f")` sub-expression with a fresh hidden variable and
+    /// returns a binding from that variable to the prepared function value.
+    pub fn prepare_unstable_fn_targets_for_eval(
```
oflatt, line 1221: Could you explain why we need this? I'm wary of new unstable-fn related code in core

src/lib.rs
```diff
@@ -1170,6 +1170,134 @@ impl EGraph {
...
+    /// output sort, and a primitive call context.
+    ///
+    /// `bindings` contains the local variables in scope while resolving `expr`.
+    /// Each tuple is `(name, span, sort)`, where `span` is used for diagnostics
+    /// tied to that binding. `output_sort` constrains overload resolution and
+    /// output-type inference for the expression. Global references are rewritten
+    /// into the same zero-argument function calls used during command execution.
+    /// `context` should match the runtime context where the resolved expression
+    /// will be evaluated.
+    pub fn typecheck_expr_with_bindings_and_output(
```
oflatt, line 1183: Doesn't this mean if users get the binding types wrong they can get ill-typed programs? Is that a concern? 

I imagine this is for custom commands

src/typechecking.rs
```diff
@@ -992,6 +999,54 @@ impl TypeInfo {
         }
     }
 
+    pub(crate) fn typecheck_expr_with_output(
```
oflatt, line 1002: Did we really not have this? Is this duplicate code with existing typecheck expr?

oflatt (changes requested):
Had some questions

oflatt:
Do we really need new primitive declarations in the body of a rule? Is that the main motivation for this PR?

src/typechecking.rs
yihozhang: It's mostly not. We can extract some common code for setting up type checking with an optional output type, but this isn't much code anyway.

saulshanabrook:
@oflatt Thanks for the comments!

This is for https://github.com/egraphs-good/egglog-experimental/pull/46 which I needed for the A/C stuff with maps, so you can do:

```
(primitive my-prim (i64 i64) i64 (* (+ _0 _1) 3))
```

> Do we really need new primitive declarations in the body of a rule? Is that the main motivation for this PR?

No we dont have new primitive declarations in the body of a rule, only at the top level!

We just have basically like "eval a resolved expression" in the body of a rule, since this `primitive` top level command is turned into resolving a primitive with the resolved body saved.

src/typechecking.rs
saulshanabrook: Yeah I did try to see if we could extract out some commonality, but I'll have another go at it, to see if it decreases repetition or just adds more issues.

src/lib.rs
saulshanabrook: If they put in the binding types wrong, it won't typecheck I believe? Is that what you mean?

src/lib.rs
saulshanabrook: This is needed to to do the same translation as in our pass that tests `if fn_name == "unstable-fn"` or whatever, that replaces the first arg of the function name as a string with the real resolved function value. So it just exposes the current behavior we have in core so others can you use it when preparing a resolved expression. Like this is how its used in the `primitive` command in experimental:

```rust
            match egraph.typecheck_expr_with_bindings_and_output(
                &args[3],
                &bindings,
                output_sort.clone(),
                context,
            ) {
                Ok(resolved)
                    if matches!(
                        (required_context(egraph, &resolved), context),
                        (Context::Pure, _)
                            | (Context::Read, Context::Read | Context::Full)
                            | (Context::Write, Context::Write | Context::Full)
                            | (Context::Full, Context::Full)
                    ) =>
                {
                    typechecked_body = Some((resolved, context));
                    break;
                }
                Ok(_) => {}
                Err(err) => last_error = Some(err),
            }
        }
        let Some((body, context)) = typechecked_body else {
            return Err(last_error
                .expect("primitive body typechecking always tries at least one context")
                .into());
        };

        let (body, hidden_bindings) = egraph.prepare_unstable_fn_targets_for_eval(&body)?;
        let primitive = DefinedPrimitive {
            name,
            input_vars: input_var_names,
            input: input_sorts,
            output: output_sort,
            body,
            hidden_bindings,
        };
```



I think to make this DRY and do the same code path on both would require more refactoring and make unstable function more in core.

I did have a go at this, but threw out that branch when it added more code, also after talking to @yihozhang about it. The "best" place to put this for overall design is in ResolvedExpr to add a specific case for a Function, but yeah this adds it even more inside core, so decided against that for now...

src/typechecking.rs
saulshanabrook: I was able to refactor it so there are a few shared helpers here, do you guys like this more? A little more LoC but less duplication:


```diff
diff --git a/src/typechecking.rs b/src/typechecking.rs
index dedb17f6..c3623d0f 100644
--- a/src/typechecking.rs
+++ b/src/typechecking.rs
@@ -6,7 +6,8 @@ use crate::{
     *,
 };
 use ast::{
-    MappedExprExt, ResolvedAction, ResolvedExpr, ResolvedFact, ResolvedRule, ResolvedVar, Rule,
+    MappedActions, MappedExprExt, ResolvedAction, ResolvedExpr, ResolvedFact, ResolvedRule,
+    ResolvedVar, Rule,
 };
 use core_relations::ExternalFunction;
 use egglog_ast::generic_ast::GenericAction;
@@ -960,6 +961,18 @@ impl TypeInfo {
         binding: &IndexMap<&str, (Span, ArcSort)>,
         context: Context,
     ) -> Result<ResolvedActions, TypeError> {
+        let (problem, mapped_action) =
+            self.build_standalone_action_problem(symbol_gen, actions, binding, context)?;
+        self.solve_standalone_action_problem(problem, mapped_action, context)
+    }
+
+    fn build_standalone_action_problem(
+        &self,
+        symbol_gen: &mut SymbolGen,
+        actions: &Actions,
+        binding: &IndexMap<&str, (Span, ArcSort)>,
+        context: Context,
+    ) -> Result<(Problem<AtomTerm, ArcSort>, MappedActions<String, String>), TypeError> {
         let mut binding_set: IndexSet<String> =
             binding.keys().copied().map(str::to_string).collect();
         // We lower to core actions with `union_to_set_optimization`
@@ -975,6 +988,15 @@ impl TypeInfo {
             problem.assign_local_var_type(var, span.clone(), sort.clone())?;
         }
 
+        Ok((problem, mapped_action))
+    }
+
+    fn solve_standalone_action_problem(
+        &self,
+        problem: Problem<AtomTerm, ArcSort>,
+        mapped_action: MappedActions<String, String>,
+        context: Context,
+    ) -> Result<ResolvedActions, TypeError> {
         let assignment = problem
             .solve(|sort: &ArcSort| sort.name())
             .map_err(|e| e.to_type_error())?;
@@ -1008,29 +1030,20 @@ impl TypeInfo {
         context: Context,
     ) -> Result<ResolvedExpr, TypeError> {
         let action = Action::Expr(expr.span(), expr.clone());
-        let mut binding_set: IndexSet<String> =
-            binding.keys().copied().map(str::to_string).collect();
-        let mut ctx = CoreActionContext::new(self, &mut binding_set, symbol_gen, false);
-        let (actions, mapped_action) = Actions::singleton(action).to_core_actions(&mut ctx)?;
-        let mut problem = Problem::default();
-
-        problem.add_actions(&actions, self, symbol_gen, context)?;
-
-        for (var, (span, sort)) in binding {
-            problem.assign_local_var_type(var, span.clone(), sort.clone())?;
-        }
-
+        let (mut problem, mapped_action) = self.build_standalone_action_problem(
+            symbol_gen,
+            &Actions::singleton(action),
+            binding,
+            context,
+        )?;
         let [GenericAction::Expr(_, mapped_expr)] = mapped_action.0.as_slice() else {
             unreachable!("typechecking an expression should produce one expression action")
         };
         let output_atom = mapped_expr.get_corresponding_var_or_lit(self);
         problem.add_binding(output_atom, output_sort.clone());
 
-        let assignment = problem
-            .solve(|sort: &ArcSort| sort.name())
-            .map_err(|e| e.to_type_error())?;
-
-        let annotated_actions = assignment.annotate_actions(&mapped_action, self, context)?;
+        let annotated_actions =
+            self.solve_standalone_action_problem(problem, mapped_action, context)?;
         match annotated_actions.0.into_iter().next().unwrap() {
             ResolvedAction::Expr(_, resolved_expr) => {
                 let actual = resolved_expr.output_type();
```

src/lib.rs
saulshanabrook: I believe from the merge from main, with the context stuff and also with supporting primtives doing the more advanced typea analysis. LMK if you have issues with how it is now.

saulshanabrook:
I talked to Oliver and he seems to generally feel OK about most of this PR since its pretty general, except for `prepare_unstable_fn_targets_for_eval`, because it leaks out details of how we are compiling unstable functions currently.

I tried to see if we could remove that API and instead just do it in experimental, however, we would have to expose the existing `resolve_function_container_target_with_context` function so that experimental could use it and this would be hard because it uses a bunch of things that are private.

We could make a wrapper around that to expose, but I am not sure if that resolves the objection.

Details from the agent:

> 
> Making the existing function pub is not sufficient. Experimental could not call it without exposing more internals.
> 
> Current parameters require:
> 
> - `&egglog_bridge::EGraph`: `EGraph.backend` is private.
> - `&IndexMap<String, Function>`: `EGraph.functions` is private, and `Function::backend_id` is private.
> - `ExternalFunctionId`: there is no public EGraph API to register the required panic handler.
> - Primitive runtime IDs: the resolver reads context-specific IDs that are pub(crate).
> - `ResolvedFunctionId` contains backend TableAction values, coupling experimental directly to backend internals.
> 
> `ResolvedFunction`, `SpecializedPrimitive`, `TypeInfo`, and `EGraph::base_to_value` are accessible, but the pieces needed to construct the value are not.
> 
> You could instead expose a wrapper:
> 
> ```rust
> pub fn resolve_function_target_for_eval(
>     &mut self,
>     name: &str,
>     primitive: &SpecializedPrimitive,
> ) -> Result<Value, Error>
> ```
> 
> Because it takes `&mut EGraph`, its implementation could access all private fields internally. Experimental could then own the tree traversal and hidden bindings.
> 
> However, that does not really address Oliver’s objection. It replaces one high-level unstable-function API with a lower-level, more explicitly unstable-function API. The existing preparation method is arguably
> less leaky because it does not expose ResolvedFunction construction details.
> 
> My recommendation:
> 
> 1. If unstable-fn support in body-defined primitives is optional, remove it from this PR.
> 2. Remove `prepare_unstable_fn_targets_for_eval`.
> 3. Reject `unstable-fn` in experimental primitive bodies with a focused fail-typecheck test.
> 4. Leave proper function-value support for a later IR/prepared-expression refactor.
> 
> If support must remain, the current preparation method is probably the least damaging public boundary. Rust has no friend visibility between egglog and egglog-experimental, even in the same workspace.


Instead to do this "right" would require refactoring the compiler toolchain to handle unstable function more consistently and as first class. This would be a larger change and would be happy to look into this, but not sure if its right for this PR.

oflatt:
Thanks for exploring. I'm going to approve to unblock this, and we can discuss next steps in the next egglog meeting.

oflatt (approved):

PR #885, opened 2026-05-18
Fix scheduler fresh matches after subsumption

saulshanabrook:
## Context

Custom schedulers collect fresh candidate matches before deciding which scheduler rows should run. Fresh collection was still allowed to use subsumed body witnesses, so a scheduled rule could run actions for a row that analysis had already subsumed before the scheduler step.

## Concrete fix

- Collect fresh scheduler matches from non-subsumed rows by querying the rule body with subsumed rows excluded.
- Add a regression where analysis subsumes `(Add (Mul (Num 0) (Num 1)) (Num 2))` before the scheduled ruleset runs. The scheduled rule would match the old affine shape, but the scheduler now sees no fresh valid match, inserts no `Hit` row, and reports no user-visible update.

## Scope

This narrowed PR only fixes fresh scheduler match collection after subsumption. It does not add held-backlog validation or cleanup for matches that were already saved by a persistent scheduler in an earlier iteration.

## Validation

- `cargo test --lib scheduler`
- `cargo fmt --check`
- `make nits` (passed; emitted existing rustdoc warnings)
- `git diff --check`


saulshanabrook:
@coderabbitai full review

src/scheduler.rs
```diff
@@ -364,7 +421,31 @@ impl SchedulerRuleInfo {
...
+            .map(|fv| validation_rule_builder.entry(&GenericAtomTerm::Var(span!(), fv.clone())))
+            .collect::<Vec<_>>();
+        entries.push(unit_entry.clone());
+        validation_rule_builder
+            .rb
+            .query_table(decided, &entries, None)
+            .unwrap();
+        validation_rule_builder.query(&rule.body, false);
+        validation_rule_builder.rb.set(validated, &entries);
+        let validation_rule = validation_rule_builder.build();
```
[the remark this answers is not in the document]
saulshanabrook: Fixed

saulshanabrook:
@coderabbitai full review

src/scheduler.rs
```diff
@@ -281,17 +328,33 @@ pub(crate) struct SchedulerRecord {
     rule_info: HashMap<String, SchedulerRuleInfo>,
 }
 
-/// To enable scheduling without modifying the backend,
-/// we split a rule (rule query action) into a worklist relation
-/// two rules (rule query (worklist vars false)) and
-/// (rule (worklist vars false) (action ... (delete (worklist vars false))))
+/// To enable scheduling without modifying the backend, we split a rule into
+/// scheduler-internal worklist and validation relations plus four rules:
+/// (rule query (worklist vars false)),
```
yihozhang: I realized the constructed rules actually don't have the `false` entry.

src/scheduler.rs
```diff
@@ -281,17 +328,33 @@ pub(crate) struct SchedulerRecord {
...
+/// (rule (worklist vars false) query (validated vars false)),
```
yihozhang: I'm worried that this will be slow because we now run the query twice, and because the conjunction of `(worklost ...)` and `query` may yield a different plan, performance becomes unpredictable.

yihozhang:
What motivates this PR? At first glance, I am more or less okay with the current semantics in the presence of non-monotonic ops like subsume or delete...

src/scheduler.rs
saulshanabrook: Fixed this comment thanks!

src/scheduler.rs
saulshanabrook: Yeah that makes sense. I wonder if there is a way to fix this bug while still not degrading performance... if this does in fact lead to slowdowns. 

Some ideas from LLM but wanna look deeper

  The key invariant we need is narrower:

  A held scheduler key must not cross a subsume / delete boundary and later run actions without either being revalidated or discarded.

 

>  So the options are:
> 
>   1. Keep current PR: always validate chosen keys
>       - Correct and simple.
>       - Preserves held valid work if another witness still exists.
>       - Performance concern remains: every chosen key path has the second body query.
> 
>   2. Drop held backlog after non-monotonic changes
>       - Track a global “invalidation epoch” bumped by subsume / delete.
>       - Store the epoch with residual scheduler keys.
>       - If a held key is from an older epoch, discard it instead of validating.
>       - Passes your repro: analysis subsumes, epoch changes, second copy drops held keys, Hit stays 0.
>       - Fastest and simplest alternative, but conservative: it can drop still-valid held matches.
> 
>   3. Epoch-gated validation
>       - Same epoch tracking, but only old held keys go through validation.
>       - Fresh keys, or held keys from the current epoch, run through the cheap direct action path.
>       - In your repro, validation runs only because analysis happened between scheduler runs.
>       - This is probably the best semantic/performance compromise.
> 
>   4. Per-relation invalidation epochs
>       - Like option 3, but only validate/drop if a relation used by that rule body had non-monotonic changes.
>       - More precise and faster for unrelated analysis, but more plumbing: rule dependency tracking plus table-level invalidation.
> 
>   5. Experimental-only workaround
>       - Change back-off in egglog-experimental to clear held matches after analysis/non-monotonic phases.
>       - Smallest if the only goal is Python param_eq.
>       - But it leaves core scheduler semantics footgunny, and other persistent schedulers can still hit the stale held-key bug.

saulshanabrook:
> What motivates this PR? At first glance, I am more or less okay with the current semantics in the presence of non-monotonic ops like subsume or delete...


Basically that a backoff scheduler can hold onto a deleted/subsumed value... So then some analysis runs, it gets subsumed, then it runs again and still triggers. This to me breaks my mental semantic model and breaks a lot of reasoning about subsumption. I would assume that after something has been subsumed in one round, nothing in later rounds can match it. 

In my Python work it depends on this behavior... Like ones round, runs analysis to saturation (which subsumes for constant folding like stuff among other things), then runs another round. If the next round picks up on things that were subsumed, certain assertions break in my domain model... I can look for a bigger example, but here is a small reproduction:


> 
> # Held Scheduler Match Repro
> 
> The smallest experimental test that needs held-key validation is:
> 
> ```rust
> #[test]
> fn test_backoff_drops_held_matches_subsumed_by_analysis() {
>     let mut egraph = egglog_experimental::new_experimental_egraph();
> 
>     egraph.parse_and_run_program(None, r#"
>         (ruleset analysis)
>         (ruleset copy)
>         (datatype Math
>           (Add Math Math)
>           (Num i64))
>         (relation Hit (Math))
> 
>         (let expr1 (Add (Num 0) (Num 1)))
>         (let expr2 (Add (Num 0) (Num 2)))
> 
>         (rewrite (Add (Num 0) x) x :subsume :ruleset analysis)
>         (rule ((= e (Add (Num 0) x))) ((Hit e)) :ruleset copy :name "copy-add")
> 
>         (let-scheduler bo (back-off :match-limit 1 :ban-length 1))
>     "#).unwrap();
> 
>     egraph.parse_and_run_program(None, "(run-schedule (run-with bo copy))").unwrap();
>     assert_eq!(egraph.get_size("Hit"), 0);
> 
>     egraph.parse_and_run_program(None, "(run-schedule (run analysis))").unwrap();
> 
>     egraph.parse_and_run_program(None, "(run-schedule (run-with bo copy))").unwrap();
>     assert_eq!(egraph.get_size("Hit"), 0);
> }
> ```
> 
> I verified it against experimental `origin/main` with the current pinned egglog, without held-key validation. It fails:
> 
> ```text
> assertion `left == right` failed: held scheduler keys whose body witnesses were subsumed must not fire
> left: 2
> right: 0
> ```
> 
> Mechanism:
> 
> 1. First `(run-with bo copy)` sees two matches.
> 2. `back-off :match-limit 1` refuses to run them and keeps them in the scheduler backlog.
> 3. `(run analysis)` subsumes both `(Add (Num 0) ...)` witnesses.
> 4. Second `(run-with bo copy)` reuses the held backlog.
> 5. Without held-key validation, core runs actions from the held scheduler key without rechecking the rule body, so it inserts two stale `Hit` rows.
> 6. With held-key validation, validation rechecks `decided + rule.body` against non-subsumed rows, finds no valid witness, and cleanup drops the stale keys.
> 
> This maps to Python `param_eq`: each loop does scheduled rewrite work, then the next loop runs analysis before scheduled rewrite again. If `back-off` held rewrite matches, analysis can subsume their witnesses before the scheduler tries them later.
>

saulshanabrook:
@yihozhang this is ready for another review. I tightened it up to only handle on thing, correctness of subsumption. I also improved performance, so it only re-checks held over rows to see if they had been subsumed or deleted.

saulshanabrook:
OK from feedback from @yihozhang and @oflatt that its unclear if this was the intended semantics and also based on performance concerns, I have stripped this PR down to just fixing an even more basic bug, that previously it would find subsumed rows even before they were deferred. I believe this fixes that bug, but let me know if somehow this was intentional for some reason.

I tested my upstream code against this and it works fine. 

Sorry for all the time spent reviewing these more complicated fixes, when the simpler one was enough at least for me!

yihozhang (approved):
Thank you!

PR #888, opened 2026-05-18
Propagate dirty ids through nested containers

saulshanabrook:
## Context

If an inner container rebuilds in place, parent containers that contain it can become newly matchable without changing their own ids. Seminaive matching needs those parent ids marked dirty too. Follow-up to https://github.com/egraphs-good/egglog/pull/832.

## Changes

- Expands container rebuild dirty ids through parent containers to a fixed point.
- Adds a focused nested-container regression in `tests/nested-container-dirty-propagation.egg`.
- Leaves `tests/container-fail.egg` focused on the original #831 regression.
- Updates the relevant snapshots, including the proof-support inventory.

## Validation

- `cargo test --test files nested_container_dirty_propagation -- --test-threads=1`
- `cargo test --test files container_fail -- --test-threads=1`
- `cargo test --test files proof_support_snapshot -- --test-threads=1`
- `cargo fmt --check`
- `git diff --check`
- CodeRabbit review: 0 issues after addressing the initial clone-cleanup suggestion.


ezrosent (approved):
* Why are you running tests just with one thread?
* This is a fun consequence of being able to 'directly nest' containers!

saulshanabrook:
> Why are you running tests just with one thread?


Thats a great question... I have no idea 🫣

saulshanabrook:
> This is a fun consequence of being able to 'directly nest' containers!

Yep I guess this is why it was disabled before lol

PR #889, opened 2026-05-18
Allow unstable-fn to target primitives

saulshanabrook:
## Context

`unstable-fn` function containers could target table functions, but primitive overloads needed matching typechecking and runtime resolution support.

This remains separate from the primitive-body API work in https://github.com/egraphs-good/egglog/pull/881. Keeping the function-container changes isolated makes the dependent experimental PR smaller and avoids coupling `(primitive ...)` support to the unstable-fn primitive-target behavior.

## Changes

- Extends `unstable-fn` type constraints to account for primitive overloads.
- Resolves primitive function containers for scheduled rules and merge functions.
- Keeps resolver APIs private to this implementation.
- Adds tests for primitive-backed function containers and merge-function use.

## Validation

- `cargo fmt --check`
- `cargo test -q --test typed_primitive unstable_app`
- `cargo test -q --test typed_primitive unstable_fn_duplicate_primitive_registration_panics_on_build`
- `cargo test -q --test files --features bin unstable_fn`
- `git diff --check`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -733,7 +733,7 @@ impl EGraph {
                             .read()
                             .unwrap()
                             .default_panic_id(),
-                    )?;
+                    );
                     translated_args[0] =
                         egglog_bridge::MergeFn::Const(self.backend.base_values().get(resolved));
                 }
@@ -1972,16 +1972,16 @@ fn resolve_function_container_target_with_context(
     name: &str,
     primitive: &core::SpecializedPrimitive,
     panic_id: ExternalFunctionId,
-) -> Result<ResolvedFunction, Error> {
+) -> ResolvedFunction {
     let Some(target_function) = type_info
         .get_sorts::<FunctionSort>()
         .into_iter()
         .find(|function| function.name() == primitive.output().name())
     else {
-        return Err(Error::BackendError(format!(
+        panic!(
             "`unstable-fn` output sort `{}` is not a function sort",
             primitive.output().name()
-        )));
+        );
     };
 
     let partial_arcsorts: Vec<_> = primitive.input().iter().skip(1).cloned().collect();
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -2367,6 +2469,35 @@ mod tests {
             .unwrap();
     }
 
+    #[test]
+    fn test_unstable_function_container_primitives() {
```
saulshanabrook: Could you just move this into the unstable-fn egg test file?

src/lib.rs
```diff
@@ -2367,6 +2469,35 @@ mod tests {
...
+        let mut egraph = EGraph::default();
+        egraph
+            .parse_and_run_program(
```
saulshanabrook: Can you move this to the unstable fn egg file?

src/lib.rs
```diff
@@ -2367,6 +2469,35 @@ mod tests {
...
+                None,
+                r#"
+                (sort IntVec (Vec i64))
+                (sort IntVecThunk (UnstableFn () IntVec))
+                (sort BoolToBool (UnstableFn (bool) bool))
+                (sort IntToUnit (UnstableFn (i64) Unit))
+                (sort F64ToF64 (UnstableFn (f64) f64))
+                (sort I64ToI64 (UnstableFn (i64) i64))
+
+                (function merge-plus () i64 :merge (unstable-app (unstable-fn "+" old) new))
```
saulshanabrook: Can you add fail typecheck test for something like where `+` is not partially applied with old so it cant resolve?

src/sort/fn.rs
```diff
@@ -272,61 +272,119 @@ impl TypeConstraint for FunctionCTorTypeConstraint {
...
+                        .chain(primitive_constraints.pop().unwrap())
+                        .collect(),
+                    _ => vec![
+                        output_sort_constraint,
+                        constraint::xor(
+                            primitive_constraints
+                                .into_iter()
+                                .map(constraint::and)
+                                .collect(),
+                        ),
```
saulshanabrook, line 402: Can you add comments for this to say what cases this covers with an example and the different matches here and what they correspond to?

src/sort/fn.rs
```diff
@@ -272,61 +272,136 @@ impl TypeConstraint for FunctionCTorTypeConstraint {
...
+                // For example, `(unstable-fn "+" old)` as `UnstableFn (i64) i64`
+                // checks each `+` overload as though it were called with
+                // `(old, future_arg) -> future_output`. The i64 overload matches;
+                // f64/string/etc. overloads become impossible constraints. If
+                // `old` is omitted, the same sort only provides one future arg,
+                // so no binary `+` overload has enough arguments to match.
+                let mut primitive_constraints = Vec::with_capacity(primitives.len());
+                for primitive in primitives {
+                    let mut primitive_args = arguments[1..arguments.len() - 1].to_vec();
+                    let mut constraints = Vec::new();
```
saulshanabrook: Do I need this constraints vec here? Can I Just append to primitive_constraints?

src/lib.rs
```diff
@@ -712,10 +712,31 @@ impl EGraph {
...
             GenericExpr::Call(_, ResolvedCall::Primitive(p), args) => {
-                let translated_args = args
+                let mut translated_args = args
                     .iter()
                     .map(|arg| self.translate_expr_to_mergefn(arg))
                     .collect::<Result<Vec<_>, _>>()?;
+                if p.name() == "unstable-fn" {
+                    let Some(GenericExpr::Lit(_, Literal::String(name))) = args.first() else {
+                        return Err(Error::BackendError(
+                            "expected string literal after `unstable-fn`".into(),
```
saulshanabrook, line 722: Fixes resolving unstable function in merge functions

tests/fail-typecheck/unstable-fn-primitive-missing-partial-arg.egg
```diff
@@ -0,0 +1,7 @@
+;; A primitive target must be partially applied enough to match the declared
+;; function sort. Here `+` is treated as an i64 => i64 function, but no `old`
+;; argument was supplied, so no `+` overload can resolve to that sort.
+
+(sort I64ToI64 (UnstableFn (i64) i64))
+(function merge-plus () i64 :merge (unstable-app (unstable-fn "+") new))
+(set (merge-plus) 1)
```
saulshanabrook: This test doesn't need to be in a merge function right? That isn't important for the failure? Can you also have it call the right number of args for the function but just not infer properly? Like if we define to binary function sorts, one i64 one f64 and that this will fail properly if we don't apply a partial but succeed if we do?

saulshanabrook:
@coderabbitai full review

yihozhang (approved):

PR #894, opened 2026-05-19
[codex] Gate CodSpeed walltime runners

saulshanabrook:
This is so that we don't keep running out of our codspeed budget. We can trigger the walltime runs manually only when we want them on performance critical PRs, instead of for every push to every PR. It will also run them on all commits to main, so that we keep historical performance and can still get a nice diff of performance against main when we do trigger one.

## Summary

- keep the existing CodSpeed simulation benchmark matrix running on GitHub-hosted runners for normal PRs
- add an issue-comment trigger so maintainers can request walltime runs with `@codspeed walltime`
- only include `codspeed-macro` runner matrix rows on pushes to `main` or those explicit walltime comments

## Validation

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/build.yml"); puts "yaml ok"'`
- `git diff --check`

## Notes

The walltime comment trigger is limited to PR comments from `OWNER`, `MEMBER`, or `COLLABORATOR` accounts so arbitrary commenters cannot schedule the CodSpeed Macro runner.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,9 +5,12 @@ on:
     branches:
       - "main"
   pull_request:
+  issue_comment:
+    types: [created]
   workflow_dispatch:
 jobs:
   test:
+    if: github.event_name != 'issue_comment'
     runs-on: ubuntu-latest
     steps:
       - run: echo "CARGO_INCREMENTAL=0" >> "$GITHUB_ENV"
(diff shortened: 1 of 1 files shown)


saulshanabrook:
@coderabbitai full review

yihozhang (approved):
I don't really understand the exact yaml syntax used here, but if you think this would work let's merge it

PR #895, opened 2026-05-22
Expose FullState and run_user_defined_command for EGraph

yihozhang:


src/lib.rs
```diff
@@ -1255,7 +1255,7 @@ impl EGraph {
         }
     }
 
-    fn run_command(&mut self, command: ResolvedNCommand) -> Result<Option<CommandOutput>, Error> {
+    fn run_command(&mut self, command: ResolvedNCommand) -> Result<Vec<CommandOutput>, Error> {
```
saulshanabrook, line 1282: Instead of changing this to a vec, could the RunReport command output that's emitted by experimental be a different struct that just has the outputted commands as a field?

Or are there other reasons you think to make this API change beyond that use case?

src/lib.rs
```diff
@@ -1912,6 +1914,52 @@ impl EGraph {
...
+        let registry_arc = self.backend.action_registry().clone();
+        let registry_guard = registry_arc.read().unwrap();
+        let result = self
+            .backend
+            .with_execution_state(|es| f(FullState::wrap(es, &registry_guard, Context::Full)));
+        drop(registry_guard);
+        self.backend.flush_updates();
+        result
+    }
+
```
saulshanabrook, line 1987: This looks good! @oflatt does this make sense to you?

src/lib.rs
```diff
@@ -1912,6 +1914,52 @@ impl EGraph {
...
+        &mut self,
+        name: &str,
+        args: &[Expr],
+    ) -> Result<Vec<CommandOutput>, Error> {
+        let command = self.commands.swap_remove(name).ok_or_else(|| {
+            Error::TypeError(TypeError::UnboundFunction(name.to_owned(), span!()))
+        })?;
+        let res = command.update(self, args);
+        self.commands.insert(name.to_owned(), command);
+        res
```
saulshanabrook: We cant just clone it and call it? 

Also I guess we won't need this if we figure out a different way in the other PR to not have to duplicate all the command calling stuff.

Seems OK though overall... Although we could do it by just constructing the command and calling it right? Could we do that here instead?

src/lib.rs
oflatt: yep!

src/lib.rs
yihozhang: Both options are fine but I like `Vec` better because the proposed RunReport change seems to open a hole in the backend just for a single `run-schedule` in experimental, and the proposed field will always be an empty Vec for the core crate's usage. It also limits the potential of having other UserDefinedCommand that returns multiple command outputs.

src/lib.rs
yihozhang: Changed to call run_command

src/lib.rs
```diff
@@ -1100,6 +1100,30 @@ impl EGraph {
...
+    ///
+    /// Raises an error if the function does not exist.
+    pub fn clear_function(&mut self, func_name: &str) -> Result<(), Error> {
+        let backend_id = self
+            .functions
+            .get(func_name)
+            .ok_or_else(|| TypeError::UnboundFunction(func_name.to_string(), span!()))?
+            .backend_id;
+        self.backend.clear_table(backend_id);
+        Ok(())
```
saulshanabrook, line 1124: Should we put this on ExecutionState instead?

src/lib.rs
```diff
@@ -1912,6 +1940,51 @@ impl EGraph {
...
+        self.commands.contains_key(name)
+    }
+
+    /// Invoke a registered user-defined command by name, passing the given
+    /// unresolved expression arguments.
+    ///
+    /// This is equivalent to writing `(name args...)` at the top level, but
+    /// callable directly from Rust.  Returns an error if no command with the given
+    /// name is registered.
+    pub fn run_user_defined_command(
```
saulshanabrook, line 1955: Thanks this looks better!

saulshanabrook:
@coderabbitai full review

src/lib.rs
yihozhang: I checked, and this would be hard to do because the implementation clears the index and immediately bumps the table generation, while ExecutionState buffers the updates. If implemented in ExecutionState, it would be the old implementation where we clear tuple-by-tuple.

saulshanabrook (approved):

Issue #872, opened 2026-05-13
Large rule slowdown from recent egglog changes


saulshanabrook:
I am sharing a reproduction from @jafioti of very large slowdowns from when changing from Egglog from december of last year to now https://github.com/egraphs-good/egglog/compare/0a8cc35...2e5657b. [Here](https://github.com/saulshanabrook/egglog_repro) is a repo with a reproduction of the slowdown.

> Hey guys, when we upgraded to the latest egglog commit, we saw a bunch of regressions specifically around rules that dont match, which used to take under a ms and now were 10s or 100s of ms.

From the README of the repo, the suggestion is that the tree composition may be the issue cc @yihozhang:

> set_report_level(ReportLevel::WithPlan) triggers a separate todo!() panic in core-relations/src/free_join/plan.rs:254 (DecomposedPlan::to_report) on exactly these rules. That strongly suggests the rules are now compiling to a tree-decomposed query plan that the new free-join planner picks poorly for this LHS shape, and that this is the same code path that drives the runtime cost. (The repro here uses TimeOnly to avoid that panic; the slowness shows up regardless.)

yihozhang:
Oh thanks for the report! I'll take a look

yihozhang:
The slowdown might be due to #863 , and the performance using CLI seems to be very good using this branch https://github.com/egraphs-good/egglog/pull/857

saulshanabrook:
I am working on this now. First hypothesis of "this is due to multiple threads" is invalidated:

>   For qwen_one_cublaslt_rule.egg, target rule time on the new commit:
> 
>   default:              109.6, 111.5, 109.8 ms
>   RAYON_NUM_THREADS=1:  104.1, 104.5, 103.3 ms
>   RAYON_NUM_THREADS=16: 109.0, 109.8, 108.8 ms
> 
>   Old commit, same file/rule:
> 
>   default:              0.072, 0.078, 0.077 ms
>   RAYON_NUM_THREADS=1:  0.056, 0.051, 0.051 ms
>   RAYON_NUM_THREADS=16: 0.080, 0.070, 0.071 ms

saulshanabrook:
https://github.com/egraphs-good/egglog/pull/857 does help but does not resolve all the slowdowns. It does make the total egglog program now faster than old, but some of the smaller pieces are still slower.  The `qwen_all` takes on average 70% of the time on serial mode on the newest egglog commit compared to the old commit on serial mode. 

[...]

PR #896, opened 2026-05-23
Options to disable tree decomposition

yihozhang:
This is a temporary fix for #872 and avoids future issues caused by tree decomposition by giving the user the ability to disable it.


saulshanabrook (approved):
Thanks for this! I am re-running the benchmarks. It seems like they help some of the benchmarks a ton to turn this off, and others less so. Some are still a lot slower it looks like even with this turned off

Issue #751, opened 2025-11-28
egglog API uses FunctionId, TableId and symbols


oflatt:
Currently the egglog API is awkward to use because there are three ways to reference a table. `Command`s use unique string names for constructors and functions. The rust API generally uses `FunctionId` on the surface, except when defining primitives which drop down to `TableId`. 

I propose we change the entire API to all use strings or symbols for simplicity, hiding the ugly database underneath.

Issue #745, opened 2025-11-21
Directly add/check facts


asvarga-sedaro:
What is the recommended way to directly add or check facts in tables via the Rust API? I can use `EGraph::{function_to_dag, lookup_function}` to check exact facts like `(R x y)`, but is there a way to perform more general queries like `(R x _)` without going through `EGraph::run_program`?

PR #901, opened 2026-05-28
Name-indexed read/write API on Read/Write traits + EGraph::update

oflatt:
## Summary

Adds a name-indexed read/write API for driving the e-graph from Rust without `parse_and_run_program`. Closes the user-facing piece of #751: the egglog crate's public surface no longer exposes `FunctionId` / `TableId` / `ExternalFunctionId`.

The methods live on the existing `Read` / `Write` capability traits — the same surface a `rust_rule` / `add_rust_rule_full` callback already sees. Outside a rule, the new `EGraph::update(|fs| ...)` hands out the same `FullState` handle.

```rust
// Inside a rule callback (WriteState / FullState): unchanged.
ws.set("f", (1_i64,), 42_i64)?;
ws.add("Cons", (1_i64, nil))?;

// Outside a rule:
eg.update(|mut fs| fs.set("f", (1_i64,), 42_i64))?;
let got = eg
    .update(|fs| fs.lookup("f", 1_i64))?
    .map(|v| eg.value_to_base::<i64>(v));

// Pattern queries stay on EGraph — they compile a one-shot rule.
let rows = eg.query(vars![x: i64], facts![(R x)])?;
for m in rows { let x = eg.value_to_base::<i64>(m["x"]); /* ... */ }
```

## What's added

- `Read::lookup<K: IntoValues>(name, key) -> Result<Option<Value>, Error>` — function-table read. Output stays as a raw `Value` so containers / eclasses pass through cleanly; extract with `EGraph::value_to_base` / `value_to_container`.
- `Read::eclass_of<K: IntoValues>(name, inputs) -> Result<Option<Value>, Error>` — constructor lookup without minting.
- `Read::contains<K: IntoValues>(name, key) -> Result<bool, Error>` — row presence, any subtype.
- `Read::function_entries(name) -> Result<Vec<(Vec<Value>, Value)>, Error>` — iterate a function table.
- `Read::constructor_enodes(name) -> Result<Vec<(Vec<Value>, Value)>, Error>` — iterate a constructor / relation table.
- `Read::table_size(name)` / `Read::table_sizes()` — row-count accessors.
- `Write::set<K, V>`, `Write::add<R>`, `Write::remove<K>`, `Write::subsume<K>`, `Write::union(x, y)`, `Write::panic()` — write surface.
- `EGraph::update(|fs| ...) -> Result<R, Error>` — top-level accessor for the same surface. Takes a `Result`-returning closure so proof-mode-incompatibility and the closure's own errors collapse into one `?`. Bridge layer's `with_execution_state` is the underlying primitive.
- `EGraph::query(vars, facts) -> Result<Vec<HashMap<String, Value>>, Error>` — pattern-query iteration. One map per match, keyed by variable name.
- Conversion trait surface in `crate::api`: `IntoValue` / `FromValue` for single columns, sealed `IntoValues` / `FromValues` for whole rows (tuples up to a fixed arity), plus `RawValues(Vec<Value>)` for variadic / pre-converted rows.

## Subtype + arity checks

Detectable misuse is reported through `Error::ApiError`, not panics:

- `WrongSubtype` — `set` on a constructor, `add` on a function, `lookup` on a constructor, `eclass_of` on a function, `function_entries` on a constructor, `constructor_enodes` on a function.
- `WrongArity` — wrong number of input columns for the table.
- `MissingTable` — table name not registered.

The subtype check is plumbed through a new `egglog_bridge::TableKind` enum on `TableAction` (Function vs Constructor), derived from the function's default-value kind at table-registration time.

## Type safety caveat

The API is **type-unsafe at the column level**. Eclass identifiers flow as bare `Value`s with no sort tag at the Rust type level or at runtime: a `Value` returned from a `Math` constructor and a `Value` returned from a `List` constructor are indistinguishable. `Write::union` does not check that its two arguments are the same eq-sort — the caller's responsibility. `Write::set` and `Write::add` do not check that their column values match the table's declared column sorts — also the caller's responsibility. (See follow-up #2 for what real safety would cost.) The bridge's `FunctionConfig.sort_names` is plumbed for that future work but currently unused by the egglog-crate API.

## Proof-mode compatibility

`rust_rule` / `rust_rule_full` and `EGraph::update` return `Error::ProofsIncompatibleApi` upfront on an `EGraph::new_with_proofs()` — the rule body is a Rust closure with no proof-encoding validator, and `update` writes bypass the proof pipeline. `tests/api_proofs.rs` locks the behavior in.

## Constant-folding case study

`tests/api_const_fold.rs` exercises the surface end-to-end: builds `(Add (Num 2) (Add (Num 3) (Num 4)))` with `update` + `add`, installs a `rust_rule` that folds `(Add (Num a) (Num b))` to `(Num (a+b))`, runs to saturation, and verifies the root collapses to `(Num 9)` via both `eclass_of` and `constructor_enodes`.

## Follow-ups

Ten strands of further work were explored on this branch and parked:

1. **Container / function IDs for the hot path.** The string-keyed `ActionRegistry` HashMap lookup on every `set` / `lookup` / `add` is small (in the noise vs main) but non-zero. A `TableHandle` / `FunctionHandle` opaque newtype obtained once at registration and reused across calls would eliminate the per-call HashMap probe and the table-name cloning. Worth doing if a real workload shows the registry lookup in profiles.

2. **Dynamic type checking via a runtime-tagged `Id` wrapper.** Six follow-on commits (Stages 2–6 + a perf-recovery commit) extended an earlier version of this PR with an `Id` type pairing each `Value` with its egglog sort name. Reads returned `Option<Id>` / `Vec<Id>`; `union` cross-sort-mismatched as an error; `extract::<T>` validated the sort tag. After the perf recovery (`SmallVec` rows, cached sort `Arc<str>`s, etc.) hot-path benches were within ~1.4× of baseline (fib: 1.42×; insert/tableaction paths actually *faster* than baseline). Removing the last user-facing escape hatches (`RawValues`, `IntoValue for Value`) cleanly required per-signature wrapper registration through `SpecializedPrimitive::external_id` — a ~1 day refactor blocked by polymorphic primitives (`=`, `+`, `unstable-fn`, container ops). The whole stack is preserved at the annotated tag **`oflatt-dynamic-type-checks`** (commit `0361ca5e`); resume by branching off it. The bridge-side `FunctionConfig.sort_names` plumbing is already in place as a hook.

3. **Efficient eclass traversal — `Read::parents(eclass)`.** Useful for e-substitution primitives (copy a sub-egraph with one eclass swapped for another, preserving congruence) and anything that needs "given eclass `e`, give me every constructor row whose output is `e`." Today the only way is `Read::constructor_enodes` per table, filtering by output column — same O(database size) cost as egglog's internal `extract.rs`. The machinery for a fast version already exists: `core_relations::ColumnIndex` (`hash_index/mod.rs`) supports per-column reverse lookup, and `WrappedTable::group_by_col(subset, col_id)` builds one in a single call. The query executor already caches these per-column indexes per query (`free_join/execute.rs:707`). The shape: add a `Read::parents(eclass: Value) -> impl Iterator<Item = (&str, &[Value])>` backed by lazily-built `ColumnIndex`es on the output column of every constructor table, cached on the `ExecutionState` so the index is paid for once per primitive invocation (or once per rule iteration). Also wants a `Read::table_kind(name) -> TableKind` to filter to constructors only, and surfacing per-column sort metadata so generic substitution can know which columns are eclasses vs base values.

4. **Proof support for `rust_rule` / `update`.** Both APIs currently error with `Error::ProofsIncompatibleApi` under `EGraph::new_with_proofs()` — see `tests/api_proofs.rs`. `rust_rule`'s RHS is a Rust closure with no proof-encoding validator, and `update` writes flush directly into the backend without going through the proof pipeline. Restoring proof compatibility would mean: (a) for `rust_rule`, requiring a user-supplied `PrimitiveValidator` (already a `pub` type, threaded through `add_*_primitive`) and registering on both the main and proof e-graphs so the auto-generated primitive name resolves in proof typechecking; (b) for `update`, threading every staged write into the proof_check egraph as a `(set ...)` / `(union ...)` / etc. action via the same desugaring path `parse_and_run_program` already uses. (b) is the harder half — the closure's writes happen inside the backend's `update`, beneath the level at which the proof system observes commands.

5. **Queries from inside `FullState` / `update`.** `EGraph::query` lives on `&mut EGraph` because it allocates a temporary ruleset, registers a `rust_rule` whose body captures matches, runs it, then frees the rule — the parser, type registry, rulesets map, and `&mut backend` are all needed at every step. None of those are reachable from `FullState`, which only holds `&mut ExecutionState`. So today there's no way to run a pattern query from inside an `eg.update(|fs| ...)` closure or a `rust_rule_full` callback. The clean fix is to split it: `EGraph::compile_query(vars, facts) -> Query` does the parse + typecheck + plan once (returns a handle); `Read::run_query(&query) -> Vec<HashMap<String, Value>>` invokes the compiled plan against the current `ExecutionState`. `EGraph::query` then becomes shorthand for `compile_query` + `run_query` + drop. This is also the natural fit for `Read::parents(eclass)` (follow-up #3) — both want a pre-compiled query plan reusable across calls. **Pre-requisite:** the typechecker currently runs through the full `EGraph` pipeline; it'd need to be split out (or at least made callable on a subset of state) so `compile_query` can typecheck patterns without going through `run_program`. That extraction is the bulk of the work.

6. **Convenience extraction by global name.** `EGraph::extract_value(&sort, value)` requires the caller to already have the sort and the value handle. The common pattern of "let-bind a term at the top level and extract it" works today via `EGraph::eval_expr` on the global name — `eg.eval_expr(&exprs::var("$root"))?` returns `(sort, Value)` which feeds straight into `extract_value`. A thin wrapper `EGraph::extract_global(name: &str) -> Result<(TermDag, TermId, Cost), Error>` (plus a `_with_cost_model` variant) would collapse those two calls. Once it lands the docs can advertise:
    ```rust
    eg.parse_and_run_program(None, "(let $root (Add (Num 1) (Num 2)))")?;
    let (dag, term, _) = eg.extract_global("$root")?;
    ```
    instead of the current two-step. Globals are already tracked (`EGraph::is_let_binding`); the missing piece is a single accessor that goes from name → (sort, value) and forwards to the existing extractor.

7. **Move extraction onto `Read`.** Currently `EGraph::extract_value` / `EGraph::extract_value_with_cost_model` / `EGraph::extract_value_to_string` are top-level methods on `EGraph` — even though they only read the database. Now that every other "read the e-graph" operation lives on the `Read` trait (`lookup`, `eclass_of`, `function_entries`, `constructor_enodes`, `contains`, `table_sizes`), extraction is the odd one out. Move the methods to `Read::extract_value` (and friends), so callers go `eg.update(|fs| fs.extract_value(sort, value))` like every other read. Same surface inside rule callbacks for free. The internals already work via the bridge state; the wrapper plumbing is straightforward (build an `Extractor` against the current `ExecutionState`, call `extract_best`). The only non-trivial piece is that `Extractor::compute_costs_from_rootsorts` takes a `&EGraph`, not an `ExecutionState` — would need either threading through `ActionRegistry` or accepting a tighter interface.

8. **Drop `&mut self` on container interning.** Container interning takes `&mut self` where `&self` would do — costs nothing now but blocks parallel read access to the registry.

9. **Trim the Rust surface.** Hide internal IR types from the prelude re-export, so the public surface signals `parse_and_run_program` as the primary path.

10. **`egglog!` proc macro for compile-time-validated inline programs.** Let users write egglog inside a Rust expression with the source parsed and typechecked at compile time, so program errors become `cargo build` errors. The full-typecheck version depends on follow-up #5 (typechecker extraction); a parsing-only version can ship first.

Refs #745, #751, #895.


egglog-bridge/src/lib.rs
```diff
@@ -277,7 +277,7 @@ impl EGraph {
     ///
     /// # Seminaive-safety trust boundary
     ///
-    /// Like [`EGraph::with_execution_state`], this is a raw escape —
+    /// Like [`EGraph::update`], this is a raw escape —
```
yihozhang: update is not a raw escape

src/exec_state.rs
```diff
@@ -132,11 +134,6 @@ pub trait Core<'a, 'db: 'a>: Internal<'a, 'db> {
     fn should_stop(&self) -> bool {
         self.es().should_stop()
     }
-    /// Human-readable name for a table id, if registered.
-    fn table_name(&self, table: TableId) -> Option<&'a str> {
```
yihozhang, line 162: why is this removed?

src/exec_state.rs
```diff
@@ -155,12 +152,13 @@ pub trait Core<'a, 'db: 'a>: Internal<'a, 'db> {
         cv.register_val(container, es)
     }
 
-    /// Convert an egglog [`Value`] to a Rust base type.
+    /// Convert an egglog [`Value`] to a Rust base type. Untyped: trust
+    /// that `x` belongs to sort `T`.
```
yihozhang:
```suggestion
    /// Convert an egglog [`Value`] to a Rust base type, assuming that the
    /// value belongs to `T`.
```

src/exec_state.rs
```diff
@@ -155,12 +152,13 @@ pub trait Core<'a, 'db: 'a>: Internal<'a, 'db> {
...
     fn value_to_base<T: BaseValue>(&self, x: Value) -> T {
         self.es().base_values().unwrap::<T>(x)
     }
 
-    /// Convert a Rust base type to an egglog [`Value`].
+    /// Convert a Rust base type to an egglog [`Value`]. Untyped.
```
yihozhang:
```suggestion
    /// Convert a Rust base type to an egglog [`Value`].
```

src/lib.md
```diff
@@ -0,0 +1,42 @@
+# egglog
+
+Egglog is a language for writing equality saturation applications.
+It is the successor to the Rust library [egg](https://github.com/egraphs-good/egg).
+Egglog is faster and more general than egg.
```
yihozhang:
```suggestion
Egglog is more expressive than egg and scales better to complex workloads.
```

src/lib.rs
```diff
@@ -1112,23 +1102,6 @@ impl EGraph {
         self.functions.iter()
     }
 
-    /// Read the contents of the given function.
-    /// The callback f is called with each row and its subsumption status.
-    ///
-    /// Raises an error if the function does not exist.
-    pub fn function_for_each(
```
yihozhang, line 1241: Why delete these methods? They are faster than creating a FullState and querying through that and should be preferred.

FullState / ExecutionState are only really needed when you want to directly update the e-graph.

src/lib.rs
```diff
@@ -1955,19 +1929,6 @@ impl EGraph {
         self.backend.table_size(function_id)
     }
 
-    /// Lookup a tuple in afunction in the e-graph.
-    ///
-    /// Returns `None` if the tuple does not exist.
-    /// `panics` if the function does not exist.
-    pub fn lookup_function(&self, name: &str, key: &[Value]) -> Option<Value> {
```
yihozhang, line 2213: Same here

src/lib.rs
```diff
@@ -1917,7 +1890,8 @@ impl EGraph {
         &self.overall_run_report
     }
 
-    /// Convert from an egglog value to a Rust type.
+    /// Convert from an egglog value to a Rust type. Untyped — does
+    /// not check whether `x` actually belongs to sort `T`.
```
yihozhang:
```suggestion
    /// Convert from an egglog value to a Rust type. 
    /// This method assumes `x` belongs to sort `T`.
```

src/lib.rs
```diff
@@ -2448,6 +2506,14 @@ pub enum Error {
         command: String,
         reason: ProofEncodingUnsupportedReason,
     },
+    #[error(
+        "`{api}` is incompatible with proof mode: {reason} \
+         Disable proofs or move the operation into a (parse-and-run) egglog program."
```
yihozhang:
```suggestion
         Disable proofs or make the operation a command in the syntax of the egglog language and use `EGraph::parse_and_run`."
```

src/api/mod.rs
```diff
@@ -0,0 +1,256 @@
...
+// Row conversion: IntoValues + FromValues (sealed)
+// ---------------------------------------------------------------------
+
+/// Convert a Rust value into a row of egglog [`Value`]s.
+///
+/// Sealed; impls cover bare [`IntoValue`] (single-column row),
+/// tuples up to arity 8 of [`IntoValue`] values, and [`RawValues`]
+/// for variadic / pre-converted rows.
+pub trait IntoValues: sealed::Sealed {
+    fn into_values(self, bv: &BaseValues) -> Vec<Value>;
```
yihozhang: This seems very costly

src/exec_state.rs
```diff
@@ -224,35 +262,123 @@ pub trait Read<'a, 'db: 'a>: Core<'a, 'db> + RegistrySealed<'a, 'db> {
...
+        let action = lookup_action(self.registry(), name)?;
+        let key_values = key.into_values(self.base_values());
+        check_arity(name, &action, key_values.len())?;
+        action.subsume(self.es_mut(), key_values.into_iter());
+        Ok(())
+    }
+
+    /// Union two values in the e-graph's union-find. The caller is
+    /// responsible for ensuring both values belong to the same
+    /// eq-sort — there is no runtime sort check.
```
yihozhang:
```suggestion
    /// eq-sort.
```

src/exec_state.rs
```diff
@@ -224,35 +262,123 @@ pub trait Read<'a, 'db: 'a>: Core<'a, 'db> + RegistrySealed<'a, 'db> {
...
 }
 
 /// Action-side write methods — name-indexed inserts/removes/subsumes
 /// plus union and panic. Implemented for [`WriteState`] and
 /// [`FullState`]; *not* for [`PureState`] or [`ReadState`].
+///
+/// Detectable misuse (wrong table subtype, wrong arity) is reported
+/// as [`crate::ApiError`] via the method's `Result`. Per-column sort
+/// matching is **not** checked at this layer — the API is type-unsafe
+/// at the column level.
```
yihozhang:
```suggestion
/// matching is **not** checked at this layer.
```

src/exec_state.rs
```diff
@@ -224,35 +262,123 @@ pub trait Read<'a, 'db: 'a>: Core<'a, 'db> + RegistrySealed<'a, 'db> {
...
+    fn function_entries(&self, name: &str) -> Result<Vec<(Vec<Value>, Value)>, Error> {
+        let action = lookup_action(self.registry(), name)?;
+        check_subtype(name, &action, TableKind::Function, "function")?;
+        Ok(collect_rows(&action, self.es()))
+    }
+}
+
+/// Split each row into `(inputs, output)`, where output is the last
+/// column. Shared by `constructor_enodes` and `function_entries`.
+fn collect_rows(action: &TableAction, state: &ExecutionState<'_>) -> Vec<(Vec<Value>, Value)> {
```
yihozhang: The return type will do `n+1` allocations for `n` tuples, which is too much. Can you return a type that wraps the `RowBuffer` or `TaggedRowBuffer`? Use `scan_bounded` to construct such a TaggedRowBuffer?

yihozhang (changes requested):
See comments!

I am also concerned that reads in ExecState in general cannot see the writes that an ExecState have made. For example, the newly added for_each/for_each_while cannot see the stage_insert/stage_remove that have been made before. I see in the documentation that you have mentioned this, but I think this could be a very bad footgun.

src/api/mod.rs
oflatt: Did you have a suggestion for what to do instead? idk how else would you convert base values to vecs of values? This is for the convenience of the api

src/exec_state.rs
oflatt: TableId isn't part of the API right now

src/api/mod.rs
yihozhang: I think make into_values return an Iterator of values, and in the call site use SmallVec to instantiate the vec. This is how TableAction is implemented.

(Some historical notes after some digging: In the initial implementation of the new backend, we used Vec<Value> for insert/subsume (similar to what's here), and that was too costly, so we switched to &[Value] and used a scratch buffer. But because of issues with ownership we removed scratch in #838 and use a SmallVec instead. It seems SmallVec does not cause a huge slowdown)

src/lib.rs
oflatt: Looks like my constructor_enodes was about 2x slower than these methods. However, with a couple optimizations (don't flush if you only did reads and your optimization above) we can get parity. 

I want to consolidate the API so we don't have duplicate endpoints, so I'm in favor of still deleting now that they are the same speed.

src/exec_state.rs
oflatt: done! helped a lot

oflatt:
> @yihozhang said: I am also concerned that reads in ExecState in general cannot see the writes that an ExecState have made. For example, the newly added for_each/for_each_while cannot see the stage_insert/stage_remove that have been made before. I see in the documentation that you have mentioned this, but I think this could be a very bad footgun.

Do you have a suggestion to mitigate this concern?  I agree this could be a stumbling point, but fundamentally egglog is transactional because you have to rebuild if you did write to the egraph. So people just have to learn that you have to commit before you read writes, which is common for databases.

src/api/mod.rs
oflatt: done

src/api/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+// ---------------------------------------------------------------------
+// Row conversion: IntoValues + FromValues (sealed)
+// ---------------------------------------------------------------------
+
+/// Convert a Rust value into a row of egglog [`Value`]s.
+///
+/// Yields the row's columns as an iterator, letting the caller choose
+/// the container — call sites collect into an inline `SmallVec`, so the
+/// common small-arity row stays off the heap.
+///
```
yihozhang:
```suggestion
```

src/exec_state.rs
```diff
@@ -32,11 +32,19 @@
...
-    TableId, Value,
+    Value,
 };
-use egglog_bridge::{ActionRegistry, TableAction};
+use egglog_bridge::{ActionRegistry, RowScan, TableAction, TableKind};
+use smallvec::SmallVec;
+
+/// Inline scratch for a row of column values. Matches the
+/// `SmallVec<[_; 8]>` that `egglog_bridge::TableAction` uses internally,
+/// so rows up to 8 columns — the common case — never touch the heap.
```
yihozhang:
```suggestion
/// `SmallVec<[_; 8]>` that `egglog_bridge::TableAction` uses internally.
```

src/exec_state.rs
```diff
@@ -224,35 +268,160 @@ pub trait Read<'a, 'db: 'a>: Core<'a, 'db> + RegistrySealed<'a, 'db> {
...
+
+/// A read-only snapshot of a table's rows, backed by a single buffer.
+///
+/// Returned by [`Read::constructor_enodes`] and [`Read::function_entries`].
+/// Iterate it with [`Rows::iter`]; each item is `(inputs, output)` as raw
+/// [`Value`]s borrowed from the buffer, so there is no per-row allocation.
+/// `output` is the trailing column — the
+/// eclass id for a constructor, the mapped value for a function. Convert
+/// individual columns with [`Core::value_to_base`] / [`Core::value_to_container`].
+pub struct Rows {
```
yihozhang: There is already a `core-relations` type called `Rows`. Maybe a different name.

src/exec_state.rs
```diff
@@ -224,35 +268,160 @@ pub trait Read<'a, 'db: 'a>: Core<'a, 'db> + RegistrySealed<'a, 'db> {
...
+}
+
+/// A read-only snapshot of a table's rows, backed by a single buffer.
+///
+/// Returned by [`Read::constructor_enodes`] and [`Read::function_entries`].
+/// Iterate it with [`Rows::iter`]; each item is `(inputs, output)` as raw
+/// [`Value`]s borrowed from the buffer, so there is no per-row allocation.
+/// `output` is the trailing column — the
+/// eclass id for a constructor, the mapped value for a function. Convert
+/// individual columns with [`Core::value_to_base`] / [`Core::value_to_container`].
```
yihozhang:
```suggestion
/// Iterate it with [`Rows::iter`]; each item is `(inputs, output)` as raw
/// [`Value`]s. `output` is the trailing column — the
/// eclass id for a constructor, the mapped value for a function. Convert
/// individual columns with [`Core::value_to_base`] / [`Core::value_to_container`].
```

We should make a repo-level skill file that cleans up documentation and comments in a diff to get rid of implementation details and duplicate comments..

src/exec_state.rs
```diff
@@ -224,35 +268,160 @@ pub trait Read<'a, 'db: 'a>: Core<'a, 'db> + RegistrySealed<'a, 'db> {
...
+    }
+
+    /// Iterate `(inputs, output)` pairs, borrowing from the backing
+    /// buffer (no per-row allocation).
+    ///
+    /// This is the monomorphized fast path — as cheap as a direct
+    /// streaming backend scan. (We deliberately do *not* implement
+    /// `IntoIterator for &Rows`: expressing its associated `IntoIter`
+    /// type would force a boxed `dyn Iterator`, whose per-row virtual
+    /// dispatch measurably slows the scan.)
```
yihozhang:
```suggestion
    /// Iterate `(inputs, output)` pairs from rows.
```

egglog-bridge/src/lib.rs
```diff
@@ -1294,6 +1334,66 @@ impl TableAction {
...
+        }
+        drain_buf!(buf);
+    }
+
+    /// Scan every row of this table into a single [`RowScan`].
+    ///
+    /// Unlike [`TableAction::for_each`], the rows materialize into one
+    /// buffer (one allocation, often pooled) that the caller can iterate
+    /// after the [`ExecutionState`] borrow ends — without a heap allocation
+    /// per row.
```
```suggestion
    /// buffer.
```

egglog-bridge/src/lib.rs
```diff
@@ -1569,6 +1669,39 @@ pub struct FunctionRow<'a> {
     pub subsumed: bool,
 }
 
+/// Every row of a table scanned into a single buffer, produced by
+/// [`TableAction::scan_all`].
+///
+/// Owns its backing storage, so it can outlive the [`ExecutionState`] it was
+/// scanned from. [`RowScan::iter`] yields one [`FunctionRow`] per row,
+/// borrowing from the buffer — there is no per-row allocation.
```
```suggestion
/// scanned from.
```

src/lib.rs
yihozhang: I still prefer function_for_each to be exposed at the top level. The reasons we had for not having top-level convenience methods are 1) they are a footgun for write operations, and 2) we should encourage people to use the language whenever possible, but neither applies to function_for_each. Moreover, in my mental model, function_for_each feels like a different modality for using the e-graph (scanning tables, extracting information) compared to using with_execution_state (more tuple-level ops), so it makes sense to me they are different top-level operators.

I am fine with deleting lookup_functions.

oflatt:
Added a skill for cleaning up docs! Sorry for all of those.

Re: function_for_each, that's fair. In the interest of making the API more uniform, I've still deleted `function_for_each` but replaced it with `EGraph::function_entries` and `EGraph::constructor_enodes`. So you only have to learn one API. Let me know what you think.

src/lib.rs
oflatt: see below-added back but modified

yihozhang:
Sorry for the late review! I am still in favor of `function_for_each`, or any function that takes a callback and does not clone the entire table, which seems unnecessary in most cases.

Actually, the documentation on "function_entries" (resp. "enode_rows") says  "Iterate every `(inputs, output)` entry of a function table." (resp. "Iterate every enode in a constructor / relation table."), which suggests it should not create a copy of the table? Is the version that takes a callback impossible to implement for `Read`?

Otherwise the PR LGTM

oflatt:
@yihozhang you're right, it was doing a copy. I've done another re-design to make it similar to the old `function_for_each` but keeping the split between `constructor_enodes` and `function_entries`. I'm happier with this design, it also led to a new helper called `EGraph::read` which gives you a read-only closure. I also removed the `scan_all` helper that was not unused, encouraging people to use the egraph without creating copies. 

lmk what you think of this version

yihozhang (commented):
I like the new design. Could you rename ScanEntry back to FunctionRow? It causes some  unneeded interface changes in extract.rs. Actually, if you feel strongly, you can also touch up the CostModel interface a little bit to be consistent about the new enode / function entry naming convention.

oflatt:
Great catch, I didn't realize extract.rs needed to be changed because there was no compile error for leaking egglog-bridge. See the new issue #932

yihozhang (approved):

PR #902, opened 2026-05-30
Report the actual type name when a primitive references an unregistered sort

### LINKED ISSUE #732: Better Error Message For Missing Sort
I noticed that when defining a primitive and it references a sort that isn't defined you get a bit of a cryptic error message. It also doesn't tell you which sort was missing, it just says `egglog_core_relations::base_values::Boxed<alloc::string::String>`, which doesn't tell you much:

```
assertion `left == right` failed: Expected exactly one sort for type egglog_core_relations::base_values::Boxed<alloc::string::String>
  left: 0
 right: 1
```

aeft:
Adds a dedicated `get_arcsort_for_value_type::<T>()` so `add_primitive!` reports the real missing type name.

Fixes #732

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1843,6 +1843,11 @@ impl EGraph {
         self.type_info.get_arcsort_by(f)
     }
 
+    /// Returns the unique sort whose runtime values have Rust type `T`.
+    pub fn get_arcsort_for_value_type<T: 'static>(&self) -> ArcSort {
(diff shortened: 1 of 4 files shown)


aeft:
@saulshanabrook Could you please have a look when you have a chance?

saulshanabrook (approved):
This looks great! Thank you!

PR #908, opened 2026-06-02
Return an error instead of panicking when running an undefined ruleset

### LINKED ISSUE #904: `run` produces unintelligable message, when given an undefined rule-set name.
```
(ruleset test)
(run test2 1)
```

where `test2` is undefined, produces a mysterious message:

```
thread 'main' (105861) panicked at /home/roman/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/egglog-2.0.0/src/lib.rs:859:28:
no entry found for key
```

SAY-5:
Running a ruleset that was never defined, e.g. `(run test2 1)`, indexed `self.rulesets` directly in `run_rules` and panicked with `no entry found for key` instead of producing a useful error. This checks that the ruleset exists first and returns the existing `NoSuchRuleset` error with the call's span, matching how `add_rule` already handles an unknown ruleset. Fixes #904.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -969,6 +969,10 @@ impl EGraph {
 
         let GenericRunConfig { ruleset, until } = config;
 
+        if !self.rulesets.contains_key(ruleset) {
+            return Err(Error::NoSuchRuleset(ruleset.clone(), span.clone()));
+        }
+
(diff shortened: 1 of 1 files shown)


saulshanabrook (approved):
Thank you!

PR #910, opened 2026-06-03
Replace Rayon with our own, non-workstealing thread pool

ezrosent:
Remove the reliance on the rayon global thread pool, while maintaining single and multi-threaded performance in the `math` and `cykjson` microbenchmarks.

This should make us a lot more resilient to issues where parallelism is enabled by accident.

### DIFF
--- a/core-relations/src/hash_index/mod.rs
+++ b/core-relations/src/hash_index/mod.rs
@@ -3,24 +3,24 @@ use std::{
     cmp,
     hash::{Hash, Hasher},
     mem,
-    sync::{Arc, Mutex},
+    sync::Mutex,
 };
 
 use crate::{
     common::IndexMap,
     numeric_id::{IdVec, NumericId, define_id},
 };
-use egglog_concurrency::{Notification, ReadOptimizedLock};
+use egglog_concurrency::{ReadOptimizedLock, ThreadPool};
 use hashbrown::HashTable;
 use indexmap::map::Entry;
 use once_cell::sync::Lazy;
-use rayon::iter::ParallelIterator;
 use rustc_hash::FxHasher;
 
 use crate::{
     OffsetRange, Subset,
     common::{ShardData, ShardId, Value},
     offsets::{RowId, SortedOffsetSlice, SubsetRef},
+    parallel,
     parallel_heuristics::parallelize_index_construction,
     pool::{Pooled, with_pool_set},
     row_buffer::{RowBuffer, TaggedRowBuffer},
@@ -290,8 +290,8 @@ impl IndexBase for ColumnIndex {
             }
         };
 
-        run_in_thread_pool_and_block(&THREAD_POOL, || {
-            rayon::in_place_scope(|inner| {
+        run_in_index_thread_pool(|| {
+            egglog_concurrency::scope(|inner| {
                 let mut cur = Offset::new(0);
                 loop {
                     let mut buf = TaggedRowBuffer::new(cols.len());
@@ -307,7 +307,7 @@ impl IndexBase for ColumnIndex {
                 }
             });
 
-            self.shards.par_iter_mut().for_each(|(shard_id, shard)| {
+            parallel::for_each_id_vec_mut(&mut self.shards, |shard_id, shard| {
                 // Sort the vector by start row id to ensure we populate subsets in sorted order.
                 let mut vec = queues[shard_id].lock().unwrap();
                 vec.sort_by_key(|(start, _)| *start);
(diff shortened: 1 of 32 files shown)


ezrosent:
Starting off as a draft as there are some more cleanups I'd like to look into

saulshanabrook:
I haven't done a full review yet, but to note this does have some significant slowdowns on some of the luminal benchmarks vs the main branch running on multiple threads:


pr910_multithreaded_timings

pr910_multithreaded_percent_change


See for how to reproduce these graphs: 

https://github.com/saulshanabrook/egglog_repro#reproduce

ezrosent:
@saulshanabrook @yihozhang 

I made progress here and got some but not all of the overhead down. I think that the remaining overhead can be fixed by changing how materialization works: I think not using work-stealing is causing more contention over the DashMap when doing a scoped materialization. I think we can probably switch away from dashmap for this step entirely though. Thoughts on submitting this for now and then continuing to optimize the concurrency of the tree decomp stuff later?

saulshanabrook:
OK yeah performance changes seems more reasonable now, some are slower, some are faster... Worst slowdown is around ~17%

pr910_multithreaded_timings
pr910_multithreaded_percent_change

core-relations/src/parallel.rs
```diff
@@ -0,0 +1,179 @@
...
+}
+
+fn default_vec<T: Default>(len: usize) -> Vec<T> {
+    vec_with(len, T::default)
+}
+
+fn vec_with<T>(len: usize, init: impl FnMut() -> T) -> Vec<T> {
+    let mut values = Vec::with_capacity(len);
+    values.resize_with(len, init);
+    values
```
saulshanabrook: These aren't parallel specific just helpers?

egglog-bridge/src/lib.rs
```diff
@@ -184,25 +244,35 @@ impl Default for EGraph {
...
+    }
+
+    /// Set the number of worker threads used by this e-graph.
+    ///
+    /// Passing `1` disables the pool. Passing `0` uses available parallelism.
+    pub fn set_num_threads(&mut self, threads: usize) {
+        let threads = normalize_thread_count(threads);
+        self.threads = threads;
+        self.thread_pool = (threads > 1).then(|| Arc::new(ThreadPool::new(threads)));
+    }
```
saulshanabrook, line 265: Cool so we can change the number of threads on the fly? That's nice

saulshanabrook (approved):

saulshanabrook:
@coderabbitai full review

saulshanabrook:
I asked my agent for a more holistic review of this feature, and was curious about this option of just using a per-egraph rayon threadpool (instead of the global one) to avoid the issue we had about the global setup... That would keep the work stealing aspect, which it wasn't clear was a problem and slowed things down in some cases?

Or are there other reasons to avoid work stealing?

You know generally in favor of keeping as little LoC as possible lol but I imagine you have thought about this more and have a good reason for moving away from Rayon.

ezrosent:
(alright, back! hopefully we can get this in soon)

The structure you outlined would be great but there is a strange interaction between the default-global thread pool and the per-egraph one. Oliver send a PR for this earlier in the year and we found that it degraded single-threaded performance by a lot, in a way that was very hard to debug. Having our own thread pool is basically us investing in a simpler chunk of code we understand, vs. a better-tuned / more full-featured one that we don't. 

If that design worked I think we'd just use it, to be clear.

core-relations/src/hash_index/mod.rs
```diff
@@ -973,23 +936,21 @@ impl BufferedSubset {
...
-        .unwrap()
-});
+/// Callers can construct indexes while holding database-level index locks. The
+/// separate pool preserves parallelism without tying up the caller's installed
+/// pool behind those locks.
+static INDEX_THREAD_POOL: Lazy<ThreadPool> =
+    Lazy::new(|| ThreadPool::new(parallel::current_num_threads().max(1)));
+
+fn run_in_index_thread_pool<R>(f: impl FnOnce() -> R) -> R {
+    INDEX_THREAD_POOL.install(f)
```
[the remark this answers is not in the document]
ezrosent: (This is intended behavior, and it works in this case because the thread pool is always installed before the first call)

core-relations/src/parallel.rs
ezrosent: They're helpers for the parallel helpers which are exposed to the rest of the crate.

egglog-bridge/src/lib.rs
ezrosent: Yes!

PR #912, opened 2026-06-03
Add `:unsafe-seminaive` rule option for arbitrary RHS reads

oflatt:
## Motivation: the proof encoder

The proof/term encoder instruments every rule with proof bookkeeping. For each eq-sort variable in a rule's body it injected a body fact `(= p (term_proof var))` just to bind that variable's proof for assembling the rule proof. But every eq-sort term has its `term_proof` set at constructor-creation time, so that proof is *always* present when the rule fires — the join never filters. It's pure overhead: one extra join per eq-sort body variable, on every instrumented rule.

We want to fetch those proofs as a direct lookup in the rule's RHS instead. That needs a way to read the database (here, a function-table lookup) from a seminaive rule's actions — which is what `:unsafe-seminaive` provides. This PR adds the option *and* switches the encoder to use it (see the second commit), dropping a join per eq-sort body variable from every proof-instrumented rule.

## The `:unsafe-seminaive` rule option

`:naive` already lets a rule's RHS read the database (read-primitives) by widening its primitive contexts from `Pure`/`Write` to `Read`/`Full` — but it gives up seminaive (delta) evaluation, matching the whole database every iteration. `:unsafe-seminaive` **decouples** the two:

|                         | query/action context | evaluation |
| ----------------------- | --------------------- | ---------- |
| default                 | `Pure` / `Write`      | seminaive  |
| `:naive`                | `Read` / `Full`       | whole-DB   |
| **`:unsafe-seminaive`** | **`Read` / `Full`**   | **seminaive** |

So the RHS can perform arbitrary database reads — read-primitives **and** function-table lookups — with seminaive performance.

### How

- `GenericRule` gains an `unsafe_seminaive` flag, set by the `:unsafe-seminaive` rule option (round-trips through `Display`/parser).
- `typecheck_rule` and `add_rule` compute `context_seminaive = seminaive && !rule.unsafe_seminaive` and use that for primitive-context selection, while evaluation still keys off `seminaive`. The "no function lookups in actions" check (`check_lookup_actions`) is gated on `!unsafe_seminaive`.
- The proof encoder's `instrument_rule` now emits eq-sort `term_proof` fetches as action-side `(let p (term_proof v))` lookups and tags the generated rule `:unsafe-seminaive`.

## Safety

It is **unsafe**, hence the name: a read on a seminaive rule's RHS observes the database mid-iteration, so it won't be re-evaluated if the data changes — results can depend on what's been derived so far. The caller takes responsibility.

User-written `:unsafe-seminaive` is rejected by the term/proof encoding (`command_supports_proof_encoding` → `UnsafeSeminaive`); it's a direct-backend feature. (The encoder's *own* generated rules use it internally — they bypass the user-program check, as encoder output.)

## Tests

Two `.egg` tests, one per arm `:unsafe-seminaive` enables (both accepted on the default backend, rejected under `--term-encoding`/`--proofs`):
- `tests/unsafe-seminaive.egg` — a **function-table lookup** in a rule's RHS (otherwise a `LookupInRuleDisallowed` error).
- `tests/unsafe-seminaive-read-prim.egg` — a **read/Full primitive** (`unstable-multiset-fill-index`) in a rule's RHS, which the default seminaive `Write` action context doesn't admit (the name doesn't even resolve without the option).

Existing proof + proof_testing snapshots are unchanged by the encoder switch.


src/proofs/proof_encoding_helpers.rs
```diff
@@ -505,6 +509,14 @@ pub(crate) fn command_supports_proof_encoding(
     command: &ResolvedCommand,
     type_info: &TypeInfo,
 ) -> Result<(), ProofEncodingUnsupportedReason> {
+    // `:unsafe-seminaive` rules perform arbitrary reads against the live
+    // database; the term/proof encoding can't represent that.
+    if let crate::ast::GenericCommand::Rule { rule } = command {
+        if rule.unsafe_seminaive {
+            return Err(ProofEncodingUnsupportedReason::UnsafeSeminaive);
+        }
+    }
```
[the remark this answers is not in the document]
oflatt: Good catch — done. Updated `tests/snapshots/files__proof_unsupported_files.snap` to include `unsafe-seminaive.egg` and `unsafe-seminaive-read-prim.egg` (commit 3590f847).

src/lib.rs
```diff
@@ -1045,11 +1045,16 @@ impl EGraph {
...
+        // can read the database. See `GenericRule::unsafe_seminaive`.
+        let context_seminaive = seminaive && !rule.unsafe_seminaive;
 
         let rule_id = {
             let mut rb = self.backend.new_rule(&rule.name, seminaive);
             rb.set_no_decomp(no_decomp);
-            let mut translator = BackendRule::new(rb, &self.functions, &self.type_info, seminaive);
+            let mut translator =
+                BackendRule::new(rb, &self.functions, &self.type_info, context_seminaive);
             translator.query(query, false);
```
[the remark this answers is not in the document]
oflatt: Renamed the field/param to `context_seminaive` and clarified its doc (commit 7ac5a350) — it selects the primitive context, not the evaluation strategy, and is `false` for `:unsafe-seminaive` rules even though they evaluate seminaive-ly.

egglog-ast/src/generic_ast.rs
```diff
@@ -103,6 +103,18 @@ where
...
+    /// with the permissive `Read`/`Full` primitive contexts (like
+    /// `:naive`), and the typechecker's "no function lookups in actions"
+    /// check is skipped. This lets the RHS perform arbitrary database
+    /// reads — read-primitives and function-table lookups — without
+    /// paying for `:naive`'s whole-database matching.
+    ///
+    /// It is **unsafe**: a read on a seminaive rule's RHS observes the
+    /// database mid-iteration, so it won't be re-evaluated if the data
+    /// changes. Defaults to `false`.
+    pub unsafe_seminaive: bool,
```
saulshanabrook: Could we collapse this with naive so its an `Option<bool>`? Like None is not set, then its either set to `:naive` or `:unsafe-seminaive`?

src/lib.rs
```diff
@@ -1085,11 +1045,16 @@ impl EGraph {
         // flag) skips tree-decomposition in query planning, forcing
         // the single-bag fast path.
         let no_decomp = self.no_decomp || rule.no_decomp;
+        // `:unsafe-seminaive` keeps delta evaluation (`new_rule(seminaive)`)
+        // but compiles the query/action with Read/Full contexts so the RHS
+        // can read the database. See `GenericRule::unsafe_seminaive`.
+        let context_seminaive = seminaive && !rule.unsafe_seminaive;
```
saulshanabrook: Then this would also be collapsed, just like use rule.naive or if its not there default to self.seminaive?

saulshanabrook (commented):
I support this feature, just I think it should be mutually exclusive with `:naive` in terms of parsing as well as storage, so we can just store one field for them, if I am understanding it correctly?

yihozhang:
Maybe we should add a snapshot test asserting that `unsafe-seminaive` and `naive` produce the same database (for proof code)?

oflatt:
@saulshanabrook collapsed naive and unsafe_seminaive to a RuleEvalMode.

oflatt:
@yihozhang Added snapshot testing, not on all the tests because that would be too slow.

egglog-ast/src/generic_ast.rs
```diff
@@ -80,6 +80,36 @@ where
...
+    pub fn is_naive(self) -> bool {
+        matches!(self, RuleEvalMode::Naive)
+    }
+
+    /// Uses the permissive `Read`/`Full` contexts: true for `:naive` and
+    /// `:unsafe-seminaive`.
+    pub fn uses_read_contexts(self) -> bool {
+        !matches!(self, RuleEvalMode::Seminaive)
+    }
+}
```
saulshanabrook: I am sorry I dont fully understand this, what do you mean uses the read contexts? Like can work within a read context? Like requires read? Would `requires_read_context` also make sense here, that would make more sense in my brain.

src/lib.rs
```diff
@@ -169,6 +169,28 @@ pub enum CommandOutput {
...
+    /// per #793, extraction variants) and reduces `ExtractBest` to its cost.
+    pub fn snapshot_stable_under_proof_encoding(outputs: &[CommandOutput]) -> String {
+        outputs
+            .iter()
+            .filter_map(|output| match output {
+                CommandOutput::OverallStatistics(_) => None,
+                CommandOutput::PrintFunction(..) => None,
+                CommandOutput::ExtractBest(_, cost, _) => {
+                    Some(format!("(extraction-costs {cost})\n"))
+                }
```
saulshanabrook, line 185: Why do we just wrap ExtractBest in a s-expr but not other outputs?

src/lib.rs
```diff
@@ -1134,16 +1156,22 @@ impl EGraph {
...
+        let seminaive = self.seminaive && !rule.eval_mode.is_naive();
         // The `:no-decomp` rule option (and the global `--no-decomp`
         // flag) skips tree-decomposition in query planning, forcing
         // the single-bag fast path.
         let no_decomp = self.no_decomp || rule.no_decomp;
+        // `:naive` (or `eg.seminaive == false`) and `:unsafe-seminaive` both
+        // compile the query/action with Read/Full contexts so the RHS can
+        // read the database. `:unsafe-seminaive` still evaluates seminaively
+        // (`new_rule(seminaive)`). See `RuleEvalMode`.
+        let context_seminaive = seminaive && !rule.eval_mode.uses_read_contexts();
```
saulshanabrook: I think this might be clearer without uses_read_contexts but just the matches directly...?

src/lib.rs
```diff
@@ -2397,28 +2425,29 @@ struct BackendRule<'a> {
...
-    seminaive: bool,
+    /// Selects the primitive [`crate::Context`] for this rule (combined
+    /// with the query/action phase). `true` picks the seminaive-safe
+    /// `Pure`/`Write` contexts; `false` widens to `Read`/`Full` so
+    /// primitives may read the database. It is named after the *context*,
+    /// not the evaluation strategy: it is `false` for global one-shots
+    /// (`eval`, `check`), `:naive` rules, `EGraph::seminaive == false`,
+    /// and `:unsafe-seminaive` rules — the last of which still evaluate
+    /// seminaive-ly but want the permissive contexts.
+    context_seminaive: bool,
```
saulshanabrook: Would calling this `requires_read_context` also make sense?

src/typechecking.rs
```diff
@@ -846,13 +846,18 @@ impl TypeInfo {
...
-        let seminaive = global_seminaive && !*naive;
-        let (query_ctx, action_ctx) = if seminaive {
-            (Context::Pure, Context::Write)
-        } else {
+        let seminaive = global_seminaive && !eval_mode.is_naive();
+        // `:naive` (or `eg.seminaive == false`) and `:unsafe-seminaive` both
+        // widen the typecheck (and backend) primitive contexts to Read/Full
+        // so the RHS may read the database. `:unsafe-seminaive` keeps
+        // seminaive evaluation; see `RuleEvalMode`.
+        let read_contexts = !seminaive || eval_mode.uses_read_contexts();
```
saulshanabrook: This logic is a bit confusing, with the eval_mode helpers.. Could you just express it based on global_seminaive and the selected eval mode directly? All the boolean operators are pretty confusing. The `seminaive` looks like its just used to produce the `read_contexts` local so could be collapsed.

src/typechecking.rs
```diff
@@ -885,15 +890,19 @@ impl TypeInfo {
         let actions: ResolvedActions =
             assignment.annotate_actions(&mapped_action, self, action_ctx)?;
 
-        self.check_lookup_actions(&actions)?;
+        // Function lookups in actions need a `Full` action context, which only
+        // `:naive` / `:unsafe-seminaive` provide; reject them otherwise.
+        if !eval_mode.uses_read_contexts() {
+            self.check_lookup_actions(&actions)?;
+        }
```
saulshanabrook, line 894: To be clear `check_lookup_actions` verifies that they dont do reads in the actions? Might be clearer to rename this if thats the only check it does so its obivous its skip this check if we use read contexts in the actions?

saulshanabrook:
@coderabbitai full review

egglog-ast/src/generic_ast.rs
oflatt: inlined and is now called required_read_context. I agree!

src/lib.rs
oflatt: yep! thx

src/typechecking.rs
oflatt: yep, should be clearer now

src/typechecking.rs
oflatt: yeah now it's check_no_lookups_in_actions

saulshanabrook (approved):

src/lib.rs
oflatt: idk, existing behavior

PR #915, opened 2026-06-04
Pass user-defined commands through term encoding instead of panicking

oflatt:
## The bug

Running plain term encoding (`--term-encoding`, proofs off) on a program that contains a user-defined command (e.g. one registered by egglog-experimental) panicked in the term-encoding instrumentation:

```rust
ResolvedNCommand::UserDefined(..) => {
    panic!("User defined commands unsupported in term encoding");
}
```

There is nothing for term encoding to instrument here — the user-defined command is dispatched later by whatever registered it (e.g. egglog-experimental's `run-schedule` rebinding or `multi-extract`).

## The fix

- Replace the panic with a pass-through that forwards the command unchanged, matching the adjacent `ResolvedNCommand` arms (`res.push(command.to_command().make_unresolved())`).
- Thread `proofs_enabled` through `command_supports_proof_encoding` so the proof-only checks (missing validators, function lookups in actions, user-defined commands, input commands, no-merge functions, non-eq `let`/`set` bindings) are skipped in proofs-off term encoding but still enforced when proofs are actually generated.

## Safety argument

The pass-through is only safe if proof generation still rejects user-defined commands. It does:

- `command_supports_proof_encoding` returns `Err(ProofEncodingUnsupportedReason::UserDefinedCommand)` for `GenericCommand::UserDefined` whenever `proofs_enabled == true`.
- This gate is enforced in `EGraph::resolve_command_before_proofs` (`src/lib.rs`), which runs **before** the term-encoding instrument path in `resolve_command`. When a user-defined command appears with proofs on, that function returns `Error::UnsupportedProofCommand` and the `UserDefined` instrument arm is never reached.

So the new pass-through is only reachable in plain term encoding (proofs off), where no proofs are emitted. Proof correctness is unaffected.

## Tests

- `cargo test --release --lib proof` — passes.
- `cargo test --release --test files proof_support_snapshot` — passes (`UserDefined` remains proof-unsupported).
- Added a unit test that registers a user-defined command and runs a program containing it under term encoding without proofs, confirming no panic and that the command is forwarded.
- `make fixnits` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased] - ReleaseDate
 
+- Pass user-defined commands through term encoding instead of panicking. Running plain term encoding (`--term-encoding`, proofs off) on a program containing a user-defined command no longer crashes; the command is forwarded to whatever registered it. Proof generation still rejects user-defined commands up front, so the pass-through only affects proofs-off term encoding.
 - Add typed `EGraph` extension state that clones with `EGraph` and is restored by `push`/`pop`.
 - Report full source file paths in egglog span and error messages.
 - Fix seminaive matching after nested containers rebuild in place by propagating dirty container ids through parent containers.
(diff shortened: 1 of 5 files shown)


oflatt:
doesn't look good

PR #917, opened 2026-06-05
[codex] Enable debug assertions in release test builds

saulshanabrook:
## Summary

Release-mode test targets now enable debug assertions through the Cargo profile environment override `CARGO_PROFILE_RELEASE_DEBUG_ASSERTIONS=true`.

This applies to:

- `make test`
- `make coverage`
- `make doctest`

Normal release builds and CodSpeed benchmark jobs continue to use the standard release profile, so debug assertions remain disabled for benchmark and production release artifacts.

Added because we missed a bug recently that would have been found if we had enabled these in testing in CI.

It doesn't seem to slow things down too much, like from 8 minutes to 9 minutes.

## Validation

- `cargo test --release -p egglog-core-relations --lib bad_offset -- --list`
  - confirmed a plain release build lists `0 tests` for `#[cfg(debug_assertions)]` tests
- `CARGO_PROFILE_RELEASE_DEBUG_ASSERTIONS=true cargo test --release -p egglog-core-relations --lib bad_offset`
  - confirmed release-mode test builds with the override execute the debug assertions
- `CARGO_PROFILE_RELEASE_DEBUG_ASSERTIONS=true cargo insta test --test-runner nextest --release --workspace --unreferenced ignore -- math`
- `make -n test coverage doctest`
- `git diff --check`


saulshanabrook:
@coderabbitai full review

yihozhang (approved):

PR #918, opened 2026-06-05
Add eager merge container primitives

saulshanabrook:
## Stack

Draft PR built on top of https://github.com/egraphs-good/egglog/pull/881.

This branch includes the commits from that PR until #881 lands, so review should focus on the final commit, `Add eager merge container primitives`.

## What changed

- Adds `Maybe` and `Either` container sorts with unwrap/match primitives.
- Adds `unstable-if`, a Unit-thunk-conditioned primitive that eagerly returns one of two already-built values.
- Registers `unstable-if`, Maybe match, and Either match in both declaration orders needed by generated egglog programs.
- Adds focused `.egg` tests for `Maybe`, `Either`, `UnstableFn`, and late registration.

## Why

eggcc is moving merge functions away from constructor-producing encodings. These primitives let merge bodies compute eagerly over container values, so generated `:merge` expressions can return an existing value rather than creating a constructor node only to rewrite it later.

## Validation

- `cargo build --release`
- `cargo build --release --bin egglog-experimental` in the dependent egglog-experimental worktree
- `fish -lc 'cargo build --release'` in eggcc
- During development, focused file tests passed for `fn`, `maybe`, `either`, and `pair`.

### DIFF
--- a/src/sort/fn.rs
+++ b/src/sort/fn.rs
@@ -12,7 +12,7 @@
 //! the list of partially applied arguments.
 use std::sync::Mutex;
 
-use crate::exec_state::{ApplyCallError, Internal};
+use crate::exec_state::Internal;
 use enum_map::EnumMap;
 
 use super::*;
@@ -517,7 +517,10 @@ impl FunctionContainer {
     where
         'db: 'a,
     {
+        let ctx = state.ctx();
         let args: Vec<_> = self.1.iter().map(|(_, x)| x).chain(args).copied().collect();
+        let can_mint = matches!(ctx, crate::Context::Write | crate::Context::Full);
+        let can_read = matches!(ctx, crate::Context::Read | crate::Context::Full);
         let panic_id = self.3;
         // On capability mismatch, trigger the egglog runtime panic
         // pre-registered at the `unstable-fn` build site (see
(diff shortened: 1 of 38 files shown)


oflatt:
Just a note, I think maybe and either seem like egglog-experimental type things right? 
I think moving egglog-experimental into this code base would make this a lot easier...

src/sort/fn.rs
```diff
@@ -99,7 +99,7 @@ impl Presort for FunctionSort {
     }
 
     fn reserved_primitives() -> Vec<&'static str> {
-        vec!["unstable-fn", "unstable-app"]
+        vec!["unstable-fn", "unstable-app", "unstable-if"]
```
saulshanabrook, line 102: TODO should remove this extra change here

saulshanabrook:
> Just a note, I think maybe and either seem like egglog-experimental type things right?

Yeah that sounds nice! The issue is that we have to inline certain type registration in the `fn.rs` registration:


        register_maybe_primitives_for_function(eg, self.clone());
        register_if_primitives_for_function(eg, self.clone());
        register_either_primitives_for_function(eg, self.clone());

To move those to experimental, we could add some kind of thing in core to allow registering callbacks here... I think I could do that, in a way that is non specific to unstable function... Would definately be nice to move a bunch of this stuff to experimental I agree! Also to be able to add particular other function registration to other sorts like Map that are in experimental.

        
> I think moving egglog-experimental into this code base would make this a lot easier...

Yeah once all my in flight PRs (besides this one) land in egglog/experimental I will make a PR to see how that looks!

saulshanabrook:
Closing and replacing with an experimental PR

Issue #898, opened 2026-05-26
Report primitive resolution panics as typed errors


saulshanabrook:
This issue is about handling errors more consistently as errors instead of panics, with unstable function. AI summary:

> Split out of PR #889 while narrowing unstable-fn primitive support.
> 
> Current behavior:
> 
> - Direct primitive duplicate resolution can panic in `ResolvedCall::from_resolution` when more than one primitive registration matches the same name, signature, and context.
> - `unstable-fn` over primitive targets now intentionally matches that behavior for duplicate same-signature primitive registrations: it panics instead of returning `Error`, so PR #889 can avoid a broader error-plumbing refactor.
> - Ordinary contextual ambiguity such as a bare `(map-empty)` without enough expected-type information already reports a normal type inference error; the problematic panic case is after types are concrete but multiple indistinguishable primitive registrations remain.
> - `step_rules_with_scheduler` temporarily takes `self.rulesets` and `self.schedulers`; existing fallible `backend.run_rules(...) ?` paths can return before those fields are restored. PR #889 briefly exposed another fallible path there, but that scheduler cleanup is broader than primitive target support.
> 
> Desired follow-up:
> 
> - Make primitive resolution return `TypeError` or `Error` instead of panicking for no-match and ambiguous exact-match cases.
> - Cover both direct primitive calls and primitive targets built through `unstable-fn` with the same diagnostics.
> - Preserve the valid overloaded primitive behavior where typechecking selects one overload, for example `+`, `map-empty`, and container-specific primitives.
> - Add regression tests for duplicate same-signature primitive registrations in direct calls, top-level `unstable-fn`, rule query, rule action, and scheduled rule compilation.
> - Separately harden `step_rules_with_scheduler` so `rulesets` and `schedulers` are restored across every fallible scheduled execution path.
> 
> This is intentionally deferred from PR #889 to keep that PR focused on allowing primitive targets for `unstable-fn` while retaining the current panic semantics for the existing duplicate-registration edge case.

saulshanabrook:
Additional case from PR #881 / CodeRabbit review:

[...]

PR #920, opened 2026-06-05
Return errors instead of panicking for reachable failures

yihozhang:
## Summary

Audits `./src` for `panic!`/`unwrap`/`expect`/`todo!`/`unreachable!` sites that are **reachable from user input** and converts them into proper recoverable errors, so malformed or edge-case programs report an error instead of aborting the process. This addresses a recurring class of bug reports where egglog panics instead of returning an error.

Of 324 panic-family sites in `./src`, an audit classified 52 as reachable (MUST_FIX), 258 as type-checker-guaranteed invariants, and the rest as tests/needs-review. All reachable sites are fixed (plus ~12 more found during implementation); invariants are intentionally left.

## What now returns an error instead of panicking

- **Malformed sort declarations** → `TypeError::BadPresortArguments`: `(sort S (Vec))`, `(sort S (UnstableFn))`, `(sort S (Pair i64))`, etc. A real `span` is threaded through `Presort::make_sort`.
- **Partial primitives** (return no result instead of crashing): out-of-bounds/negative `vec-set`/`vec-remove`, empty `multiset-pick`, count overflow in `multiset` ops, and the numeric primitives `bigint <<`/`>>`, `bigrat`, `log`, `cbrt`, `log2`.
- **Command execution** → `Error`: negative `extract` variant counts, duplicate rule names, bad `input`/`print-function` files & column types, `output`/`extract` of deleted/subsumed values.
- **`unstable-fn`**: threaded `Result` through `prim`/`query`/`resolve_function_container_target_with_context` so unknown, non-literal, or mis-typed targets error instead of panic.
- **Parsing/desugar**: `(fail (include ...))`, `(fail (datatype*))`, subsuming a non-call rewrite, and multi-command `fail`.
- **Proofs**: new `ProveExistsError` variants; running `prove`/`prove-exists` without proofs enabled; graceful re-parse errors in the term-encoding chain.
- **CLI / scheduler / serialize**: CLI write/path failures log and exit cleanly; scheduler no-free-var and unknown-ruleset edge cases no longer panic; serializing a sort whose name contains `-` (e.g. `Math-Expr`) no longer panics and round-trips correctly.

## Intentionally left (genuine invariants)

Type-checker-guaranteed `get_val`/`base_values().unwrap` in primitives, post-typecheck resolution in `core.rs`, documented Rust-API preconditions (`Match::get_value`, empty-vars `query`), a wasm-only thread-count panic, and compile-time proc-macro guards.

## Breaking change

`EGraph::print_function` now takes its output sink as `Option<(File, PathBuf)>` plus a `Span`, so write failures return `Error::IoError` instead of panicking.

## Testing

- Added regression tests (duplicate rule name, negative `extract`, missing `input` file, scheduler edge cases, dashed-sort serialization).
- Verified 33 confirmed reproductions now error gracefully instead of crashing.
- `make test` passes (958 tests); `make nits` passes (clippy `-D warnings`, fmt, doc).


src/proofs/proof_encoding.rs
yihozhang, line 246: Is it true that this internally generated program must parse?

src/proofs/proof_encoding.rs
```diff
@@ -406,7 +406,7 @@ impl<'a> ProofInstrumentor<'a> {
     /// The view table stores child terms and their eclass.
     /// The view table is mutated using delete, but we never delete from term tables.
     /// We re-use the original name of the function for the term table.
-    fn term_and_view(&mut self, fdecl: &ResolvedFunctionDecl) -> Vec<Command> {
+    fn term_and_view(&mut self, fdecl: &ResolvedFunctionDecl) -> Result<Vec<Command>, Error> {
```
yihozhang: Is it true that this internally generated program must parse?

src/proofs/proof_encoding.rs
oflatt: It's supposed to or it's an internal error

src/proofs/proof_encoding.rs
```diff
@@ -1107,13 +1107,13 @@ impl<'a> ProofInstrumentor<'a> {
         ))
     }
 
-    fn instrument_schedule(&mut self, schedule: &ResolvedSchedule) -> Schedule {
+    fn instrument_schedule(&mut self, schedule: &ResolvedSchedule) -> Result<Schedule, Error> {
```
yihozhang: See above comments. I wonder how many of the results of the functions there can be safely "unwrapped"

src/proofs/proof_encoding_helpers.rs
```diff
@@ -142,7 +142,7 @@ impl ProofInstrumentor<'_> {
     }
 
     /// Header commands for term encoding, setting up rulesets.
-    pub(crate) fn term_header(&mut self) -> Vec<Command> {
+    pub(crate) fn term_header(&mut self) -> Result<Vec<Command>, Error> {
```
yihozhang: Can this ever err?

src/sort/bigrat.rs
```diff
@@ -75,7 +75,7 @@ impl BaseSort for BigRatSort {
             if a.is_one() {
                 Some(Q::new(BigRational::zero()))
             } else {
-                todo!("log of bigrat")
+                None
```
yihozhang: I think `todo!` here is fine, because `None` has a very different semantics

src/sort/bigrat.rs
```diff
@@ -96,7 +96,7 @@ impl BaseSort for BigRatSort {
             if a.is_one() {
                 Some(Q::new(BigRational::one()))
             } else {
-                todo!("cbrt of bigrat")
```
yihozhang, line 99: same here

saulshanabrook:
This issue (and the comment on it) has a few places where we panic instead or raising errors. I had them included in other PRs and took them out to make reviewing them simpler. If you feel like handling those cases here feel free! https://github.com/egraphs-good/egglog/issues/898

src/extract.rs
```diff
@@ -687,8 +687,9 @@ impl<C: Cost + Ord + Eq + Clone + Debug> Extractor<C> {
                 let find_root_variants = |row: egglog_bridge::FunctionRow| {
                     if !row.subsumed {
                         let target = &row.vals[output_idx];
-                        if *target == canonical_value {
-                            let cost = self.compute_cost_hyperedge(egraph, &row, func).unwrap();
```
yihozhang, line 635: Can this ever fail?

src/scheduler.rs
```diff
@@ -73,7 +73,14 @@ impl Matches {
 
     /// The number of matches in total.
     pub fn match_size(&self) -> usize {
-        self.matches.len() / self.vars.len()
+        // A rule whose actions reference no variables produces empty tuples, so
+        // `vars` is empty and `matches` carries no values. Guard the division to
+        // avoid dividing by zero in that degenerate case.
+        if self.vars.is_empty() {
```
yihozhang: Should it be 0 or 1? Does 0 mean the actions will not be applied at all, even if the query succeeds?

src/proofs/proof_encoding.rs
yihozhang: Agreed — this is an internal invariant. You were right that the generated program must parse; the only reason it could fail was a real bug: `Display for Literal` didn't escape `"`/`\` in string literals, so re-parsing instrumented text with an embedded string broke. Fixed that (strings now round-trip) and reverted `parse_program` and the whole term-encoding chain back to panicking on parse failure (`expect("internally generated ... must parse")`). Done in 5386910.

yihozhang: Yes — reverted to an internal-error panic. See the sibling thread: the underlying parse failure was a string-escaping bug in `Display for Literal`, now fixed so the invariant actually holds.

src/proofs/proof_encoding.rs
yihozhang: `term_and_view` reverted to `-> Vec<Command>`; it only returned `Result` because of the parse helpers, which are now back to panicking on the (internal) invariant.

src/proofs/proof_encoding.rs
yihozhang: All of them, as it turns out — the only error source in `instrument_schedule`/`rebuild`/`declare_sort`/`term_and_view`/`instrument_rule`/`term_encode_command` was the internal parse helpers. Reverted the entire chain (including `add_term_encoding`) back to non-`Result`, and made the parse helpers panic on failure. The `(fail ...)`-empty case I'd turned into a `DesugarError` is also unreachable (every term-encodable inner command pushes ≥1 command), so it's back to an `expect`.

src/proofs/proof_encoding_helpers.rs
yihozhang: No — reverted `term_header` to `-> Vec<Command>`. It only parses internally generated ruleset declarations, which must parse.

src/sort/bigrat.rs
yihozhang: Agreed, reverted to `todo!`. `None` would be actively wrong here: e.g. `cbrt 8/1 = 2` is rational but unhandled, and `None` would claim no cube root exists. `todo!` correctly signals unimplemented. (Also removed the corresponding no_panic assertion.)

src/sort/bigrat.rs
yihozhang: Same — reverted to `todo!`.

src/extract.rs
yihozhang: Yes, reachably. `compute_cost_hyperedge` returns `None` when a variant has a child e-class with no finite extraction (e.g. a purely cyclic child). The old `.unwrap()` was a latent panic; such a variant can never be part of a minimal extraction, so skipping it is correct and the target e-class still extracts via its costed variants. Kept the change and added a comment explaining why.

src/scheduler.rs
yihozhang: Good catch — 0 was wrong: it (plus the early-return I added in `instantiate`) meant a variable-free rule silently never applied its actions when scheduled, even though its query matched. The root issue is that such rules collect empty tuples, so the match count is unrecoverable. Fixed properly: collect a single unit marker per match for variable-free rules, track the stored tuple width explicitly, and insert the single `(unit)` row so the action fires. Added `test_no_free_vars_rule_applies_actions` to lock it in.

src/proofs/proof_extraction.rs
```diff
@@ -13,6 +14,18 @@ pub enum ProveExistsError {
     PrimitivesUnsupported,
     #[error("Could not find a proof due to query not matching (constructor {constructor}).")]
     QueryDidNotMatch { constructor: String },
+    #[error("prove/prove-exists requires proofs to be enabled (run with --proofs).")]
```
oflatt, line 16: I don't like this change because it's mixing errors that are expected (requiresConstructor, PrimitivesUnsupported) with errors that should never happen if the proof code is correct. I also think that we check if proofs are enabled before this point? But I'm not sure.

oflatt (changes requested):
just reviewed proof related stuff

src/proofs/proof_extraction.rs
yihozhang: Good call — this is the same issue as the proof-encoding threads, applied to `prove_exists`. Fixed in 7bdf046.

You were right on both counts:
- **Mixing expected and internal errors.** Of the variants the PR added, five (`ConstructorNotDeclared`, `NoProofRecorded`, `ProofTermExtractionFailed`, `NotProvableByRule`, and `ProofCheck`) only fire if the proof machinery itself is buggy — they were `panic!`/`expect` before and Iʼve reverted them to that, and dropped them from `ProveExistsError`. The enum is now just the genuinely user-facing cases: `RequiresConstructor`, `PrimitivesUnsupported`, `QueryDidNotMatch`, and `ProofsNotEnabled`.
- **Checking proofs-enabled.** There is no earlier check — the gate is the `proof_func_parent.get(output_sort)` lookup right here: a missing annotation means the proof infra was never set up for the sort, which is exactly "proofs not enabled". I kept that as `ProofsNotEnabled` (itʼs the one real reachable case — before this PR it `panic!`d, covered by the `prove_exists_without_proofs_errors` regression test). I tried hoisting it into an explicit `if !proofs_enabled` guard up front, but that breaks the desugar-proof-testing mode, which re-runs an already-lowered program with `proofs_enabled == false` even though the proof tables are present — so the annotation lookup is the correct precondition, not the flag.

yihozhang:
@saulshanabrook This PR is ready for review. It should have also fixed #898 .

saulshanabrook:
@coderabbit full review

saulshanabrook (approved):
This looks good to me! Thank you for doing all of this :)

oflatt (approved):
Nice! Thanks

PR #925, opened 2026-06-10
run-schedule size limit

noahhuck:
- Add an optional `size-limit` tag to the `run-schedule` command.
- For the duration of the schedule, keeps track of an upper bound on how many enodes are added. When the estimated size reaches the limit, stops the schedule. Estimated size may be less than the actual egraph size after running the schedule, since matches may produce duplicate enodes.
- Updates the estimate after each iter to the actual size of the egraph after rebuild.
- Any matches found during the last iter but not applied will be lost and can't be recovered by seminaive.

### DIFF
--- a/core-relations/src/free_join/execute.rs
+++ b/core-relations/src/free_join/execute.rs
@@ -3,7 +3,7 @@
 use std::{
     cmp, iter, mem,
     ops::Range,
-    sync::{Arc, OnceLock, RwLock, atomic::AtomicUsize},
+    sync::{Arc, OnceLock, RwLock, atomic::AtomicBool, atomic::AtomicUsize},
 };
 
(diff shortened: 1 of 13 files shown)


yihozhang:
This seems to be very specialized to Herbie's usage. Is it fine to keep this in a branch and not merge it into main?

saulshanabrook:
I think it would be good to have herbie be able to work off egglog main at some point. So if this technique isn't appropriate, then it would be nice to brainstorm alternate ways to support them that would fit better in main/experimental.

A size limit isn't that specialized is it? Doesn't egg have a size limit? It would be nice to more easily adapt egg based projects to egglog by being able to support those features even if they aren't as ergonomic in egglog.

PR #927, opened 2026-06-22
Proof support for containers 

oflatt:


src/proofs/proof_encoding_helpers.rs
```diff
@@ -534,9 +542,17 @@ pub(crate) fn command_supports_proof_encoding(
...
-        } => Err(ProofEncodingUnsupportedReason::SortWithPresort),
+        } => match presort.as_str() {
+            // Built-in container sorts participate in proof/term encoding;
+            // their contents are canonicalized via the container rebuild path
+            // (see `rebuilding_rules`), which recurses through nested
+            // containers. Other presorts (e.g. `UnstableFn`) and unknown custom
+            // presorts stay unsupported.
+            "Pair" | "Vec" | "Set" | "Map" | "MultiSet" => Ok(()),
+            _ => Err(ProofEncodingUnsupportedReason::SortWithPresort),
+        },
```
saulshanabrook: Can we just treat all containers as supported?

src/sort/set.rs
```diff
@@ -110,26 +122,118 @@ impl ContainerSort for SetSort {
     fn register_primitives(&self, eg: &mut EGraph) {
         let arc = self.clone().to_arcsort();
 
-        add_primitive!(eg, "set-empty" = {self.clone(): SetSort} |                      | -> @SetContainer (arc) { SetContainer {
+        // Proof term form of a set: `(set-of e0 e1 ...)`, matching
```
saulshanabrook: I need to understand the validator's better... Are they run on extracted terms? like from reconstruct termdag? I dont understand I think if they are supposed to take in any possible termdag of unevaluated primitive calls or only the canonical constructor.

Do they have to effectively serve as term only re-implementation of the container's behavior?

src/sort/map.rs
```diff
@@ -139,14 +140,47 @@ impl ContainerSort for MapSort {
...
+        // in canonical key order (like `set-of`/`vec-of`), matching
+        // `reconstruct_termdag`. The `map-empty`/`map-insert` validators
+        // normalize into that flat form too, so the checker evaluates
+        // user-written map terms to the same canonical term.
+        let map_empty_validator = |termdag: &mut TermDag, _args: &[TermId]| -> Option<TermId> {
+            Some(termdag.app("map-of".into(), vec![]))
+        };
+        let map_of_validator = |termdag: &mut TermDag, args: &[TermId]| -> Option<TermId> {
+            let raw = termdag.app("map-of".into(), args.to_vec());
+            Some(termdag.normalize_container_term(raw))
```
saulshanabrook: Like this seems pretty awkward and not viable... special casing all other constructors in map's definition?

+ what's the deal with accepting `map-insert` + `map-of` + `map-empty` but not other map functions?

src/termdag.rs
```diff
@@ -124,6 +124,108 @@ impl TermDag {
...
+                cs.dedup();
+                self.app("set-of".to_string(), cs)
+            }
+            "multiset-of" => {
+                let mut cs = args;
+                cs.sort_by(|a, b| self.ast_cmp(*a, *b));
+                self.app("multiset-of".to_string(), cs)
+            }
+            "map-of" | "map-empty" | "map-insert" => {
+                // Accept either the flat `map-of` form or a `map-insert`/
```
saulshanabrook: again this seems to duplicate the map code... and also is some seemingly arbitrary special cases?

src/termdag.rs
```diff
@@ -124,6 +124,108 @@ impl TermDag {
...
+    /// set or multiset) into a stable, reproducible term order.
+    pub fn ast_cmp(&self, a: TermId, b: TermId) -> std::cmp::Ordering {
+        use std::cmp::Ordering;
+        match (self.get(a), self.get(b)) {
+            (Term::Lit(x), Term::Lit(y)) => x.cmp(y),
+            (Term::Lit(_), _) => Ordering::Less,
+            (_, Term::Lit(_)) => Ordering::Greater,
+            (Term::Var(x), Term::Var(y)) => x.cmp(y),
+            (Term::Var(_), _) => Ordering::Less,
+            (_, Term::Var(_)) => Ordering::Greater,
```
saulshanabrook: It's just for Term::App that we wanna override cmp right? For the rest can we just call `l.cmp(r)`?

src/ast/mod.rs
```diff
@@ -557,6 +568,12 @@ where
         /// The name of the proof function for this sort.
         /// Set by proof desugaring to record where proofs are stored for this sort.
         proof_func: Option<String>,
+        /// For container sorts under the term/proof encoding: a serialized spec
+        /// (built by the encoder) used to register the container's rebuild
+        /// primitives. Carried as the `:internal-container-rebuild` annotation
```
saulshanabrook: I am sorry I am probably just missing it, but where is `:internal-container-rebuild` actually written to a sort? It's something the proof encoding should write?

src/sort/map.rs
oflatt: @saulshanabrook said to resolve this one, oops

core-relations/src/containers/mod.rs
```diff
@@ -737,3 +760,56 @@ fn incremental_rebuild(uf_size: usize, table_size: usize, parallel: bool) -> boo
...
+/// closure. Used by [`ContainerValues::rebuild_val_with`] to rebuild a single
+/// container against an explicit value mapping (rather than a backend
+/// union-find). Only the value-level methods are reachable from
+/// [`ContainerValue::rebuild_contents`]; the table-level methods are never
+/// called in that path.
+struct ClosureRebuilder<'a> {
+    remap: &'a (dyn Fn(Value) -> Value + Send + Sync),
+}
+
+impl Rebuilder for ClosureRebuilder<'_> {
```
oflatt: from talking with Saul: is this used by both users and the backend? Can users call rebuild_buf? Seems bad if so

src/ast/mod.rs
```diff
@@ -61,6 +61,13 @@ where
...
+        /// so the (otherwise programmatically registered) rebuild primitives
+        /// survive the desugar/serialize round-trip — they are re-registered
+        /// when the Sort is typechecked. See `proofs::proof_encoding`.
+        container_rebuild: Option<String>,
```
oflatt: where did we set this in the code? didn't see it in proof instrumentation

src/typechecking.rs
```diff
@@ -423,12 +424,22 @@ impl EGraph {
                 if !unionable {
                     self.type_info.non_unionable_sorts.insert(name.clone());
                 }
+                // A container sort under the term/proof encoding carries a spec
+                // for its rebuild primitives; register them here so they are
+                // available both during encoding and when the desugared program
+                // is re-parsed.
+                if let Some(spec) = container_rebuild {
+                    crate::proofs::proof_encoding::register_container_rebuild_from_spec(
```
oflatt: don't use absolute paths- we should add a skill for this

src/termdag.rs
```diff
@@ -124,6 +124,108 @@ impl TermDag {
...
+                cs.sort_by(|a, b| self.ast_cmp(*a, *b));
+                cs.dedup();
+                self.app("set-of".to_string(), cs)
+            }
+            "multiset-of" => {
+                let mut cs = args;
+                cs.sort_by(|a, b| self.ast_cmp(*a, *b));
+                self.app("multiset-of".to_string(), cs)
+            }
+            "map-of" | "map-empty" | "map-insert" => {
```
oflatt: this whole helper is bad because it is hard-coded to the existing containers, users can write their own. perhaps we should have a small generic helper for the sorting that people can optionally call after stuff like set duduplication

src/proofs/proof_encoding_helpers.rs
oflatt: fixed

src/sort/set.rs
oflatt: Yeah they work on termdag extracted terms. Yep!

src/termdag.rs
oflatt: I looked into this, implementing `Ord` for `Term` would allow us to do that, but then users might accidently use `Ord` instead of using `ast_cmp`. I decided to stick with this.

src/sort/map.rs
```diff
@@ -161,16 +282,117 @@ impl ContainerSort for MapSort {
...
-        }
+        // Flat `(map-of k0 v0 k1 v1 ...)` in canonical key order, so proof
+        // checking can reproduce it from terms alone (and the rebuild proof's
+        // Congr indices are flat, like `set-of`/`vec-of`).
+        let raw = termdag.app("map-of".into(), element_terms);
+        normalize_map_term(termdag, raw).unwrap_or(raw)
+    }
 
-        term
+    fn container_term_normalizer(&self) -> Option<(String, PrimitiveValidator)> {
```
saulshanabrook: rebuild_container_normalizer

src/sort/map.rs
```diff
@@ -36,6 +36,65 @@ impl ContainerValue for MapContainer {
     }
 }
 
+/// Canonicalize a `map-of`/`map-empty`/`map-insert` term to a flat `(map-of k0
+/// v0 ...)` in sorted key order, with last-write-wins on duplicate keys.
+/// Returns `None` for a malformed map term (an odd-arity `map-of`, a
+/// `map-insert` not of arity 3, or a spine not bottoming out in
+/// `map-empty`/`map-of`), so callers leave such terms unchanged.
+fn normalize_map_term(termdag: &mut TermDag, term_id: TermId) -> Option<TermId> {
```
saulshanabrook: instead go back and forth with rust maps for each validator

src/sort/mod.rs
```diff
@@ -140,6 +149,25 @@ pub trait Sort: Any + Send + Sync + Debug {
         let _termdag = termdag;
         todo!("reconstruct_termdag_leaf: {}", self.name());
     }
+
+    /// Canonicalize a container value for the term encoding by remapping each
+    /// contained value through `leaders` (an old-value -> union-find-leader
+    /// map), returning the new interned container value. Only container sorts
+    /// implement this; it is the value-level half of the container rebuild
+    /// performed by the term encoding's rebuild rules.
+    fn rebuild_container_with_leaders(
```
saulshanabrook: inline or move out of sort

src/typechecking.rs
```diff
@@ -1085,6 +1127,17 @@ impl TypeInfo {
         self.primitives.get(sym).map(Vec::as_slice)
     }
 
+    /// Map from a container's constructor head to the validator that canonicalizes
+    /// its term form, collected from the container sorts that support proofs (see
+    /// [`Sort::container_term_normalizer`]). Proof checking uses this to normalize
+    /// a container term by its head.
+    pub(crate) fn container_term_normalizers(&self) -> HashMap<String, PrimitiveValidator> {
```
saulshanabrook: Inline?

src/sort/map.rs
```diff
@@ -36,65 +36,59 @@ impl ContainerValue for MapContainer {
...
+    termdag: &'a TermDag,
     term_id: TermId,
-    out: &mut Vec<(TermId, TermId)>,
-) -> Option<()> {
+) -> Option<BTreeMap<OrdTerm<'a>, TermId>> {
     match termdag.get(term_id) {
         Term::App(head, args) if head == "map-insert" => {
             let [inner, k, v] = args.as_slice() else {
                 return None;
             };
```
saulshanabrook: I think we dont need this case since map-insert is handled below in that validator?

saulshanabrook:
Added one last comment and seeing if the AI has anything else useful, since a lot of the term encoding code I didn't look through super carefully: @coderabbitai full review

src/sort/map.rs
oflatt: yeah good catch

saulshanabrook (approved):

PR #928, opened 2026-06-24
Avoid redundant proof merge rows for eager current values

saulshanabrook:
## Summary

- Add a hidden current-value function for proof-encoded custom functions that have a merge function.
- Keep that current-value function updated from ordinary proof-view writes, using the original eager backend merge.
- Add a cleanup rule that deletes stale proof-view candidates when the eager current value already has a proof witness, while leaving the existing merge-proof path as the fallback.
- Add a focused regression test for a selector-style `:merge old` function.

## Why

This is needed for eggcc proof export to finish in reasonable time. The old proof encoding could keep every candidate proof-view row for a merged function and then run `@merge_rule` over every pair. For selector/eager merges, the backend already knows the current value, so those pairwise proof merges are redundant when the selected row already has a proof witness.

## Slow Example

A generated file with this shape was slow before this change and is faster now:

```lisp
(sort Key)
(constructor K () Key)
(function Best (Key) i64 :merge old)
(relation Candidate (Key i64))
(rule ((Candidate k v))
      ((set (Best k) v)))

;; Repeat many same-key candidates, e.g. 3000 rows:
(Candidate (K) 0)
(Candidate (K) 1)
;; ...
(Candidate (K) 2999)

(run 1)
(check (= (Best (K)) 0))
```

Run with proof mode, for example:

```sh
egglog --proofs --save-report selector-old-3000-report.json selector-old-3000.egg
```

On the same 3000-candidate file:

- before: `@merge_rule = 4,498,500`, `@merge_cleanup = 4,498,500`, wall time about `7.04s`
- after: `@merge_rule = 0`, `@merge_current_cleanup = 2,999`, wall time about `4.61s`

The wall-time improvement is partially hidden by parsing and report overhead, but the quadratic proof-merge work is eliminated for this selector case.

## Validation

- `cargo test --release --test proof_selector_merge`
- `git diff --check`


saulshanabrook:
@coderabbitai full review

oflatt:
I'm working on https://github.com/egraphs-good/egglog/pull/933, which hopefully will replace this. We'll keep this open as an option for now

oflatt (approved):
makes sense to me! nice fix

PR #929, opened 2026-06-24
[codex] Fix proof-mode macro type info

saulshanabrook:
## Summary

- make proof-mode command macro expansion read from the original pre-instrumentation type environment
- keep the type environment borrowed rather than cloning `TypeInfo` for every command
- add a regression test showing proof-mode macros see the original function arity

## Why

Proof mode rewrites functions during term/proof instrumentation. Command macros run before each command is processed and need the user-facing type information, not the instrumented proof helper arities.

The first implementation cloned `TypeInfo` for each command, and CodSpeed flagged broad slowdowns. This update keeps the same behavior but passes a borrowed `TypeInfo` reference.

## Validation

- cargo fmt --check
- cargo test --test proof_mode_regression --release
- cargo test -p egglog test_command_macros --release
- cargo test -p egglog proof_tests --release
- git diff --check

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2022,12 +2022,12 @@ impl EGraph {
                 .proof_state
                 .original_typechecking
                 .as_ref()
-                .map(|egraph| egraph.type_info.clone())
-                .unwrap_or_else(|| self.type_info.clone());
+                .map(|egraph| &egraph.type_info)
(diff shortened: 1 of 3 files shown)


saulshanabrook:
@coderabbitai full review

oflatt (approved):
lgtm

PR #930, opened 2026-06-24
[codex] Support eq-sort primitive fact proofs

saulshanabrook:
## Summary

- allow proof/term encoding for primitive fact expressions that return equality sorts
- read the generated term proof on the rule RHS, reusing the existing action-side lookup path instead of adding another body join
- reject user `:naive` rules that use eq-sort primitive results in body facts, since that proof lookup invariant only holds for seminaive-safe query evaluation
- add proof-mode regressions for eq-sort, eq-container-sort, and the unsupported `:naive` shape

## Why

This is the standalone primitive-output piece needed before the container/eager-merge work can rely on primitives such as pair accessors in proof-mode facts. It also avoids putting proof-only lookups into the rule body, which was the source of the proof-mode join blow-up seen in the eggcc 2mm experiment.

For `:naive` rules, the body can use read-capable primitive contexts, so we do not assume an eq-sort primitive result has an existing stored term proof. Those commands now get a targeted unsupported-proof error instead.

## Validation

- `cargo fmt --check`
- `cargo test proof_support_rejects_naive_eq_sort_primitive_results_in_facts --lib --release`
- `cargo test proof_mode_allows_eq_sort_primitive_results_in_facts --lib --release`
- `cargo test proof_mode_allows_eq_container_primitive_results_in_facts --lib --release`
- `cargo check --lib --release`
- `cargo test -p egglog proof_tests --release`
- `git diff --check`


saulshanabrook:
@coderabbitai full review

oflatt (changes requested):
Interesting, the argument for why this works is subtle. I think it basically goes like this:

- a primitive that returns an eqsort in a query is read-only
- therefore it must give back a term which already exists
- and all terms that exist have proofs
- so we will be able to look up a proof for it

For :naive rules I think this chain of reasoning doesn't work so maybe we should write a test case for that case and throw an error

saulshanabrook:
Thanks @oflatt that's a valid concern! I catch this case now and trigger an unsupported error.

oflatt (changes requested):
Nice, though I think we shouldn't add tests cases to lib.rs files. See my issue in lib.rs getting huge

saulshanabrook:
@oflatt fixed!

oflatt (approved):
thanks!

PR #934, opened 2026-06-24
Expose extraction support APIs

saulshanabrook:
# Extraction cost model and support APIs

This is the core half of the greedy-DAG extraction split. It does not add
`:extractor greedy-dag` to core egglog; that syntax and heuristic live in the
downstream experimental PR:

- <https://github.com/egraphs-good/egglog-experimental/pull/55>

Related context:

- Python issue motivating shared DAG extraction: <https://github.com/egraphs-good/egglog-python/issues/387>
- Cost-model flexibility discussion: <https://github.com/egraphs-good/egglog/issues/294>

## Changes

- Separates the requirements of tree and DAG extraction:
  - `Cost` is the ordered, cloneable value needed to rank tree candidates.
  - `TreeCostModel` computes enode/container annotations and folds them with
    selected child costs. Its cost type does not need an addition operation.
  - `MonoidCost` adds a lawful identity and associative, commutative,
    monotone combination operation.
  - `DagCostModel` returns intrinsic enode, container, and base-value costs
    suitable for once-paid DAG accounting.
  - `TreeCostModelFromDag` explicitly adapts a DAG model to tree extraction by
    combining each intrinsic cost with its selected child costs.
- Replaces the old additive model with `AdditiveCostModel { node_cost }`. Its
  default `u64` configuration preserves unit-node costs and constructor
  `:cost` declarations.
- Renames and re-exposes the reusable tree extractor as
  `TreeExtractor<'g, C>`. Callers can prepare costs for selected root sorts and
  repeatedly extract individual values while its borrow keeps the e-graph
  immutable.
- Adds batch `EGraph::extract_best_with_cost_model` and
  `EGraph::extract_variants_with_cost_model` methods with named result structs
  and one shared `TermDag`.
- Keeps the singular default-cost convenience API as
  `EGraph::extract_value(&sort, value)`.
- Represents an unextractable root as `None` in `ExtractedTerms`; strict
  command paths convert that case to `ExtractError` where required.
- Routes builtin extraction, output, and function-printing paths through the
  refactored tree extractor. Batched best extraction reuses one reconstruction
  memo across roots.
- Exposes normal-mode support needed by downstream custom extractors:
  container traversal/reconstruction and `Function::is_unextractable`.
  Downstream combines these with the already-landed
  `Read::enodes_for_eclass` lookup.

## Public Interfaces

Extraction:

- `Cost`, `MonoidCost`
- `TreeCostModel<C>`, `DagCostModel<C>`, `TreeCostModelFromDag<M>`
- `AdditiveCostModel`, `DefaultCost`
- `ExtractedTerm<C>`, `ExtractedTerms<C>`, `ExtractedTermVariants<C>`
- `TreeExtractor<'g, C>`
  - `compute_costs_from_rootsorts(rootsorts, egraph, cost_model)`
  - `extract_best_with_sort(termdag, value, sort)`
  - `extract_variants_with_sort(termdag, value, nvariants, sort)`

`EGraph`:

- `extract_best_with_cost_model(roots, cost_model)`
- `extract_variants_with_cost_model(roots, nvariants, cost_model)`
- `extract_value(sort, value)`
- `container_inner_values(sort, value)`
- `reconstruct_base_value(sort, value, termdag)`
- `reconstruct_container_value(sort, value, termdag, element_terms)`

`Function`:

- `Function::is_unextractable()`

## Review Notes

The normal `(extract ...)` command remains tree extraction. Proof extraction
is root-directed and separate after
<https://github.com/egraphs-good/egglog/pull/941>; this PR does not modify that
path.

`MonoidCost` laws are semantic contracts that Rust cannot enforce. Built-in
implementations are limited to unsigned saturating integers and exact
`BigInt`/`BigRational` addition; arbitrary tree costs need only implement
`Cost`.

## Validation

- Full all-feature test suite
- `cargo clippy --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS='-D warnings' cargo doc --no-deps --all-features`
- `git diff --check`
- Taylor 51 tree extraction, 30 release runs:
  - pre-PR: `691.2 ms +/- 21.5 ms`
  - current: `680.3 ms +/- 8.9 ms`
  - mean change `-1.59%`, 95% CI `[-3.10%, -0.04%]`


saulshanabrook:
@coderabbitai full review

saulshanabrook:
@coderabbitai full review

saulshanabrook:
Things to look into from talking to @oflatt:

* verify that cycles are not extracted with a small test case, if you combine two extractions that it gives you a cycle?
* Verify that this works on all examples, that the extraction after output can be turned into an s-expr and then checked against the e-graph to make sure its equal, also do for normal extraction
* look in again if it could be moved to experimental
* change producer language to parent if appropriate?

saulshanabrook:
@coderabbitai full review

oflatt (approved):
Looks good to me (PR descriptions very out of date)

oflatt:
One thing I'd love is to have extraction work inside a ReadState as well as on the EGraph. But not this PR's problem.

core-relations/src/action/mod.rs
```diff
@@ -483,6 +485,76 @@ impl<'a> ExecutionState<'a> {
...
+            .split_fast_slow(std::slice::from_ref(&constraint));
+
+        slow.retain(|c| {
+            let (col, val) = match c {
+                Constraint::EqConst { col, val } => (*col, *val),
+                Constraint::Eq { .. }
+                | Constraint::LtConst { .. }
+                | Constraint::GtConst { .. }
+                | Constraint::LeConst { .. }
+                | Constraint::GeConst { .. } => return true,
```
yihozhang: Should they be impossible cases?

src/extract.rs
```diff
@@ -103,48 +111,80 @@ pub type DefaultCost = u64;
...
-impl CostModel<DefaultCost> for TreeAdditiveCostModel {
-    fn fold(
+impl BaseCostModel<DefaultCost> for TreeAdditiveCostModel {
+    fn base_value_cost(&self, _egraph: &EGraph, _sort: &ArcSort, _value: Value) -> DefaultCost {
+        1
+    }
+}
+
+impl TreeCostModel<DefaultCost> for TreeAdditiveCostModel {
+    fn total_enode_cost(
```
yihozhang, line 121: I remember we discussed this and decided to keep the existing enode_cost? The original one is also more natural to me.

src/extract.rs
```diff
@@ -289,18 +329,22 @@ impl<C: Cost + Ord + Eq + Clone + Debug> Extractor<C> {
...
             children: &row.vals[..output_idx],
             eclass: row.vals[output_idx],
             subsumed: row.subsumed,
         };
-        Some(self.cost_model.fold(
-            head_name,
-            &ch_costs,
-            self.cost_model.enode_cost(egraph, func, &enode),
-        ))
+        let cost_func = func
```
yihozhang, line 451: Could you explain what this does to me as someone who hasn't caught up with the proof development?

src/extract.rs
```diff
@@ -103,48 +111,80 @@ pub type DefaultCost = u64;
...
+    pub variants: Vec<Vec<ExtractedTerm<C>>>,
+}
+
+/// The default, Bellman-Ford like extractor. This extractor is optimal for [`TreeCostModel`].
 ///
 /// Note that this assumes optimal substructure in the cost model, that is, a lower-cost
 /// subterm should always lead to a non-worse superterm, to guarantee the extracted term
 /// being optimal under the given cost model.
 /// If this is not followed, the extractor may panic on reconstruction
-pub struct Extractor<C: Cost + Ord + Eq + Clone + Debug> {
```
yihozhang, line 127: Why is `Extractor` no longer public?

src/extract.rs
```diff
@@ -762,6 +755,49 @@ impl Function {
 }
 
 impl EGraph {
+    /// Extract the best tree term for each requested `(sort, value)` root.
+    ///
+    /// This is the normal user extraction path: it respects `:unextractable`
+    /// and hidden internal functions.
+    pub fn extract_best<C: Cost, M: TreeCostModel<C> + 'static>(
```
yihozhang: They seem to overlap with extract_value and extract_value_with_cost_model. Do we need all of them?

src/extract.rs
```diff
@@ -37,36 +45,40 @@ pub trait CostModel<C: Cost> {
...
+pub trait Cost: CommutativeMonoid + Ord + Eq + Debug {}
+
+impl<T: CommutativeMonoid + Ord + Eq + Debug> Cost for T {}
+
 macro_rules! cost_impl_int {
     ($($cost:ty),*) => {$(
-        impl Cost for $cost {
+        impl CommutativeMonoid for $cost {
             fn identity() -> Self { 0 }
-            fn unit()     -> Self { 1 }
```
yihozhang, line 69: Why do we no longer have `unit()` for a cost? This provides a default for e.g., base values.

src/extract.rs
```diff
@@ -37,36 +45,40 @@ pub trait CostModel<C: Cost> {
...
-    fn base_value_cost(&self, egraph: &EGraph, sort: &ArcSort, value: Value) -> C {
-        let _egraph = egraph;
-        let _sort = sort;
-        let _value = value;
-        C::unit()
-    }
 }
 
-/// Requirements for a type to be usable as a cost by a [`CostModel`].
-pub trait Cost {
```
yihozhang, line 53: Do we have to change the name?

yihozhang (commented):
Thank you!

yihozhang:
Small nit: it seems some relatively big files are checked into the git history. Maybe worth doing a squash merge at the end.

src/extract.rs
saulshanabrook: The existing enode_cost has the problem where its only ever used to produce total cost with the fold, so it seemed like odd to keep it. Like the functions were ever only called changed together, with enode_cost feeding into the total call.

So that it returned a "Cost" type was also a bit weird, really it could return anything that the fold took in, like I had to return odd stuff from it from Python to basically just pass data between them and do all the computation in total_enode_cost.

src/extract.rs
saulshanabrook: We dont need it. Every use case of it just extracted a few particular nodes, so I cleaned up the interface to just expose that to keep the details more private.

src/extract.rs
saulshanabrook: Yeah `extact_value` and `extract_value_with_cost_model` were previous helpers, I can remove them.

src/extract.rs
saulshanabrook: You dont need it. It was only used in the default trait implementation and not used at all in the actual extractor, it was just used in the deafult trait implementation, but that wasn't really needed. so it seems cleaner to remove it. Like there is no reason for a cost to require it.

src/extract.rs
saulshanabrook: I can change it back

src/extract.rs
yihozhang: It provides a default head cost for functions and values. For example, in your implementation of TreeAdditiveCostModel and CustomCostModel, you use the magic number `1`. It does not need to be magic if we have `unit()`. Alternatively, consider how you would make a generic `AstSizeCostModel<C> where C: Cost` if without `unit`.

src/extract.rs
yihozhang: Did you pass odd stuff because there was no DAG cost model and extractor, or do you still need to do it even with the new DAG extractor?

If we use enode_cost here, TreeCostModel exposes a superset of cost functions of DAG cost model's, which is nice and matches our (my) intuition. My mental model of the tree cost model is that the term is annotated with cost tags and the cost of a term is just a fold over these cost tags. So, for example, the proposed cost model forbids you from doing it if you want to visualize an e-graph with per-node cost annotation, or serialize an e-graph with a custom cost model.

src/extract.rs
yihozhang: I thought part of the point of the extractor interface is to provide a zoo of different extractor implementations (like extraction-gym for egglog). It would be weird if someone implemented a faster extractor but found they couldn't get a handle on the plain Extractor for comparison. I'm also in favor of minimizing interface changes unless it's a strict improvement.

Another potential use case of the extractor is if the user wants to interactively extract a series of terms, without building the Extractor every time

saulshanabrook: We don't have an extractor interface. We just have a tree extractor and a dag extractor, and they expose different methods on the e-graph. Without anything that calls this interface, I am not sure what it would be for?

You can of course compare against the main extractor, you just call the methods on the e-graph.



> Another potential use case of the extractor is if the user wants to interactively extract a series of terms, without building the Extractor every time

Yeah I just thought since I hadn't seen any use cases in experimental and main that exercised this pattern, we could remove it. That also frees us up to change how the extractor performs in the future. Like now it does this search at a sort level of the e-graph to trim, and does all extractions for that. But we could change it to do a more precise traversal of the e-graph and only build extractions for those expressions, instead of all pieces. By keeping the interface smaller then that doesn't have to be a breaking change.

src/extract.rs
oflatt: In proof mode, view tables like AddView are used for querying and extraction, but the actual name of the function is still Add. The `term_constructor` attribute tells you the original name. Does that help?

It's annoying, but in proof mode everywhere we expose the API we need to keep in mind to use the "term_cosntructor" names

oflatt: Might be a better way to clean this up! Some sort of wrapper internally so we get the right names and right tables.

src/extract.rs
saulshanabrook: As I understand it, we should have the cost trait reflect what the underlying extraction algorithm needs. The fact that you can split this up into a fold and an e-node cost is an implementation choice of a particular cost model, not a requirement of the extractor.

So for example, you could have a cost model for the tree extractor which says "on this particular e-node multiply the child costs by ten and on this other one multiply the child costs by twenty. In the other interface, you would have to smuggle the e-node ID through the cost output of enode_cost and then interpret it in fold.

> If we use enode_cost here, TreeCostModel exposes a superset of cost functions of DAG cost model's, which is nice and matches our (my) intuition.

We can already use any dag cost model as a tree cost model, do you mean something else?

> So, for example, the proposed cost model forbids you from doing it if you want to visualize an e-graph with per-node cost annotation, or serialize an e-graph with a custom cost model.

This is already true with `fold`, which is more general than what is allowed in the visualizer or in serialization. Both of those in fact *do not* currently support the `set-cost` even, so if we want to somehow make those features aware of this, we would have to change things.

src/extract.rs
yihozhang: We will want an extractor interface, right? There are many DAG extractor implementations (ILP, branch and bound, ASP, etc.), and there's the Dijkstra algorithm for tree extraction. Just like we have an interface for extensible schedulers, we should do the same for extractors. I'm all in for standardizing the interfaces around DAG and tree extractors, just like what we did in extraction-gym.

Not in experimental and main does not mean users won't have this pattern? A user could import egglog, build an e-graph, and interactively extract terms from the Rust side.

src/extract.rs
yihozhang: To me this is not an implementation choice, but the definition of the tree cost model.

The idea is that enode_cost returns all e-graph specific cost information, and fold is an e-graph--agnostic way to combine such information. This gives a clear description of the tree extraction problem, which can be specified as an e-graph with per-node cost information, and a generic `fold` function. I believe you cannot do this with `total_node_cost`. 

"On this particular e-node multiply the child costs by ten and on this other one multiply the child costs by twenty" is a combine function we can define, while I think smuggling e-node ID is a bad idea. If you want to return odd stuff from enode_cost (e.g., the constant folded result of the children as a loop count estimate), the user can define the cost as a pair (ActualCost, Option<Estimate>) and ignore the second part for comparison, or better, we can tweak the egglog interface to support that.

> We can already use any dag cost model as a tree cost model, do you mean something else?

I meant you could convert a tree cost model into a DAG one by disregarding combine.

> Both of those in fact do not currently support the set-cost even

But we can implement this, right? Again, a serialized e-graph with node costs + an e-graph--agnostic fold function gives a complete specification of the extraction problem. But committing to total_enode_cost would eliminate this possibility.

Maybe another way to think about this is that TreeAdditiveCostModel and SetCostModel are essentially the same cost model, besides how they obtain the per-node cost, but the new interface loses this connection.

src/extract.rs
saulshanabrook: Sure but would that interface be? We dont define it in this PR and I would opt to defer it.

If we define that interface, then we want a way to actually use that interface in main somehow, like being able to switch extractors.

> A user could import egglog, build an e-graph, and interactively extract terms from the Rust side.

Yeah you can do that today with the methods on the e-graph.

src/extract.rs
saulshanabrook: Yeah it provides a default cost, but its not clear to me its worth having there, because if you define your own cost model, its not needed. I.e. cost should be about what the extraction/cost model needs, and its not required, its optional basically. 

Like in the current implementation, this is how we implement the cost model for `u64`:


```rust
impl CostModel<DefaultCost> for TreeAdditiveCostModel {
    fn fold(
        &self,
        _head: &str,
        children_cost: &[DefaultCost],
        head_cost: DefaultCost,
    ) -> DefaultCost {
        children_cost.iter().fold(head_cost, |s, c| s.combine(c))
    }

    fn enode_cost(&self, egraph: &EGraph, func: &Function, _enode: &Enode<'_>) -> DefaultCost {
        func.extraction_head_cost(egraph)
    }
}
```

It isn't clear to me how this is better than doing:

```rust
impl CostModel<DefaultCost> for TreeAdditiveCostModel {
    fn fold(
        &self,
        _head: &str,
        children_cost: &[DefaultCost],
        head_cost: DefaultCost,
    ) -> DefaultCost {
        children_cost.iter().fold(head_cost, |s, c| s.combine(c))
    }

    fn enode_cost(&self, egraph: &EGraph, func: &Function, _enode: &Enode<'_>) -> DefaultCost {
        func.extraction_head_cost(egraph)
    }

    fn base_value_cost(&self, egraph: &EGraph, sort: &ArcSort, value: Value) -> C {
        1
    }
}
```

saulshanabrook: I updated it to make `TreeAdditiveCostModel` generic, so that you can parameterize it with different unit costs for different cost types.

src/extract.rs
saulshanabrook: Basically I think there are two related points here:

1. There is a question about having a re-usable extractor interface. I like this idea, but I didn't implement it in this PR. There was no previous extractor trait, just a concrete implementation of extractor. That extractor happened to work in a way that traverses first the e-graph at a sort level first, prepares all of those items, then allows extractions. It's not clear that we would want to use this as "the" interface, because for example, the DAG extractor does not traverse the whole e-graph by sort. It does a depth first traversal based on the actual nodes you want to extract. I would prefer not to standardize an extractor interface in this PR and defer that to later work. If we do have an interface, we should have a way to use that in the e-graph in some way.
2. There is the loss of functionality with this changes around the default extractor. Previously, you could do the work of creating costs for a subset of the e-graph based on a list of sorts and all their possible children. I have shrunk the API to simply providing a list of expressions and getting their extractions out. I didn't see any uses for this more flexible API. I decided to shrink the exposed syntax to make it easier in the future if we want to change out default tree based extractor to use a different design that might not require traversing the e-graph by sort first. By keeping this public API it keeps us locked into that syntax, so further performance improvements would need to conform to it as well. If there are actual use cases today for this interface that would be helpful to understand why it's important to preserve.

src/extract.rs
saulshanabrook: Removed

src/extract.rs
saulshanabrook: Reverted, I think its clearer now

core-relations/src/action/mod.rs
saulshanabrook: Fixed, thank you!

src/extract.rs
saulshanabrook: Just talked to Yihong, will keep public and change to TreeExtractor

saulshanabrook:
@CodeRabbit full review

src/extract.rs
```diff
@@ -1,150 +1,314 @@
...
-use ordered_float::OrderedFloat;
-cost_impl_num!(f32, f64, OrderedFloat<f32>, OrderedFloat<f64>);
 
+combinable_exact_cost!(num::BigInt, num::BigRational);
+
+/// Computes the cost of primitive values, which have no selected children.
+///
+/// Repeated calls with the same arguments during one extraction must return
+/// equal costs.
+pub trait BaseCostModel<C: Cost> {
```
yihozhang: Can we collapse this into the other cost models? A cost model defines a function from a term to a cost, and base cost model alone does not define such a function. My understanding is this is for implementation reuse, but this does not seem to save much code I feel.

src/extract.rs
```diff
@@ -1,150 +1,314 @@
...
+    /// Computes the cost of a non-container primitive value.
+    fn base_value_cost(&self, egraph: &EGraph, sort: &ArcSort, value: Value) -> C;
+}
+
+/// Computes intrinsic costs that exclude selected child and element costs.
+///
+/// Repeated calls with equivalent arguments during one extraction must return
+/// equal costs. Greedy DAG extraction may recompute marginal costs while
+/// reconciling producer choices.
+pub trait MarginalCostModel<C: CombinableCost>: BaseCostModel<C> {
```
yihozhang: Similarly, I think we should bake in to this cost model the idea that we combine node costs with `+` and "DAG" cost. Otherwise, this trait alone does not tell us how to define a cost function. As a corollary of this, `FoldCostModel` should not require `MarginalCostModel`

yihozhang: If we do that, I suggest we call `MarginalCostModel` "DagCostModel" and `FoldCostModel` "TreeCostModel", to align with the literature. If you want to distinguish them from TotalCostModel, maybe we can call them something like "TreeNodeCostModel" or "TreeMarginalCostModel"?

src/extract.rs
```diff
@@ -1,150 +1,314 @@
...
 
-/// Requirements for a type to be usable as a cost by a [`CostModel`].
-pub trait Cost {
-    /// An identity element, usually zero.
+/// An extraction cost with an identity and a combination operation.
+///
+/// Implementations must make `identity` a two-sided identity and `combine`
+/// associative, commutative, deterministic, non-panicking, and monotone with
+/// respect to [`Cost`]'s ordering. Rust cannot enforce these laws.
+pub trait CombinableCost: Cost {
```
yihozhang: alt name idea: MonoidCost

src/extract.rs
```diff
@@ -1,150 +1,314 @@
...
+    ) -> C;
+}
+
+/// Derives total costs by folding marginal costs with selected child costs.
+///
+/// Implement this trait to accept the default combine-based folds or override
+/// either fold for context-sensitive tree extraction. This opt-in is separate
+/// from [`MarginalCostModel`] so custom models can choose between this derived
+/// implementation and a direct [`TotalCostModel`] implementation.
+pub trait FoldCostModel<C: CombinableCost>: MarginalCostModel<C> {
```
yihozhang: If you agree with the idea above, here we can duplicate the methods from MarginalCostModel.

src/extract.rs
```diff
@@ -1,150 +1,314 @@
...
+#[derive(Clone, Debug)]
+pub struct ExtractedTermVariants<C> {
+    /// Shared term storage for every extracted variant.
+    pub termdag: TermDag,
+    /// Outer vector follows requested-root order; each inner vector contains that root's variants.
+    pub variants: Vec<Vec<ExtractedTerm<C>>>,
 }
 
-/// The default, Bellman-Ford like extractor. This extractor is optimal for [`CostModel`].
+/// The default Bellman-Ford-like extractor. This extractor is optimal for a
```
yihozhang: There's some repetition in this doc currently

src/extract.rs
```diff
@@ -1,150 +1,314 @@
...
 /// If this is not followed, the extractor may panic on reconstruction
-pub struct Extractor<C: Cost + Ord + Eq + Clone + Debug> {
-    rootsorts: Vec<ArcSort>,
+///
+/// Cost models should usually make a term no cheaper than its subterms; that
+/// rules out cycles in common extraction workloads. More specialized models may
+/// assign lower costs to larger terms, but then the model is responsible for
+/// avoiding negative-cost cycles and cyclic extracted terms.
+pub struct TreeExtractor<'g, C: Cost> {
+    egraph: &'g EGraph,
```
yihozhang, line 274: Why do we need to store the EGraph?

src/extract.rs
```diff
@@ -1,150 +1,314 @@
...
-    /// The later extractions only reuses saved results.
-    /// This means a new extractor must be created if the egraph changes.
-    /// Holding a reference to the egraph would enforce this but prevents the extractor being reused.
+impl<'g, C: Cost> TreeExtractor<'g, C> {
+    /// Computes reusable eq-sort extraction costs for the requested root sorts.
     ///
-    /// For convenience, if the rootsorts is `None`, it defaults to extract all extractable rootsorts.
+    /// Later calls to [`TreeExtractor::extract_best_with_sort`] and
+    /// [`TreeExtractor::extract_variants_with_sort`] reuse the prepared best
+    /// costs and producer choices for reachable eq-sort values. Primitive and
```
yihozhang: "Primitive and container ... enodes" - I find this sentence hard to parse. Similarly for the last paragraph "The cost model must remain ... extractor"

src/extract.rs
```diff
@@ -662,61 +800,42 @@ impl<C: Cost + Ord + Eq + Clone + Debug> Extractor<C> {
...
-        )
-    }
 }
 
 impl Function {
-    /// Returns the extraction head cost for this table.
+    /// Returns the extraction head cost for this table, falling back to `default`.
     /// View tables inherit the cost of their referenced hidden term constructor.
-    pub(crate) fn extraction_head_cost(&self, egraph: &EGraph) -> DefaultCost {
+    pub(crate) fn extraction_head_cost(
```
yihozhang: nit: let this return `Option<DefaultCost>` and the caller does `extraction_header_cost(..).unwrap_or(..)`

src/extract.rs
```diff
@@ -767,48 +886,63 @@ impl Function {
...
 impl EGraph {
-    /// Extract a value to a [`TermDag`] and [`TermId`] in the [`TermDag`] using the default cost model.
-    /// See also [`EGraph::extract_value_with_cost_model`] for more control.
-    pub fn extract_value(
+    /// Extract the best tree term for each requested `(sort, value)` root.
+    ///
+    /// This is the normal user extraction path: it respects `:unextractable`
+    /// and hidden internal functions. The cost model must satisfy the
+    /// optimal-substructure and convergence requirements on [`TreeExtractor`].
+    pub fn extract_best<C: Cost, M: TotalCostModel<C>>(
```
yihozhang: Maybe name this as extract_best_with_cost_model (similarly for the next function) and bring back `extract_value` that defaults `TreeAdditiveCostModel`?

src/extract.rs
```diff
@@ -1,150 +1,314 @@
...
+/// Repeated calls with the same arguments during one extraction must return
+/// equal costs.
+pub trait BaseCostModel<C: Cost> {
+    /// Computes the cost of a non-container primitive value.
+    fn base_value_cost(&self, egraph: &EGraph, sort: &ArcSort, value: Value) -> C;
+}
+
+/// Computes intrinsic costs that exclude selected child and element costs.
+///
+/// Repeated calls with equivalent arguments during one extraction must return
```
yihozhang: Why greedy dag needs to recompute marginal cost? Can't it just cache that?

One of the reasons why I preferred FoldCM over TotalCM was because you only need to call `marginal_enode_cost` once to pin down the cost of each e-node. So that nothing stops us if later we want to export/serialize an extraction problem. Does our extractor implementation have to get the cost multiple times?

yihozhang (commented):
See comments! My main feedback is about trait naming and organization.

I also noticed some documents are sometimes repetitive about, say, optimal substructure property. Also, maybe we should only document these required properties at the cost model declaration site and not the extractor declaration site, since they are properties that make a cost model sound?

src/extract.rs
saulshanabrook: Fixed so we cache it in experimental

src/extract.rs
saulshanabrook: fixed

src/extract.rs
saulshanabrook: fixed

src/extract.rs
saulshanabrook: We could also pass it in on methods, but this way we know its the same e-graph which is required anyways for things to work, so seems better to store it on construction instead of requiring a new one to be passed in.

Also this prevents using the extractor after the e-graph has been mutated which could cause errors.

src/extract.rs
saulshanabrook: Done

src/extract.rs
saulshanabrook: Done

src/extract.rs
saulshanabrook: Updated those and simplified the trait hierarchy

src/extract.rs
saulshanabrook: Done

src/extract.rs
saulshanabrook: Updated

src/extract.rs
saulshanabrook: Fixed

yihozhang (approved):
Thank you!

PR #935, opened 2026-06-25
Add nightly

yihozhang:
A screenshot of the nightly below

image

I was not too careful about the nightly design, but just to get something in there for now, and we can further iterate.

PR #936, opened 2026-06-25
Render nightly benchmark report with eval-live

oflatt-claude:
Follow-up to #935 (targets its `nightly` branch) — ports the nightly benchmark report from the hand-rolled HTML/JS table to [eval-live](https://github.com/oflatt/eval-live), as planned in that PR.

- `render_html()` now flattens each benchmark's per-config cells into eval-live row objects (one row per benchmark, one column per configuration) and embeds `eval_live.css()` / `eval_live.js()` + `initEvalLive(...)` into a self-contained page. Skipped programs are surfaced as a second table.
- Measured cells stay numeric (so eval-live's column filters offer numeric dropdowns); not-applicable cells render as `—`; timeout/error cells keep their status string. The embedded JSON is guarded against a `</script>` breakout.
- `scripts/requirements.txt` pins eval-live (git dependency); `make nightly` installs it first; and `nightly_bench.py` fails fast if eval-live is missing *before* the (potentially long) sweep.

Net diff: roughly −136 / +64 lines — the hand-rolled table markup, CSS, and click-to-sort JS are replaced by eval-live's filtering UI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased] - ReleaseDate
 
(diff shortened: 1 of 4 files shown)


oflatt:
Looks like this works well and negative diff
[index.html](https://github.com/user-attachments/files/29357747/index.html)

PR #938, opened 2026-06-26
Add tuple-output functions

oflatt-claude:
Adds **tuple-output functions** as a first-class alternative to the implicit pair-unboxing in #933.

A function may declare more than one output sort:

```
(function interval (Math) (i64 i64)
  :merge (values (max old0 new0) (min old1 new1)))   ; col 0 = lo, col 1 = hi
```

Outputs are stored as **separate value columns** (no boxing); the functional dependency maps the keys to a tuple of values. They are:
- **destructured** in queries: `(= (values lo hi) (interval x))`
- **written** in actions: `(set (interval x) (values 0 100))`
- **merged** with a `(values e0 e1 …)` clause where `ei` merges column `i`, referencing the bound variables `old0`, `new0`, `old1`, `new1`, …

This collapses analyses that previously needed one function per output (e.g. the two-function interval analysis in `tests/interval.egg`) into a single function.

### Implementation
- `egglog-bridge`: `SchemaMath` now tracks key vs. value columns independently; new per-column `MergeFn::Columns` with `OldCol`/`NewCol` leaves. `core-relations` already supported multi-value tables.
- front-end: `Schema`/`FuncType`/`ResolvedSchema` carry extra output sorts; new `ResolvedCall::Values` and `values` lowering in facts and `set` actions; per-column merge type-checking.
- Tuple outputs are restricted to plain functions (not constructors/relations/view tables) and are not supported by the term/proof encoding (reported cleanly via `Error::UnsupportedProofCommand`).

### Tests
`tests/tuple_outputs.rs` covers the interval analysis, column-independent merge, rule destructuring, eq-sort outputs with rebuild, and the error cases. Full suite green (`files` snapshot suite: 744 passed), `fmt`/`clippy -D warnings` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## [Unreleased] - ReleaseDate
 
+- **Tuple-output functions.** A function may declare more than one output sort, e.g.
+  `(function interval (Math) (i64 i64) :merge (values (max old0 new0) (min old1 new1)))`. Such a
+  function stores its outputs as separate value columns (no boxing); the functional dependency is
+  `keys -> (value0, value1, ...)`. Outputs are destructured in queries with
+  `(= (values lo hi) (interval x))`, written with `(set (interval x) (values 0 100))`, and merged
+  with a `(values ...)` clause whose `i`-th element merges column `i` using the bound variables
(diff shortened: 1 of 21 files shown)


oflatt:
out of date

PR #940, opened 2026-06-29
[codex] Speed up direct rule proof checks

saulshanabrook:
## Summary

This PR makes rule-head proof checking stop once the queried proposition has been produced. The existing `process_actions` evaluator now accepts an optional target proposition; `check_rule_produces_equality` passes the claimed equality so action processing can stop early.

This avoids materializing propositions for irrelevant later head actions when a rule proof only needs one equality from the rule head. The PR also makes subterm reflexive-equality collection track visited `TermId`s, so shared terms in the `TermDag` are not traversed repeatedly as if they were a tree.

This is needed so eggcc proof-testing can finish once extraction gives the checker a proof to validate.

## Repro

Added `tests/proofs/rule-head-fast-path.egg`. The rule first does `(union (A) (B))`, then builds a shared `Pair` tail through 28 `let` levels. The queried equality is produced by the first action, so the checker should not recursively collect propositions from the tail.

Before the checker change, in a clean `upstream/main` worktree with only the fixture added:

```sh
python3 -c "import subprocess, time, sys; cmd=[\"target/release/egglog\",\"--proof-testing\",\"tests/proofs/rule-head-fast-path.egg\"]; start=time.perf_counter();
try:
    r=subprocess.run(cmd, text=True, capture_output=True, timeout=15)
    print(f\"elapsed={time.perf_counter()-start:.3f}s exit={r.returncode}\")
    sys.exit(r.returncode)
except subprocess.TimeoutExpired:
    print(f\"timeout after {time.perf_counter()-start:.3f}s\", file=sys.stderr)
    sys.exit(124)"
```

Result: `timeout after 15.003s`.

After this PR, with the rebuilt release binary:

```sh
target/release/egglog --proof-testing tests/proofs/rule-head-fast-path.egg
```

Result: `real 0.00s` in a direct `/usr/bin/time -p` run after the binary was already built.

## Validation

```sh
cargo test --release --test files rule_head_fast_path -- --nocapture
cargo test --release --test files primitive_args -- --nocapture
cargo test --release selector_merge_cleanup_uses_existing_current_proof -- --nocapture
cargo fmt --check
git diff --check
```

All passed locally. The filtered files harness reported `3 passed` for `rule_head_fast_path` and `3 passed` for `primitive_args`; the selector merge regression reported `1 passed`.


saulshanabrook:
@coderabbitai full review

src/proofs/proof_checker.rs
```diff
@@ -79,11 +79,14 @@ pub(crate) fn run_merge(
 ///    - Reflexive equalities for all subterms
 ///    - Ground equalities from union statements (bidirectional)
 ///    - Reflexive equalities from set statements
+///
+/// If `target` is present, stop once that proposition has been produced.
```
oflatt: Checker is meant to be a reference and as simple as possible, so let's drop this `target` optimization.

src/proofs/proof_checker.rs
```diff
@@ -206,13 +212,31 @@ fn add_subterm_reflexive_equalities(
...
+        propositions,
+        &mut Default::default(),
+    );
+}
+
+fn add_subterm_reflexive_equalities_helper(
+    term_id: TermId,
+    term_dag: &TermDag,
+    propositions: &mut HashSet<Proposition>,
+    seen: &mut HashSet<TermId>,
```
oflatt, line 221: nice change

oflatt (changes requested):
Nice fix and test case, just want to simplify as much as possible

src/proofs/proof_checker.rs
saulshanabrook: Sounds good! I tested and works fine without this.

oflatt (approved):
nice!

PR #941, opened 2026-06-29
[codex] Add root-directed proof extraction

saulshanabrook:
## Summary

This changes proof-mode `prove_exists` to extract only the requested proof root instead of computing extraction costs for the whole e-graph. The new extractor is private to `src/proofs/` and returns any reconstructable proof term for the requested `(Value, Sort)` root.

The previous path used `Extractor::compute_costs_from_rootsorts_allow_unextractable(None, ...)`, which can spend time on unrelated proof terms before proof checking even starts. For proof existence checks we only need one proof term for the recorded proof value.

## Notes

- Moves the proof-specific extraction behavior into `src/proofs/proof_extractor.rs`.
- Removes the proof-only `compute_costs_from_rootsorts_allow_unextractable` helper from the public extraction module.
- Keeps the extractor private and adds only a minimal direct-constructor root test.

## Validation

```sh
cargo test --release --lib extracts_direct_constructor_root -- --nocapture
cargo test --release --lib proof_mode_allows -- --nocapture
cargo test --release --test files primitive_args -- --nocapture
cargo test --release --test files eqsat_basic_proof -- --nocapture
cargo fmt --check
git diff --check
```

### DIFF
--- a/src/extract.rs
+++ b/src/extract.rs
@@ -134,24 +134,6 @@ pub struct Extractor<C: Cost + Ord + Eq + Clone + Debug> {
     parent_edge: HashMap<String, HashMap<Value, (String, Vec<Value>)>>,
 }
 
-/// Options for configuring extraction behavior.
-struct ExtractionOptions<C: Cost> {
-    /// The cost model to use for extraction.
-    cost_model: Box<dyn CostModel<C>>,
-    /// Root sorts to extract from. If None, all extractable root sorts are used.
-    rootsorts: Option<Vec<ArcSort>>,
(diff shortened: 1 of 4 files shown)


saulshanabrook:
@coderabbitai full review

saulshanabrook:
@coderabbitai full review

oflatt (changes requested):
Sweet! I'm guessing there is stuff (flags?) you can now remove from our normal extractor to make things cleaner.

saulshanabrook:
Fixed!

oflatt (approved):
nice

PR #947, opened 2026-07-13
Egglog AST macros with unquote and unquote splice

oflatt:
See the changelog for this PR!


yihozhang (commented):
Just a few thoughts and comments. I have not done a full review of the code

* The lsp seems to be confused currently
Image

* This PR is nice in that it reuses and parametrizes over the parser (so supports egglog macros), and it does not touch most of the existing code (so less pressure for code review).
  * We need to have test in experimental making sure the parser also works there.
* I think the team needs to discuss if this is the interface we will commit to (since it seems this will replace prelude). My worry is this is not as general as the rust function-based apis, even though we have unquote. (edit: this looks very general)
* I was originally thinking this could be in experimental or its own repo, but it seems if this is replacing prelude then it should be in this repo?
* quote-macro is currently not tested?
* Do we want both quote-macro and checked-macro? It seems both are useful, but I'm worried that this will confuse users.

quote-macro/src/lib.rs
```diff
@@ -0,0 +1,647 @@
...
+}
+
+/// Build an *un-parsed* s-expression: `sexp!(<egglog>)` → `Sexp` (no parser
+/// argument). Supports `#` / `#..` / `:#` splices, and is meant to be
+/// `#`-spliced into another quasiquote later (where a `Sexp` is used as-is).
+///
+/// ```ignore
+/// let kind = "MyOp";
+/// let fields = vec!["?a", "?b"];
+/// let frag = sexp!((#kind #..fields));   // Sexp `(MyOp ?a ?b)`
```
yihozhang, line 201: Why is one `sexp!` while another is `expr!`?

oflatt: An Expr can't have special custom syntax if you have your own parser macros, so we provide it in case you are using it.
Luminal might use this so they can use experimental named arguments with these macros

PR #953, opened 2026-07-18
Skip building indexes for probers that will only be probed once

yihozhang:
Profiling shows >50% of the indices are only built and then probed only once, so it is wasteful. This PR avoids building the index when this happens, and instead just scans the subset.

It also adds an index_stats module and a compile flag that, when enabled, shows the stats of indices built.

Also updated bench.py to interleave the execution when benchmarking.

Results:

```
# Benchmark diff

  Generated  : 2026-07-18T15:42:14
  Baseline   : 56b84600 (56b84600db70)  —  Merge pull request #948 from egraphs-good/perf-sorted-column-index
  Comparison : 64e8fe39 (64e8fe3953d0)  —  interleave run

  hardboiled_conv1d_32.egg                        0.134       0.139  +  0.005  +   3.4%  ▲ slower
  hardboiled_conv1d_128.egg                       0.340       0.355  +  0.015  +   4.5%  ▲ slower
  luminal-llama.egg                               0.083       0.081   -0.002    -1.9%  ▼ faster
  python_array_optimize.egg                       0.266       0.230   -0.035   -13.3%  ▼ faster
  cykjson.egg                                     0.033       0.033  +  0.000  +   0.1%  ·
  eggcc-extraction.egg                            0.213       0.215  +  0.002  +   0.9%  ▲ slower
  llama.egg                                       0.435       0.415   -0.021    -4.7%  ▼ faster
  paged_llama.egg                                 1.679       1.686  +  0.007  +   0.4%  ·
  qwen.egg                                        0.485       0.455   -0.030    -6.1%  ▼ faster
  qwen3_moe.egg                                   0.678       0.642   -0.035    -5.2%  ▼ faster
  whisper.egg                                     1.513       1.374   -0.139    -9.2%  ▼ faster

  Summary: 6 faster  ·  3 slower  ·  2 unchanged  ·  0 missing
  Overall average Δ: -2.84%

  VERDICT: IMPROVEMENT
```

I am not too sure how to eliminate the regression.


ezrosent:
What about the regressions for hardboiled?

yihozhang:
@ezrosent I'm not exactly sure about what leads to the regression (the regression seems to be consistent for this benchmark). My suspicion is that this is due to the atomic counter and the additional checking, and maybe the more bloated `run_plan`. I can look deeper if you think this can be further improved!

PR #954, opened 2026-07-19
Improve parallel scaling of e-matching

ezrosent:
(branched off of `perf-shared-trie-nodes`)

This change improves parallel e-matching scaling with a few optimizations:

* Avoids highly-contended `Arc::clone` calls with index construction: handles on shared indexes are grabbed before the plan starts running to avoid big clone calls. This alone was a large win.
* Add a "fast path" where the top relation is coarse-grained parallelized rather than the fine-grained behavior we had before. This is another big speedup, when the top relation is large enough. (Removing Arc clones was still higher though!).
* Reintroduce a limited form of work-stealing by adding a separate `spawn_local` method onto the thread pool that allows for work to be appended to a local deque first, only migrating to the global queue if sufficient threads are stalled. This appears to help cases where there isn't a lot of parallelism to expose. The thread pool essentially sprayed data randomly across threads, which makes cache locality much harder to achieve. (The original morsel-driven parallelism paper talks about this too). Local spawns maintain locality while still improving parallel utilization if there's a lot of skew.

To evaluate this I had codex write up a benchmark using the dataset and queries from the Honeycomb paper. This is just a single query over a large dataset, so it's a helpful test case for this part of the code: other egglog benchmarks have complex schedules and many rules that we parallelize across sometimes.

Here are the overall results on my m4 max laptop:

(Note that M4 max only has 12 P-cores, so some amount of flattening after 12 is expected. Still, I suspect there's more to do here)

Workload | 1 thread | 8 threads | 12 threads | 16 threads
-- | -- | -- | -- | --
Honeycomb q6/count | 9.271s → 9.126s (1.02×) | 3.693s → 1.778s (2.08×) | 3.476s → 1.480s (2.35×) | 3.691s → 1.411s (2.62×)
gemma.egg | 3.436s → 3.615s (0.95×) | 1.462s → 1.214s (1.20×) | 1.524s → 1.200s (1.27×) | 1.651s → 1.273s (1.30×)
gemma4_moe.egg | 12.366s → 12.864s (0.96×) | 4.695s → 3.865s (1.21×) | 4.748s → 3.713s (1.28×) | 4.870s → 3.894s (1.25×)
hardboiled_conv1d_128.egg | 0.220s → 0.219s (1.00×) | 0.256s → 0.210s (1.22×) | 0.339s → 0.286s (1.19×) | 0.415s → 0.330s (1.26×)
hardboiled_conv1d_32.egg | 0.089s → 0.090s (0.99×) | 0.106s → 0.107s (1.00×) | 0.109s → 0.109s (1.00×) | 0.115s → 0.116s (0.99×)
llama.egg | 0.306s → 0.308s (1.00×) | 0.340s → 0.345s (0.99×) | 0.344s → 0.346s (1.00×) | 0.350s → 0.350s (1.00×)
luminal-llama.egg | 0.065s → 0.081s (0.80×) | 0.078s → 0.095s (0.82×) | 0.085s → 0.102s (0.83×) | 0.085s → 0.102s (0.83×)
paged_llama.egg | 1.204s → 1.204s (1.00×) | 1.262s → 1.262s (1.00×) | 1.289s → 1.295s (1.00×) | 1.281s → 1.287s (0.99×)
qwen.egg | 0.349s → 0.354s (0.99×) | 0.394s → 0.364s (1.08×) | 0.423s → 0.391s (1.08×) | 0.446s → 0.407s (1.09×)
qwen3_moe.egg | 0.458s → 0.466s (0.98×) | 0.502s → 0.440s (1.14×) | 0.535s → 0.468s (1.14×) | 0.587s → 0.510s (1.15×)
whisper.egg | 0.994s → 1.018s (0.98×) | 0.742s → 0.605s (1.23×) | 0.784s → 0.629s (1.25×) | 0.868s → 0.679s (1.28×)
Geomean, .egg workloads | 0.96× | 1.08× | 1.09× | 1.10×

### DIFF
--- a/core-relations/src/free_join/execute.rs
+++ b/core-relations/src/free_join/execute.rs
@@ -27,7 +27,7 @@ use web_time::Instant;
 use crate::{
     Constraint, OffsetRange, Pool, SubsetRef,
     action::{Bindings, ExecutionState},
-    common::{DashMap, ShardData, ShardId, Value},
+    common::{DashMap, Value},
     free_join::{
         frame_update::{FrameUpdates, UpdateInstr},
         get_index_from_tableinfo,
@@ -35,9 +35,8 @@ use crate::{
     hash_index::{ColumnIndex, Index, IndexBase, TupleIndex},
     offsets::{Offsets, RowId, SortedOffsetSlice, SortedOffsetVector, Subset},
     parallel_heuristics::{
-        GjChildCache, GjTopPromotion, action_batch_size, free_join_fork_depth, gj_child_cache,
-        gj_local_depth, gj_local_morsel, gj_local_queue_limit, gj_min_keys_per_worker,
-        gj_top_index_sharding, gj_top_promotion, parallelize_db_level_op,
+        MIN_TOP_INDEX_KEYS_PER_WORKER, action_batch_size, free_join_fork_depth,
+        parallelize_db_level_op,
     },
     pool::Pooled,
     query::RuleSet,
@@ -63,41 +62,6 @@ fn top_index_shape_is_eligible(
     workers > 1
         && leader_keys >= min_keys_per_worker.saturating_mul(workers)
         && nonempty_shards >= workers
-}
-
-/// Select one already-evaluated eligible candidate. The iterator remains lazy,
-/// so `Current` evaluates only position zero and `Eligible` stops at its first
-/// usable fallback; only `Largest` evaluates the full
-/// leading prefix.
-fn select_top_candidate<T>(
-    policy: GjTopPromotion,
-    current_estimate: usize,
-    min_threshold: usize,
-    mut candidates: impl Iterator<Item = Option<T>>,
-    leader_keys: impl Fn(&T) -> usize,
-) -> Option<T> {
-    let current = candidates.next().flatten();
-    match policy {
-        GjTopPromotion::Current => current,
-        GjTopPromotion::Eligible => {
-            if current.is_some() {
-                return current;
-            }
-            let promotion_limit = current_estimate.saturating_mul(4).max(min_threshold);
-            candidates
-                .flatten()
-                .find(|candidate| leader_keys(candidate) <= promotion_limit)
-        }
-        GjTopPromotion::Largest => {
-            current
-                .into_iter()
-                .chain(candidates.flatten())
-                .fold(None, |best, candidate| match best {
-                    Some(best) if leader_keys(&best) >= leader_keys(&candidate) => Some(best),
-                    _ => Some(candidate),
-                })
-        }
-    }
 }
 
 struct SparseColumnIndex {
@@ -1034,88 +998,35 @@ type ColumnIndexes = IdVec<ColumnId, OnceLock<Arc<SortedColumnIndex>>>;
 // (node, col, value) with different slow constraints; they are almost always
 // empty, in which case the guard is a cheap length check.
 type ChildEntry = (Arc<TrieNode>, Box<[Constraint]>);
-type ChildLock = RwLock<HashMap<Value, ChildEntry>>;
-type ChildrenMaps = IdVec<ColumnId, OnceLock<ChildMap>>;
+pub(crate) type ChildLock = RwLock<HashMap<Value, ChildEntry>>;
+type ChildrenMaps = IdVec<ColumnId, OnceLock<ChildLock>>;
 
-/// Cache of child trie nodes for one `(TrieNode, ColumnId)` pair.
-///
-/// Descendant nodes lazily initialize [`ChildMap::Single`]. An eligible coarse
-/// top-index partition may initialize a root column as [`ChildMap::Sharded`]
-/// before its jobs start, using the same [`ShardData`] layout as the leader's
-/// column index. There is deliberately no promotion from `Single` to `Sharded`:
-/// an already-used cache remains valid and simply falls back to the old layout.
-pub(crate) enum ChildMap {
-    Single(ChildLock),
-    Sharded {
-        shard_data: ShardData,
-        shards: IdVec<ShardId, CachePadded<ChildLock>>,
-    },
-}
-
-impl ChildMap {
-    fn single() -> Self {
-        Self::Single(RwLock::new(HashMap::default()))
-    }
-
-    fn sharded(shard_count: usize) -> Self {
-        let shard_data = ShardData::new(shard_count);
-        debug_assert_eq!(shard_data.n_shards(), shard_count);
-        let mut shards = IdVec::with_capacity(shard_data.n_shards());
-        shards.resize_with(shard_data.n_shards(), || {
-            CachePadded::new(RwLock::new(HashMap::default()))
-        });
-        Self::Sharded { shard_data, shards }
-    }
-
-    fn is_aligned_to(&self, shard_count: usize) -> bool {
-        matches!(
-            self,
-            Self::Sharded { shard_data, .. } if shard_data.n_shards() == shard_count
-        )
-    }
-
-    /// Select the lock with the same `Value` hashing and high-bit shard mapping
-    /// used by [`crate::hash_index::ColumnIndex`].
-    fn lock_for(&self, value: Value) -> &ChildLock {
-        match self {
-            Self::Single(lock) => lock,
-            Self::Sharded { shard_data, shards } => {
-                shard_data.get_shard(&value, shards) as &ChildLock
-            }
-        }
-    }
-
-    fn get_or_insert(
-        &self,
-        value: Value,
-        edge_cs: &[Constraint],
-        sub: impl FnOnce() -> Subset,
-    ) -> Arc<TrieNode> {
-        let map = self.lock_for(value);
-        // Optimistic read path: most calls are cache hits, so try a shared lock
-        // first. A hit is only valid when the edge constraints match.
-        {
-            let guard = map.read().unwrap();
-            if let Some((node, stored_cs)) = guard.get(&value)
-                && &**stored_cs == edge_cs
-            {
-                return node.clone();
-            }
-        }
-        // Cache miss (or constraint mismatch): acquire the write lock and insert.
-        let mut guard = map.write().unwrap();
+fn get_or_insert_child(
+    map: &ChildLock,
+    value: Value,
+    edge_cs: &[Constraint],
+    sub: impl FnOnce() -> Subset,
+) -> Arc<TrieNode> {
+    // Optimistic read path: most calls are cache hits, so try a shared lock
+    // first. A hit is only valid when the edge constraints match.
+    {
+        let guard = map.read().unwrap();
         if let Some((node, stored_cs)) = guard.get(&value)
             && &**stored_cs == edge_cs
         {
             return node.clone();
         }
-        // Child nodes intentionally start with no cache layout. If they are
-        // probed later, they lazily choose `Single` and never inherit root
-        // sharding from this map.
-        let new_node = Arc::new(TrieNode::new(sub()));
-        guard.insert(value, (new_node.clone(), Box::from(edge_cs)));
-        new_node
     }
+    // Cache miss (or constraint mismatch): acquire the write lock and insert.
+    let mut guard = map.write().unwrap();
+    if let Some((node, stored_cs)) = guard.get(&value)
+        && &**stored_cs == edge_cs
+    {
+        return node.clone();
+    }
+    let new_node = Arc::new(TrieNode::new(sub()));
+    guard.insert(value, (new_node.clone(), Box::from(edge_cs)));
+    new_node
 }
 
 /// Canonical signature of a trie root: the table plus its sorted header (fast)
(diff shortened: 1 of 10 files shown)


core-relations/src/free_join/execute.rs
```diff
@@ -440,8 +644,39 @@ impl Database {
...
+                TrieCache::compute_shared(rule_set.plans.values().map(|(plan, _, _)| plan));
+            (!shared.is_empty()).then(|| Arc::new(TrieCache::with_shared(shared)))
+        };
 
         let search_and_apply_timer = Instant::now();
+        // Prepare one lazy slot for every table-index access named by the
+        // immutable plans before any worker starts. A slot acquires and refreshes
+        // its Arc through the regular catalog helper on first cached use. The
+        // sidecars are dropped before `merge_all` resets catalog entries.
+        let prepared_plans = rule_set
```
yihozhang: nit: call them `prepared_indexes` instead?

core-relations/src/free_join/execute.rs
```diff
@@ -606,38 +847,34 @@ impl Database {
...
+                            plan.header().iter().filter(|h| h.atom == id).collect();
+                        match join_state.root_node(info.table, &headers) {
+                            Some(node) => binding_info.insert_node(id, node),
+                            None => break 'eval,
                         }
-                        binding_info.move_back_node(*atom, Arc::new(cur));
                     }
-                    match plan {
-                        Plan::SinglePlan(plan) => {
+                    match (plan, prepared_plan) {
```
yihozhang: nit: `prepared_index`?

core-relations/src/free_join/execute.rs
```diff
@@ -975,14 +1445,120 @@ impl<'a> JoinState<'a> {
...
+        }
+
+        let mut leader = 0;
+        let mut leader_size = usize::MAX;
+        let mut probers = Vec::with_capacity(scans.len());
+        for (i, (scan, prepared)) in scans.iter().zip(prepared).enumerate() {
+            let prober =
+                self.get_column_index(atoms, binding_info, scan.atom, scan.column, prepared);
+            let size = prober.len();
+            // The two-way hot path historically breaks ties in favor of the
```
yihozhang: I wonder if it's cleaner to instead change run_plan to match the natural tie-breaking behavior here.

core-relations/src/free_join/execute.rs
```diff
@@ -2390,7 +3363,7 @@ fn sort_plan_by_size(
     let mut last_pos = start;
     for i in start..instrs.len() {
         if matches!(
-            &instrs[i],
+            &instrs[order.get(i)],
```
yihozhang: is it true that i == order.get(i) at this point?

core-relations/src/free_join/execute.rs
```diff
@@ -2507,8 +3480,8 @@ fn sort_plan_by_size_inner(
     let mut times_refined = with_pool_set(|ps| ps.get::<DenseIdMap<AtomId, i64>>());
 
     // Count how many times each atom has been refined so far.
-    for ins in instrs[..range.start].iter() {
-        match ins {
+    for position in 0..range.start {
```
yihozhang: i also don't quite understand this change

core-relations/src/free_join/execute.rs
```diff
@@ -1015,35 +1592,92 @@ impl<'a> JoinState<'a> {
         let mut order = InstrOrder::from_iter(0..stages.instrs.len());
         let mut leaf_scans: LeafScans = smallvec::smallvec![false; stages.instrs.len()];
         sort_plan_by_size(&mut order, &mut leaf_scans, 0, &stages.instrs, binding_info);
+
+        // A cached top index is already partitioned by hash. Schedule one
+        // coarse global job per nonempty shard and run each shard's subtree
+        // serially. Besides avoiding an intermediate key copy, this keeps
+        // related nested probes on one worker. Buffers that cannot construct an
+        // independent partition (the in-place executors) decline this path.
+        if !stages.instrs.is_empty()
```
yihozhang, line 1593: Is the idea that if the top level has enough entries, we will only do top-level parallelism? This coarser-grained parallelism can be technically more sensitive to skews but it does not show up in practice and it avoids some overheads with nested parallelism?

Have you measured the percentage of hits on this path versus the nested parallelism path? I feel this top-level parallelism may overshadow the default nested parallelism paths for most of the rules.

core-relations/src/free_join/execute.rs
```diff
@@ -1015,35 +1592,92 @@ impl<'a> JoinState<'a> {
...
-    /// which is the current "fan-out" node we are in. The latter parameter is an experimental
-    /// index used to detect if we are at the "top" of a plan rather than the "bottom", and is
-    /// currently used as a heuristic to determine if we should increase parallelism more than the
-    /// default.
+    /// This method takes the plan, mutable data structures for variable binding
+    /// and staging actions, and `cur`, the current stage of the plan. A top-level
+    /// coarse partition also passes `index_shard`; recursive calls clear it so
+    /// only the first intersection is restricted to that physical shard.
     #[allow(clippy::too_many_arguments)]
     fn run_plan<'buf, A: NumericId + 'buf, BUF: ActionBuffer<'buf, A>>(
```
yihozhang, line 1666: maybe document the assumption that index_shard.some() implies no more parallelism at this level.

core-relations/src/free_join/execute.rs
```diff
@@ -2030,6 +2751,103 @@ impl<'a, 'outer: 'a> ActionBuffer<'a, ActionId> for InPlaceActionBuffer<'outer>
...
+        = Self
+    where
+        'scope: 'a;
+
+    fn push_bindings(
+        &mut self,
+        action: ActionId,
+        bindings: &DenseIdMap<Variable, Value>,
+        mut to_exec_state: impl FnMut() -> ExecutionState<'scope>,
+    ) {
```
yihozhang, line 2807: should we set  `self.needs_flush = true;` here? (was comparing this and `ScopedActionBuffer::push_bindings`)

core-relations/src/free_join/execute.rs
```diff
@@ -2030,6 +2751,103 @@ impl<'a, 'outer: 'a> ActionBuffer<'a, ActionId> for InPlaceActionBuffer<'outer>
...
+        }
+    }
+
+    fn flush(&mut self, exec_state: &mut ExecutionState) {
+        flush_action_states(
+            exec_state,
+            &mut self.batches,
+            self.rule_set,
+            self.match_counter.as_ref(),
+        );
```
yihozhang, line 2835: should we set `self.needs_flush = false;`?

yihozhang (commented):
Had some nits and questions. I didn't carefully review the new threadpool implementation besides the high-level description. I also didn't carefully read the new tests.

yihozhang:
From running `uv run scripts/bench.py run perf-shared-trie-nodes ezr-better-parallel` (serial mode).

```
# Benchmark diff

  Generated  : 2026-07-25T14:17:12
  Baseline   : 225ad44d (225ad44d2379)  —  add a comment
  Comparison : a384d253 (a384d25325da)  —  Remove retired generic-join tuning paths

  hardboiled_conv1d_32.egg                        0.118       0.115   -0.003    -2.5%  ▼ faster
  hardboiled_conv1d_128.egg                       0.301       0.291   -0.010    -3.3%  ▼ faster
  luminal-llama.egg                               0.083       0.111  +  0.028  +  33.5%  ▲ slower
  python_array_optimize.egg                       0.233       0.235  +  0.002  +   0.8%  ▲ slower
  cykjson.egg                                     0.032       0.034  +  0.002  +   5.5%  ▲ slower
  eggcc-extraction.egg                            0.203       0.216  +  0.012  +   6.0%  ▲ slower
  llama.egg                                       0.373       0.379  +  0.006  +   1.7%  ▲ slower
  paged_llama.egg                                 1.616       1.634  +  0.018  +   1.1%  ▲ slower
  qwen.egg                                        0.424       0.447  +  0.023  +   5.4%  ▲ slower
  qwen3_moe.egg                                   0.586       0.588  +  0.002  +   0.3%  ·
  whisper.egg                                     1.249       1.278  +  0.029  +   2.3%  ▲ slower
```

Update: numbers from nightly are mostly consistent with the above

* ezr-better-parallel: https://nightly.cs.washington.edu/reports/egglog/1784552964:ezr-better-parallel:a384d253/
* perf-shared-trie-nodes: https://nightly.cs.washington.edu/reports/egglog/1784364193:perf-shared-trie-nodes:225ad44d/

But the packed trie PR #959  shows some significant speedups.
* ezr-better-parallel-packed-trie: https://nightly.cs.washington.edu/reports/egglog/1784967636:ezr-better-parallel-packed-trie:c7fc1d77/

core-relations/src/free_join/execute.rs
ezrosent: This change is bad! good catch. It's fixed in a follow-up and I can try and backport it if the rest looks good.

ezrosent:
I *think* the llama slowdown is due to the buggy variable ordering heuristic that codex snuck in; it's reverted in the packed-trie branch.

core-relations/src/free_join/execute.rs
ezrosent: >Have you measured the percentage of hits on this path versus the nested parallelism path? I feel this top-level parallelism may overshadow the default nested parallelism paths for most of the rules.

Yes, this ends up dominating most of the heavy benchmarks. The overhead of fine-grained parallelism at the toplevel was too high: too much of the total work was done in the top two layers I think.

core-relations/src/free_join/execute.rs
ezrosent: Renamed the ruleset-level collection to `prepared_indexes`; the later packed-trie layer also uses `prepared_index` for each query-local sidecar.

core-relations/src/free_join/execute.rs
ezrosent: Renamed the per-plan binding to `prepared_index`, including the later lazy query-local construction.

core-relations/src/free_join/execute.rs
ezrosent: Done: the two-way `run_plan` path now chooses the first scan on equal sizes (`<=`), matching the natural first-minimum behavior of the general and partition-discovery loops. The special two-way tie case is gone.

core-relations/src/free_join/execute.rs
ezrosent: Not after a recursive reorder. I restored the pre-PR behavior here: barriers are identified in the cached logical order with `instrs[i]`.

core-relations/src/free_join/execute.rs
ezrosent: Agreed; this was the DVO regression. The bottom PR now restores the original logical-prefix refinement seed from `instrs[..range.start]`.

core-relations/src/free_join/execute.rs
ezrosent: Yes. The production path uses coarse top partitions with each shard subtree inline; the fine-grained path remains the fallback when the top shape is ineligible. I expanded the code comment to state that choosing this path commits the subtree to coarse-only parallelism. The later experimental policy PR documents when it may deliberately fan out again.

core-relations/src/free_join/execute.rs
ezrosent: Documented: a present top partition is paired with a serial action buffer in this PR, so no nested parallelism is available in that subtree. Recursive calls clear only the physical-shard restriction; the serial buffer keeps later work inline.

core-relations/src/free_join/execute.rs
ezrosent: No flag is needed for `SerialScopedActionBuffer`: its owning shard job unconditionally calls `flush` once after the callback, and recursive calls reuse the same buffer synchronously. I documented that invariant on the type and at the flush site.

core-relations/src/free_join/execute.rs
ezrosent: Likewise, there is no `needs_flush` state to clear: the serial shard buffer is always flushed exactly once immediately before its owner exits.

yihozhang (approved):

yihozhang:
new perf numbers using this [script](https://github.com/egraphs-good/egglog/blob/64e8fe3953d03e3219d1898ffdeee7dfc5d4c8d2/scripts/bench.py)

(I had another run using the existing bench.py that's worse but I think it's the artifact of that script. The updated regression is much slower.)

```
# Benchmark diff

  Generated  : 2026-08-11T17:18:41
  Baseline   : 95f539cc (95f539cca027)  —  Merge pull request #973 from egraphs-good/fix/971-intersect-transient-unsorted
  Comparison : e4ebef25 (e4ebef25b5d3)  —  Address parallel join review feedback

  hardboiled_conv1d_32.egg                        0.115       0.116  +  0.001  +   1.0%  ▲ slower
  hardboiled_conv1d_128.egg                       0.291       0.293  +  0.002  +   0.6%  ▲ slower
  luminal-llama.egg                               0.071       0.072  +  0.001  +   0.7%  ▲ slower
  python_array_optimize.egg                       0.230       0.234  +  0.004  +   1.9%  ▲ slower
  cykjson.egg                                     0.032       0.033  +  0.001  +   1.9%  ▲ slower
  eggcc-extraction.egg                            0.213       0.215  +  0.002  +   1.0%  ▲ slower
  llama.egg                                       0.371       0.372  +  0.000  +   0.1%  ·
  paged_llama.egg                                 1.655       1.680  +  0.024  +   1.5%  ▲ slower
  qwen.egg                                        0.414       0.415  +  0.000  +   0.1%  ·
  qwen3_moe.egg                                   0.549       0.550  +  0.001  +   0.2%  ·
  whisper.egg                                     1.148       1.154  +  0.007  +   0.6%  ▲ slower

  Summary: 0 faster  ·  8 slower  ·  3 unchanged  ·  0 missing
  Overall average Δ: +0.87%
```

gemma4_moe.egg is up to ~10% slower
- https://nightly.cs.washington.edu/reports/egglog/1786383546:ezr-better-parallel:e4ebef25/
- https://nightly.cs.washington.edu/reports/egglog/1786263356:main:95f539cc/

saulshanabrook:
FWIW if you want confidence intervals on your % change, feel free to crib this from our proof repo, using fieller's theorem for the CI of ration of two means: https://github.com/saulshanabrook/egglog-encoding/blob/ffb8ae435bd6421077b1c15826f32a6aeecf5b1b/benchmarking/reports/analysis.py#L248-L278 https://github.com/saulshanabrook/egglog-encoding#statistics

PR #955, opened 2026-07-20
Patch canonicalized rows in place in the rebuild output buffer

yihozhang:
Canonicalizer::rebuild_buf copied every changed row into a pooled scratch vector, patched it there, and copied it again into the output TaggedRowBuffer. Add TaggedRowBuffer::add_row_with, which appends the original row and lets the caller patch the copy in place, and use it to drop the scratch buffer entirely (-4.7% instructions on eggcc-2mm).

Added another micro-optimization that doesn't radix sort an index if it's already sorted.

### DIFF
--- a/core-relations/src/hash_index/mod.rs
+++ b/core-relations/src/hash_index/mod.rs
@@ -435,7 +435,21 @@ pub(crate) fn radix_sort_slice_by_value(
         return;
     }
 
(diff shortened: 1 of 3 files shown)


oflatt (approved):
I'll say it's ok as long as the unsafe part is ok!

PR #959, opened 2026-07-23
Use packed arena tries for generic joins

ezrosent:
Superseded by the smaller draft stack #993 through #1000. Those eight PRs jointly replace this PR; this draft remains open only as a historical reference for the original combined diff.

Stacked on the packed-node and positioned-index abstractions PR.

This was the original combined packed-executor integration:
- switch generic joins to arena-packed trie nodes
- use positioned indexes throughout join execution
- integrate packed cursors, constraints, and probing into the executor

Replacement-stack validation:
- cargo fmt --check
- 113 tests passed

### DIFF
--- a/core-relations/src/free_join/execute.rs
+++ b/core-relations/src/free_join/execute.rs
@@ -1,10 +1,11 @@
 //! Core free join execution.
 
 use std::{
-    cmp, iter, mem,
+    cell::RefCell,
+    cmp, mem,
     ops::Range,
     sync::{
-        Arc, OnceLock, RwLock,
+        Arc, Mutex, OnceLock,
         atomic::{AtomicUsize, Ordering},
     },
 };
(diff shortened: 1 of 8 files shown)


ezrosent:
I'm currently leaning towards this not being worth it. I think we can probably get away with the old layout: the code is simpler and I think most benchmarks don't show a clear benefit.

yihozhang:
I got some decent performance gains from the packed arena tries. Maybe there's a less aggressive and more maintainable version of packed tries we can try?

```
  hardboiled_conv1d_32.egg                        0.111       0.116  +  0.004  +   4.0%  ▲ slower
  hardboiled_conv1d_128.egg                       0.283       0.282   -0.001    -0.4%  ·
  luminal-llama.egg                               0.079       0.084  +  0.005  +   6.3%  ▲ slower
  python_array_optimize.egg                       0.237       0.252  +  0.016  +   6.6%  ▲ slower
  cykjson.egg                                     0.033       0.035  +  0.002  +   6.3%  ▲ slower
  eggcc-extraction.egg                            0.212       0.211   -0.001    -0.4%  ·
  llama.egg                                       0.379       0.340   -0.039   -10.3%  ▼ faster
  paged_llama.egg                                 1.686       1.172   -0.515   -30.5%  ▼ faster
  qwen.egg                                        0.414       0.378   -0.037    -8.8%  ▼ faster
  qwen3_moe.egg                                   0.553       0.482   -0.071   -12.8%  ▼ faster
  whisper.egg                                     1.119       0.791   -0.328   -29.3%  ▼ faster
```

core-relations/src/free_join/execute.rs
ezrosent, line 1: I think we probably need to split up this file... it's a bit of a monster at this point and is getting impossible to understand.

PR #966, opened 2026-07-30
Gallop sorted free-join intersections

ezrosent:
## Summary

- Intersect sorted two-way generic-join probes with a two-finger merge for small or similarly sized inputs and exponential galloping for skewed inputs.
- Preserve row-carrying `ProbeMatch` semantics for `Packed`, `ProjectedRoot`, and `SmallColumn` indexes, with the existing scalar fallback for hash-backed and unsupported shapes.
- Drop the serial action-batch increase after it changed fixed-iteration `eggcc-2mm` results and regressed `stresstest_large_expr`.

## Stack

Draft, stacked on #962. This PR now contains one focused performance commit.

## Performance

Local wall-time comparisons against #962 after removing the batch-size change:

- `stresstest_large_expr`: 196.7 ms base vs 196.8 ms head median (neutral).
- `eggcc-2mm`: 965.5 ms base vs 949.7 ms head median (1.6% faster).
- Three-sample neural-network smoke checks ranged from about 4.1% faster on Gemma4-MoE to about 1.4% slower on Llama; the smaller movements are within the noise floor of this short run.

The previous CodSpeed PR comment compares the now-removed `a9c79c8` batch-size version and also warns that its base and head used different runtime environments. GitHub did not emit a new `pull_request:synchronize` event for the stack force-update, so that bot comment remains stale.

The full Build workflow was explicitly dispatched on `3ae16b04`; all 30 jobs passed, including fresh `stresstest_large_expr` and `eggcc-2mm` simulation shards:

https://github.com/egraphs-good/egglog/actions/runs/30564246486

## Validation

- `cargo fmt --check`
- `cargo clippy -p egglog-core-relations --all-targets -- -D warnings`
- `cargo test -p egglog-core-relations`: 148 tests passed
- 2 doctests passed
- `cargo test --release --test files -- eggcc_2mm`: all three variants passed
- Full GitHub Build workflow: 30 of 30 jobs passed

### DIFF
--- a/core-relations/src/free_join/execute.rs
+++ b/core-relations/src/free_join/execute.rs
@@ -912,6 +912,61 @@ fn intersect_with_dense_ref<'a>(v: SubsetRef<'a>, range: OffsetRange) -> Option<
     }
 }
 
+/// Seek `key` in a sorted scalar index without moving backward.
+///
+/// Exponential search followed by a bounded binary search avoids walking a
+/// large target when the sorted query is sparse.
+fn seek_sorted_key(
+    key: Value,
+    target_len: usize,
+    target_cursor: &mut usize,
+    target_at: impl Fn(usize) -> Value,
+) -> bool {
+    if *target_cursor >= target_len {
+        return false;
+    }
+
+    let current = target_at(*target_cursor);
+    match current.cmp(&key) {
+        cmp::Ordering::Equal => true,
+        cmp::Ordering::Greater => false,
+        cmp::Ordering::Less => {
+            let base = *target_cursor;
+            let mut step = 1usize;
+            while let Some(position) = base.checked_add(step).filter(|&pos| pos < target_len) {
+                if target_at(position) >= key {
+                    break;
+                }
+                let Some(next) = step.checked_mul(2) else {
+                    step = target_len;
+                    break;
+                };
+                step = next;
+            }
+
+            let previous = step / 2;
+            let mut lo = base
+                .saturating_add(previous)
+                .saturating_add(1)
+                .min(target_len);
+            let mut hi = base.saturating_add(step).saturating_add(1).min(target_len);
+            while lo < hi {
+                let mid = lo + (hi - lo) / 2;
+                if target_at(mid) < key {
+                    lo = mid + 1;
+                } else {
+                    hi = mid;
+                }
+            }
+            *target_cursor = lo;
+            if lo >= target_len {
+                return false;
+            }
+            target_at(lo) == key
+        }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


core-relations/src/free_join/execute.rs
```diff
@@ -1023,6 +1023,61 @@ fn intersect_with_dense_ref<'a>(v: SubsetRef<'a>, range: OffsetRange) -> Option<
...
+        cmp::Ordering::Equal => true,
+        cmp::Ordering::Greater => false,
+        cmp::Ordering::Less => {
+            let base = *target_cursor;
+            let mut step = 1usize;
+            while let Some(position) = base.checked_add(step).filter(|&pos| pos < target_len) {
+                if target_at(position) >= key {
+                    break;
+                }
+                let Some(next) = step.checked_mul(2) else {
```
yihozhang, line 1051: Nit: we can get rid of the checked_add/checked_mul if we assert that target_cursor is within u32?

core-relations/src/free_join/execute.rs
```diff
@@ -1023,6 +1023,61 @@ fn intersect_with_dense_ref<'a>(v: SubsetRef<'a>, range: OffsetRange) -> Option<
...
+                step = next;
+            }
+
+            let previous = step / 2;
+            let mut lo = base
+                .saturating_add(previous)
+                .saturating_add(1)
+                .min(target_len);
+            let mut hi = base.saturating_add(step).saturating_add(1).min(target_len);
+            while lo < hi {
```
yihozhang, line 1064: nit: we can use std's binary_search?

core-relations/src/free_join/execute.rs
```diff
@@ -1034,6 +1184,31 @@ enum ProbeIndex<'ctx, 'rows, 'exec> {
     Packed(PackedProbe<'ctx, 'rows, 'exec>),
 }
 
+#[derive(Clone, Copy)]
+enum SortedScalarProbe<'a, 'exec> {
```
yihozhang, line 1188: nit: document what this does and its relationship to `ProbeIndex` (which I have not reviewed yet)

core-relations/src/free_join/execute.rs
```diff
@@ -359,6 +359,105 @@ fn top_index_shape_is_eligible(
         && nonempty_shards >= workers
 }
 
+const TOP_COVER_PARTITIONS_PER_WORKER: usize = 4;
```
yihozhang, line 362: I think this used to be part of #962

yihozhang (commented):
Only reviewed galloping related code and not the new `Prober` struct integration (which requires me to first understand the `execute.rs` refactor).

PR #967, opened 2026-07-30
Add demand-driven coarse join scheduling

ezrosent:
Stacked on #959.

This isolates the demand-driven coarse scheduling change from the packed-executor integration:

- schedule coarse join work when it becomes available
- preserve the existing join results and packed representation
- keep the scheduling policy independently reviewable

Validation:

- cargo fmt --check


ezrosent:
This one is mostly about handling skew. It helps a bit on my laptop but theoretically helps more at higher core counts. We'll want to benchmark to make sure.

PR #968, opened 2026-07-30
Stabilize dynamic generic-join ordering

ezrosent:
Stacked on #954.

This isolates the dynamic variable-ordering fixes ahead of the packed-trie work:

- keep a stable logical prefix for mixed and fused joins
- refine only the physical frontier for pure intersections
- use deterministic size, stable-order, and relation tie-breaks
- add focused regression coverage for the ordering policy

The small explicit hash-code cast commit is a prerequisite for compiling the current #954 base with Rust 1.91.

Validation:

- cargo fmt --check

### DIFF
--- a/core-relations/src/free_join/execute.rs
+++ b/core-relations/src/free_join/execute.rs
@@ -3353,6 +3353,23 @@ fn num_intersected_rels(join_stage: &JoinStage) -> i32 {
     }
 }
 
+fn is_reorder_barrier(stage: &JoinStage) -> bool {
+    matches!(
+        stage,
+        JoinStage::FusedIntersectMat {
+            mode: MatScanMode::Lookup(_) | MatScanMode::Value(_) | MatScanMode::Full,
+            ..
+        }
+    )
+}
+
+/// Choose the physical execution order for a suffix of an already-planned
+/// stage block.
+///
+/// The cached logical order fixes stage identity and provides a stable
+/// refinement anchor. At execution time, all-`Intersect` blocks additionally
+/// follow the branch-local physical frontier; mixed and fused blocks retain
+/// the stable-prefix bias to avoid producing many divergent probe paths.
 fn sort_plan_by_size(
     order: &mut InstrOrder,
     leaf_scans: &mut LeafScans,
@@ -3362,14 +3379,7 @@ fn sort_plan_by_size(
 ) {
     let mut last_pos = start;
     for i in start..instrs.len() {
-        if matches!(
-            &instrs[order.get(i)],
-            // These nodes don't commute
-            JoinStage::FusedIntersectMat {
-                mode: MatScanMode::Lookup(_) | MatScanMode::Value(_) | MatScanMode::Full,
-                ..
-            }
-        ) {
+        if is_reorder_barrier(&instrs[order.get(i)]) {
             sort_plan_by_size_inner(order, last_pos..i, instrs, binding_info);
             last_pos = i + 1;
         }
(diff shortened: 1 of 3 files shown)


core-relations/src/free_join/execute.rs
```diff
@@ -3478,98 +3488,93 @@ fn sort_plan_by_size_inner(
...
+    }
+    // Pure intersection blocks also track the stages actually selected on this
+    // branch, keeping recursive ordering connected to its physical frontier.
+    if let Some(physical_refinements) = &mut physical_refinements {
+        for position in 0..range.start {
+            update_refinements(&instrs[order.get(position)], physical_refinements);
         }
     }
 
-    // We prioritize variables by
```
yihozhang, line 3509: keep this comment

Edit: This PR is proposing a new sort criterion. We need to document it and briefly talk about the intuition.

core-relations/src/free_join/execute.rs
```diff
@@ -3478,98 +3488,93 @@ fn sort_plan_by_size_inner(
     }
     // How many times an atom has been intersected/joined
     let mut times_refined = with_pool_set(|ps| ps.get::<DenseIdMap<AtomId, i64>>());
-
-    // Count how many times each atom has been refined so far.
-    for position in 0..range.start {
-        match &instrs[order.get(position)] {
+    let use_physical_refinements = instrs
```
yihozhang, line 3491: Please describe in a comment what use_physical_refinement is.

core-relations/src/free_join/execute.rs
```diff
@@ -3478,98 +3488,93 @@ fn sort_plan_by_size_inner(
...
+        for j in i..range.end {
+            let stage = &instrs[order.get(j)];
+            let stable_refinement = refinement_count(stage, &times_refined);
+            let size = estimate_size(stage, binding_info);
+            let relations = num_intersected_rels(stage);
+            let key = if let Some(physical_refinements) = &physical_refinements {
+                (
+                    -refinement_count(stage, physical_refinements),
+                    size,
+                    -stable_refinement,
```
yihozhang, line 3559: What is a stable refinement? I doubt this criterion matters for most benchmarks, since it's rather unlikely that two possible stages would already have the same size.

yihozhang (commented):
Left some comments for your agent!

core-relations/src/free_join/execute.rs
```diff
@@ -3478,98 +3488,93 @@ fn sort_plan_by_size_inner(
...
                 to_intersect.iter().for_each(|(spec, _)| {
-                    *times_refined.get_or_default(spec.to_index.atom) +=
+                    *refinements.get_or_default(spec.to_index.atom) +=
                         spec.to_index.vars.len() as i64;
                 });
             }
+        };
+
+    // The cached logical prefix remains the stable locality signal.
+    for stage in &instrs[..range.start] {
```
yihozhang, line 3523: Why in other places we use `instrs[order.get(...)]` while here we use `instrs[..]`?

yihozhang (commented):
I also wonder if we should just shuffle instrs in place (instead of shuffling the permutation `order` over it), as this has a slightly slightly more local/predictable access pattern.

Locally, comparing against the base `shared-trie` (two PRs back)

```

  hardboiled_conv1d_32.egg                        0.117       0.114   -0.002    -2.1%  ▼ faster
  hardboiled_conv1d_128.egg                       0.286       0.284   -0.002    -0.7%  ▼ faster
  luminal-llama.egg                               0.082       0.083  +  0.001  +   1.2%  ▲ slower
  python_array_optimize.egg                       0.228       0.230  +  0.002  +   0.8%  ▲ slower
  cykjson.egg                                     0.034       0.032   -0.002    -6.3%  ▼ faster
  eggcc-extraction.egg                            0.200       0.209  +  0.008  +   4.2%  ▲ slower
  llama.egg                                       0.370       0.377  +  0.006  +   1.8%  ▲ slower
  paged_llama.egg                                 1.667       1.682  +  0.015  +   0.9%  ▲ slower
  qwen.egg                                        0.441       0.419   -0.021    -4.8%  ▼ faster
  qwen3_moe.egg                                   0.574       0.554   -0.020    -3.5%  ▼ faster
  whisper.egg                                     1.274       1.179   -0.095    -7.4%  ▼ faster

  Summary: 6 faster  ·  5 slower  ·  0 unchanged  ·  0 missing
  Overall average Δ: -1.46%
```

Comparing against #954 

```

  hardboiled_conv1d_32.egg                        0.111       0.113  +  0.002  +   1.9%  ▲ slower
  hardboiled_conv1d_128.egg                       0.280       0.296  +  0.016  +   5.6%  ▲ slower
  luminal-llama.egg                               0.101       0.083   -0.018   -17.9%  ▼ faster
  python_array_optimize.egg                       0.234       0.239  +  0.006  +   2.5%  ▲ slower
  cykjson.egg                                     0.031       0.032  +  0.001  +   4.4%  ▲ slower
  eggcc-extraction.egg                            0.208       0.218  +  0.009  +   4.5%  ▲ slower
  llama.egg                                       0.365       0.373  +  0.008  +   2.1%  ▲ slower
  paged_llama.egg                                 1.653       1.689  +  0.037  +   2.2%  ▲ slower
  qwen.egg                                        0.420       0.409   -0.011    -2.6%  ▼ faster
  qwen3_moe.egg                                   0.553       0.558  +  0.005  +   0.9%  ▲ slower
  whisper.egg                                     1.166       1.118   -0.048    -4.1%  ▼ faster

  Summary: 3 faster  ·  8 slower  ·  0 unchanged  ·  0 missing
  Overall average Δ: -0.05%
```

so looks like real but acceptable overhead. Though I'm wondering if we can eliminate it by simply reverting to the old heuristic code.

ezrosent:
> so looks like real but acceptable overhead. Though I'm wondering if we can eliminate it by simply reverting to the old heuristic code.

Yeah I think you're right. This is too clever by half. I think we should probably get rid of this PR. Modify the previous one to restore the original heuristic. And then probably remove the fancy packed trie stuff.

That way, we'll have less moving parts and probably just about the same performance overall. Sound good?

ezrosent:
Superseded by restoring the pre-#954 logical-order DVO heuristic in #954 and rebasing #969 directly onto it.

core-relations/src/free_join/execute.rs
ezrosent: This proposed sort criterion has been removed. PR #968 is closed as superseded, and #954 now restores the original logical-order DVO heuristic instead.

core-relations/src/free_join/execute.rs
ezrosent: `use_physical_refinements` and the physical-prefix policy have been removed. PR #968 is superseded by the original DVO restoration in #954.

core-relations/src/free_join/execute.rs
ezrosent: The stable-versus-physical refinement split has been removed rather than retained. #954 now uses the original logical-prefix criterion.

core-relations/src/free_join/execute.rs
ezrosent: The inconsistent physical/logical-prefix experiment has been removed. #954 restores the original `instrs[..range.start]` logical-prefix behavior.

PR #969, opened 2026-07-30
Add packed join node and positioned-index abstractions

ezrosent:
Stacked on the DVO policy PR.

This introduces the reusable pieces of the packed generic-join design without switching the executor over:

- arena-packed trie nodes and cursors
- stable positioned-index APIs
- compatibility wrappers that keep the existing executor behavior unchanged

The executor integration is split across #993–#1000, which replace the original combined PR #959 in the main stack.

Validation:

- cargo fmt --check

### DIFF
--- a/core-relations/src/free_join/packed_trie.rs
+++ b/core-relations/src/free_join/packed_trie.rs
@@ -2,14 +2,12 @@
 //!
 //! A node is one allocation containing an immutable sorted column index and,
 //! when another level follows it, one child publication slot per distinct key.
-//! The allocation is tied to a [`Handle`] and deliberately has no destructor:
-//! its only non-`Copy` trailing values are `OnceLock`s containing references
-//! into the same execution-scoped arena.
+//! The raw allocation is tied to a [`Handle`] and never runs destructors.
 
 use std::{
     alloc::Layout,
     marker::PhantomData,
-    mem::{ManuallyDrop, align_of, size_of},
+    mem::{align_of, size_of},
     ptr,
     sync::{
         OnceLock,
@@ -20,17 +18,71 @@ use std::{
 use egglog_concurrency::Handle;
 
 use crate::{
-    OffsetRange, SubsetRef, Value,
-    numeric_id::NumericId,
+    SubsetRef, Value,
     offsets::{RowId, SortedOffsetSlice},
     table_spec::{ColumnId, WrappedTableRef},
 };
 
-const HAS_CHILDREN: u32 = 1 << 31;
-const DYNAMIC_CHILDREN: u32 = 1 << 30;
-const KEY_LEN_MASK: u32 = DYNAMIC_CHILDREN - 1;
+/// A packed distinct-key count and description of the trailing child storage.
+///
+/// The low 30 bits store the key count. Bit 30 selects dynamic child families,
+/// while bit 31 records that the node has children at all.
+#[repr(transparent)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+struct PackedKeyLen(u32);
+
+impl PackedKeyLen {
+    const HAS_CHILDREN: u32 = 1 << 31;
+    const DYNAMIC_CHILDREN: u32 = 1 << 30;
+    const LEN_MASK: u32 = Self::DYNAMIC_CHILDREN - 1;
+
+    fn new(key_len: usize, child_shape: ChildShape) -> Self {
+        assert!(
+            key_len <= Self::LEN_MASK as usize,
+            "a packed trie node cannot contain 2^30 or more keys"
+        );
+        let flags = match child_shape {
+            ChildShape::Leaf => 0,
+            ChildShape::Direct => Self::HAS_CHILDREN,
+            ChildShape::Dynamic { families } => {
+                assert!(
+                    families > 0,
+                    "a dynamic packed trie node needs at least one child family"
+                );
+                assert!(
+                    u32::try_from(families).is_ok(),
+                    "a packed trie node cannot contain more than u32::MAX child families"
+                );
+                Self::HAS_CHILDREN | Self::DYNAMIC_CHILDREN
+            }
+        };
+        Self(key_len as u32 | flags)
+    }
+
+    fn len(self) -> usize {
+        (self.0 & Self::LEN_MASK) as usize
+    }
+
+    fn has_children(self) -> bool {
+        self.0 & Self::HAS_CHILDREN != 0
+    }
+
+    fn has_dynamic_children(self) -> bool {
+        self.0 & Self::DYNAMIC_CHILDREN != 0
+    }
+}
 
 type ChildSlot<'exec> = OnceLock<&'exec PackedTrieNode<'exec>>;
+
+/// The child-publication storage trailing a packed trie node.
+///
+/// Returning slices from one accessor centralizes the unsafe offset arithmetic
+/// for both inline direct slots and lazily allocated dynamic families.
+enum PackedChildren<'node, 'exec> {
+    Leaf,
+    Direct(&'node [ChildSlot<'exec>]),
+    Families(&'node [AtomicPtr<ChildSlot<'exec>>]),
+}
 
 /// How one packed trie level reaches its children.
 ///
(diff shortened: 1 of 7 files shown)


core-relations/src/free_join/packed_trie.rs
```diff
@@ -0,0 +1,1294 @@
...
+}
+
+/// One immutable, scalar trie level allocated for the lifetime of an execution.
+///
+/// The dynamically sized portions immediately following this header are laid
+/// out as keys, boundaries, grouped row ids, and (optionally) child locks. Use
+/// the accessors rather than relying on those offsets outside this module.
+#[repr(C)]
+#[derive(Debug)]
+pub(crate) struct PackedTrieNode<'exec> {
```
yihozhang, line 141: Maybe use some ASCII art to visualize how a PackedTrieNode is laid out in the arena memory.

ezrosent: Added an ASCII layout for the common packed index, direct child slots, and dynamic child-family pointer table, including the separate lazy family allocation and alignment padding.

core-relations/src/free_join/packed_trie.rs
```diff
@@ -0,0 +1,1327 @@
...
+/// | AtomicPtr families[F]      |-------------+--> ChildSlot family[K]
+/// +----------------------------+
+/// ```
+///
+/// Alignment may insert padding between regions. Use the accessors rather
+/// than relying on these offsets outside this module.
+#[repr(C)]
+#[derive(Debug)]
+pub(crate) struct PackedTrieNode<'exec> {
+    key_len_and_flags: u32,
```
yihozhang: nit: make this a wrapped u32 with access method or document the state it represents. I lean towards the latter.

core-relations/src/free_join/packed_trie.rs
```diff
@@ -0,0 +1,1327 @@
...
+        if rows
+            .windows(2)
+            .all(|pair| pair[0].index().checked_add(1) == Some(pair[1].index()))
+        {
+            SubsetRef::Dense(OffsetRange::new(first, last.inc()))
+        } else {
+            // SAFETY: the constructor requires `(Value, RowId)` order, so the
+            // rows within each equal-value range are non-decreasing.
+            SubsetRef::Sparse(unsafe { SortedOffsetSlice::new_unchecked(rows) })
+        }
```
yihozhang: Do we need this?

yihozhang: If this turns out to be useful, it's still better to do this on the build side, I think (i.e., have the current RowID array support both Sparse and Dense).

core-relations/src/free_join/packed_trie.rs
```diff
@@ -0,0 +1,1327 @@
...
+    fn child_slot(
+        &self,
+        arena: &Handle<'exec>,
+        key_index: usize,
+        family: usize,
+    ) -> &ChildSlot<'exec> {
+        assert!(key_index < self.key_len(), "packed trie key out of bounds");
+        match self.child_shape() {
+            ChildShape::Leaf => panic!("cannot build a child index below a packed trie leaf"),
+            ChildShape::Direct => {
```
yihozhang: Should we panic in this case? Is this code path ever exercised?

core-relations/src/free_join/packed_trie.rs
```diff
@@ -0,0 +1,1327 @@
...
+    /// family array when this is its first use.
+    ///
+    fn child_slot(
+        &self,
+        arena: &Handle<'exec>,
+        key_index: usize,
+        family: usize,
+    ) -> &ChildSlot<'exec> {
+        assert!(key_index < self.key_len(), "packed trie key out of bounds");
+        match self.child_shape() {
```
yihozhang: Proposal: we make a method on PackedTrieNode that returns an enum of `Leaf | Direct(&[ChildSlot]) | Families(&[AtomicPtr])`. Basically a generalized child_shape. This could simplify places that pattern match on `child_shape()` and centralize `ChildSlot` addressing.

core-relations/src/free_join/packed_trie.rs
```diff
@@ -0,0 +1,1327 @@
...
+    numeric_id::NumericId,
+    offsets::{RowId, SortedOffsetSlice},
+    table_spec::{ColumnId, WrappedTableRef},
+};
+
+const HAS_CHILDREN: u32 = 1 << 31;
+const DYNAMIC_CHILDREN: u32 = 1 << 30;
+const KEY_LEN_MASK: u32 = DYNAMIC_CHILDREN - 1;
+
+type ChildSlot<'exec> = OnceLock<&'exec PackedTrieNode<'exec>>;
```
yihozhang, line 75: Is it an issue that we assume these `OnceLock`s don't drop? Will their `drop` ever be called?

core-relations/src/free_join/packed_trie.rs
```diff
@@ -0,0 +1,1327 @@
...
+/// ```text
+/// +----------------------------+
+/// | PackedTrieNode header      |
+/// +----------------------------+
+/// | Value keys[K]              |
+/// +----------------------------+
+/// | u32 boundaries[K + 1]      |
+/// +----------------------------+
+/// | RowId rows[R]              |
+/// +----------------------------+
```
yihozhang, line 115: Maybe not for this PR, but from my experience, even with DVO which let tries have multiple access patterns, most of tries only have one access pattern throughout the execution. Can we always keep one `direct[K]` in place even for the dynamic case? This could avoid a separate allocation for the majority of the cases. But of course this will make the code more complicated, and maybe not worth it because indices that have children live higher up so there aren't a lot of them.

yihozhang (commented):
After some poking around it looks pretty reasonable to me.

core-relations/src/free_join/packed_trie.rs
ezrosent: By 'this' do you mean special-casing 'Dense'? If so... yeah I think we can leave it out.

core-relations/src/free_join/packed_trie.rs
ezrosent: OnceLock doesn't drop if its inner value doesn't drop. I think we can add a test that asserts this though.

EDIT: turns out it _has_ a drop method unconditionally but it just calls drop on the inner `T` if it's initialized. In this case, since `T` is a reference, drop has no effect.

core-relations/src/free_join/packed_trie.rs
ezrosent: This is a good idea!! Agree that it should land in a follow-up though.

core-relations/src/free_join/packed_trie.rs
ezrosent: Simplified a bunch of stuff here. In general `mem::forget`ing is safe in Rust. And in this case it's completely benign and doesn't leak any resources.

core-relations/src/free_join/packed_trie.rs
yihozhang: Another overlapping idea: We can support inlining the last two layers of the trie (in place of the `Direct` variant). Because the last layer of the trie is usually very small (e.g., a singleton), and they are known to be a leaf, this could save a lot of allocations.

P.s. Also, this could potentially subsume the SmallVec indexes we currently have

core-relations/src/free_join/packed_trie.rs
ezrosent: Done. This is the new PackedChildren type.

core-relations/src/free_join/packed_trie.rs
```diff
@@ -548,36 +627,23 @@ impl<'exec> PackedTrieNode<'exec> {
...
+                unreachable!("a packed trie leaf cannot publish a child index")
+            }
+            PackedChildren::Direct(slots) => {
                 // Direct nodes have exactly one deterministic successor. The
                 // executor passes its plan-local AccessId uniformly as
                 // `family`; it is intentionally ignored in this shape.
-                return self
-                    .direct_child_slot(key_index)
-                    .expect("direct packed trie node must have child slots");
+                return &slots[key_index];
```
yihozhang, line 638: Had the same question about whether we should panic here, and whether this code path is ever exercised in the follow-up PR?

yihozhang (approved):

Issue #911, opened 2026-06-03
Back-off scheduler trips hash-index sortedness invariant (SortedOffsetSlice::new_unchecked "slice is not sorted", debug)


oflatt:
### Summary

Running a congruence-heavy ruleset under a back-off scheduler panics in **debug** builds:

```
thread 'main' panicked at core-relations/src/offsets/mod.rs:
slice is not sorted: [RowId(3655), RowId(4274), RowId(3655), RowId(4274)]
```

Backtrace path:

```
core-relations::offsets::SortedOffsetSlice::new_unchecked
core-relations::hash_index::...::make_ref
egglog::scheduler::<impl EGraph>::step_rules_with_scheduler
```

Release builds compile out the `debug_assert!`, so they don't crash — but they then use the unsorted/duplicated subset as if it were sorted, which can yield mis-ordered or duplicated iteration in the scheduler path.

### Reproducer

Needs [egglog-experimental](https://github.com/egraphs-good/egglog-experimental) for the `back-off` scheduler. 9 lines, debug build:

```scheme
(datatype Math (Num i64) (Add Math Math) (Mul Math Math))
(rewrite (Add a b) (Add b a))
(rewrite (Mul a b) (Mul b a))
(rewrite (Add a (Add b c)) (Add (Add a b) c))
(rewrite (Mul a (Mul b c)) (Mul (Mul a b) c))
(rewrite (Mul a (Add b c)) (Add (Mul a b) (Mul a c)))
(rewrite (Add (Mul a b) (Mul a c)) (Mul a (Add b c)))
(let e (Mul (Add (Num 1) (Add (Num 2) (Num 3))) (Add (Num 4) (Add (Num 5) (Num 6)))))
(run-schedule (let-scheduler bo (back-off)) (repeat 50 (run-with bo)))
```

It's near-minimal — dropping the factoring rule or the associativity rules makes it stop triggering (the match explosion that drives back-off to ban rules is needed).

### Root cause

[...]

PR #973, opened 2026-08-03
Fix spurious `slice is not sorted` assert in `Subset::intersect` (#971)

### LINKED ISSUE #971: Bug: debug_assert failure: slice is not sorted
This looks like an internal error.

Reproduction (far from minimal): [code](https://github.com/ProteusLab/LIRA/tree/egglog_repro)

```
> (cd rust && cargo test -- repro)

thread 'repro' (1345333) panicked at /home/dan/.cargo/git/checkouts/egglog-7c9491f597880af0/53b9721/core-relations/src/offsets/mod.rs:136:9:
slice is not sorted: [RowId(13), RowId(1), RowId(2), RowId(3), RowId(4), RowId(5), RowId(11), RowId(12), RowId(13), RowId(14), RowId(15), RowId(16), RowId(17), RowId(18), RowId(19), RowId(20), RowId(21)]
```

UPD: I do know about #911 and #914, but this is on the latest commit (53b9721 in panic message)

yihozhang:
Fixes #971.

## Problem

`Subset::intersect`'s `(Sparse, Sparse)` "other is much smaller: iterate other and gallop in cur" branch compacts the intersection **into `cur` in place**, but re-derived a `SortedOffsetSlice` over the *whole* vector on every iteration:

```rust
let result = cur.slice().scan_for_offset(ci, target);   // asserts all of cur is sorted
match result {
    Ok(found) => { cur.0[write] = target; write += 1; ci = found + 1; }
```

Once the first match is written to `cur.0[write]`, the region between `write` and `ci` still holds stale originals, so `cur` as a whole is transiently unsorted and `slice()`'s `debug_assert!` fires on the *next* iteration.

That is exactly the panic in #971. `cur` was `[0,1,2,3,4,5,11..=21]` (17 rows) and the first match was `RowId(13)`, written over index 0:

```
[RowId(13), RowId(1), ..., RowId(5), RowId(11), RowId(12), RowId(13), ..., RowId(21)]
 ^^^^^^^^^ written result   ^^^^^^^^^^^^^^^^^^ untouched original tail
```

Despite the identical assertion message, this is a **different bug from #911/#914** — that one was in `hash_index`; this one is in `offsets`, reached via `Database::process_constraints`.

## This was only the assertion, not the results

`write <= ci` always holds (a match at `found >= ci` writes to `write <= found`, then sets `ci = found + 1`) and `scan_for_offset` never reads below its `start`, so every read already landed in the untouched suffix. Confirmed empirically: with the assertion compiled out, the pre-fix code produces correct intersections. So release builds were unaffected; debug builds panicked.

## Fix

Search `cur.0[ci..]` — the suffix that is guaranteed untouched — instead of all of `cur`, and make the returned offsets absolute again. `scan_for_offset`'s duplicate back-walk floors at `start`, which becomes suffix-relative `0`, so the semantics are unchanged. Same complexity, no extra work in the hot path.

## Testing

The existing `offsets::tests::intersect` corpus missed this because `add_row_sorted` collapses contiguous row sets to `Dense`, so its skewed pairs never put *two* `Sparse` subsets into this branch. Added:

- `offsets::tests::intersect_sparse_sparse_skewed`, a targeted regression test that reproduces #971's slice byte-for-byte in ~20 lines with no external dependencies.
- Two entries (a 17-row sparse subset and a 3-row sparse subset) to the exhaustive `intersect` corpus, so all-pairs coverage now reaches this branch.

Verified at `53b9721`:

- `cargo test --workspace`: **1132 passed, 0 failed** (including the 792-test integration suite).
- The reporter's reproducer in [ProteusLab/LIRA@egglog_repro](https://github.com/ProteusLab/LIRA/tree/egglog_repro) — `cd rust && cargo test -- repro` — passes against the patched egglog, as does the rest of that suite.
- `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean.

I also audited the other `SortedOffsetVector::slice()` / `scan_for_offset` call sites; this was the only one whose receiver can be mid-mutation. One adjacent fragility, left alone as it is not a live bug: `binary_search_from`'s duplicate back-walk is bounded by `found > 0` rather than `found > start`, so it can return an index below `start`; at the sole `start != 0` call site the values make that unreachable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,6 @@
 - Report full source file paths in egglog span and error messages.
 - Fix seminaive matching after nested containers rebuild in place by propagating dirty container ids through parent containers.
 - Fix multi-column secondary index rebuilds so each value's rows come back sorted by row id, and make all rebuild paths (serial, parallel, and bulk) record a row once even when its value repeats across covered columns (#914).
-- Fix a spurious `slice is not sorted` debug assertion in `Subset::intersect` when galloping through a sparse subset that is being compacted in place, which panicked debug builds on some sparse/sparse intersections (#971).
 - Render nullary AST calls without a trailing space, e.g. (foo) instead of (foo ).
 - Escape `"` and `\` when displaying string literals so printed/serialized programs round-trip through the parser.
 - Add a BigRat to-i64 primitive for integral rationals.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Report full source file paths in egglog span and error messages.
 - Fix seminaive matching after nested containers rebuild in place by propagating dirty container ids through parent containers.
 - Fix multi-column secondary index rebuilds so each value's rows come back sorted by row id, and make all rebuild paths (serial, parallel, and bulk) record a row once even when its value repeats across covered columns (#914).
+- Fix a spurious `slice is not sorted` debug assertion in `Subset::intersect` when galloping through a sparse subset that is being compacted in place, which panicked debug builds on some sparse/sparse intersections (#971).
 - Render nullary AST calls without a trailing space, e.g. (foo) instead of (foo ).
 - Escape `"` and `\` when displaying string literals so printed/serialized programs round-trip through the parser.
 - Add a BigRat to-i64 primitive for integral rationals.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -18,4 +18,4 @@ When you edit the files, make sure to respect the following:
 - When running tests, always use the `--release` mode. Alternatively, you can also run `make test`.
 - If your change is performance-critical, use `script/bench.py` as the ground truth to evaluate the performance impact.
 - Keep your documentation concise and avoid duplicate information. The `tidy-diff-docs` skill (`.claude/skills/tidy-diff-docs/`) cleans the doc and code comments in a diff down to the caller-facing contract.
-- Update CHANGELOG.md with a concise bullet when you make major changes (e.g., breaking changes or new features added) in the codebase.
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Report full source file paths in egglog span and error messages.
 - Fix seminaive matching after nested containers rebuild in place by propagating dirty container ids through parent containers.
 - Fix multi-column secondary index rebuilds so each value's rows come back sorted by row id, and make all rebuild paths (serial, parallel, and bulk) record a row once even when its value repeats across covered columns (#914).
+- Fix a spurious `slice is not sorted` debug assertion in `Subset::intersect` when galloping through a sparse subset that is being compacted in place, which panicked debug builds on some sparse/sparse intersections (#971).
```
oflatt: Probably too much detail, we should try to keep changelog to just new features / breaking changes

oflatt (approved):

CHANGELOG.md
saulshanabrook: If this is a bug that is fixed good to have a changelog entry for it, just maybe would be nice if it's a little bit more descriptive on what the bug is

oflatt: It's a bug introduced since last release right? So not relevant to people following egglog releases?

yihozhang: I removed this CHANGELOG entry and updated CLAUDE.md. There are a few other entries (both "unreleased" and pre-LLM) that discuss bug fixes, though. Maybe a good criterion is whether this bug was introduced in a previous release.

saulshanabrook: Ah gotchya yeah I forgot we hadn't released in a while

PR #980, opened 2026-08-05
Move Map containers to sequence storage

ezrosent:
## Summary

- Encode maps as `[rebuild mask, sorted key/value pairs...]`.
- Implement lookup and editing directly over the canonical sequence using binary search and slice splicing.
- Preserve selective key/value rebuilding and last-old-key-wins behavior when rebuilt keys collide.
- Cover codec round trips, rebuild masks, collisions, predictions, and last-write-wins updates.

## Stack

Draft, stacked on #979. This is PR 7 of 12 in the sequence-backed container stack.

## Validation

Validated on the complete 12-commit stack:

- `cargo nextest run --workspace`: 1,362 tests passed
- `cargo test --workspace --doc`
- `cargo clippy --workspace --all-targets -- -D warnings`

### DIFF
--- a/src/sort/map.rs
+++ b/src/sort/map.rs
@@ -1,5 +1,9 @@
 use super::*;
+use crate::numeric_id::NumericId;
 use std::collections::BTreeMap;
+
+const REBUILD_KEYS: usize = 1;
+const REBUILD_VALS: usize = 1 << 1;
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct MapContainer {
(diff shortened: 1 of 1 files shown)


src/sort/map.rs
```diff
@@ -279,6 +492,230 @@ impl ContainerSort for MapSort {
...
+}
+
+#[derive(Clone, Copy)]
+enum MapEditOp {
+    Insert,
+    Remove,
+}
+
+/// Map updates splice the borrowed alternating sequence directly and only
+/// round-trip through `MapContainer` for a legacy backend.
```
saulshanabrook, line 626: What does "legacy backend" mean in this context? If this changes things, would prefer to break everything and not preserve any fallback paths for simplicity of maintenance/review

PR #986, opened 2026-08-06
Store a function's signature once, and expose it to primitives

oflatt-claude:
Three additions to what a primitive body can see. None are specific to any one
extension: together they are what an out-of-tree primitive needs to walk the
term structure under an e-class and build a modified copy of it.

The motivating consumer is an `unstable-subst` substitution primitive in
egraphs-good/egglog-experimental#60, which lives there rather
than here because substitution is a language experiment, not core e-graph
machinery. This PR is only the introspection; it is worth reviewing on its own
terms.

### `Read::enodes_for_eclass(name, eclass, f)`

Walks a constructor's rows by output e-class through the backend's lazy column
index, instead of scanning the table and filtering. Cherry-picked from #934,
along with the `core-relations` `ExecutionState::for_each_matching_col` and
`egglog-bridge` `TableAction::for_each_output_value` it rests on — credit to
that PR; this takes only the traversal hunks and none of the extraction work.

### `Read::constructor_schema(name)`, `Read::function_schema(name)`, `Read::table_subtype(name)`

A table's declared signature, and whether it is a constructor or a function.
`EGraph::functions_iter` already exposes this from `&EGraph`, but a primitive
body only ever sees a state wrapper, and those carried no sort information at
all — so a primitive could read rows without being able to tell an e-class
column from a base value, or a constructor's eclass column from a function's
output.

The schema accessor is split by subtype and errors with `WrongSubtype` on a
mismatch, matching how the rest of `Read` and `Write` already work
(`lookup`/`eclass_of`, `constructor_enodes`/`function_entries`, `set`/`add`).
That split is not just cosmetic: a constructor's last column is an e-class and
a function's is an output. `table_subtype` is the error-free predicate for code
that accepts either.

### **Breaking:** a signature is now stored once

The first cut of this branch added a `FunctionSchemas` registry to back those
accessors. That was a third copy of data egglog already kept twice —
`TypeInfo::func_types` holds a `FuncType {name, subtype, input, output}`, and
`Function` held the same content again as a `ResolvedSchema` plus
`decl.subtype` (and a third time, unresolved, in `decl.schema`). What was
missing was never the data, only a way to reach it from a state wrapper.

So `TypeInfo::func_types` became the single store and the shared cell, and the
new registry is gone:

- `TypeInfo::get_func_type` returns `Option<Arc<FuncType>>`, not
  `Option<&FuncType>`.
- `Function` points at the same `Arc<FuncType>`;
  `Function::schema() -> &ResolvedSchema` becomes
  `Function::func_type() -> &FuncType`.
- `ResolvedSchema` is removed; its `get_by_pos` moves to `FuncType`.
- `declare_function` reuses the signature typechecking already resolved instead
  of resolving the sorts a second time. It still resolves and records the
  functions desugaring generates (global bindings, proof tables), which never
  go through typechecking — so `func_types` now covers every declared table.
- `GenericFunctionDecl::resolved_schema` is gone, along with the placeholder
  `String::new()` an unresolved decl carried in it and its
  `ResolvedCall::view_types` reader. Its only consumer was proof encoding, which
  runs over decls desugaring generates *after* typechecking (the functions
  global bindings lower to) and so cannot look those up by name — but it never
  needed a `FuncType`, only the column *sorts*, and those resolve by name from
  `TypeInfo` like any other sort. (Recording the generated signatures into
  `TypeInfo` instead does not work: the desugared program is typechecked again
  afterwards, so recording early trips the already-bound check.)
- `ResolvedCall::Func` holds an `Arc<FuncType>`, so a resolved call site shares
  the signature rather than cloning it — and there is one of those per call in
  every rule. Patterns that matched
  `ResolvedCall::Func(FuncType { subtype: Custom, .. })` cannot look through an
  `Arc`, so they match the variant and test the subtype in a guard via a new
  `ResolvedCall::is_custom_func`.

Two subtleties worth a reviewer's eye. `TypeInfo::clone` deep-copies the map: a
clone is an independent e-graph — a `push`ed copy, or the parallel typechecking
the proof checker keeps in `original_typechecking` — and declaring a function in
one must not make it resolve in the other. And `pop` restores the pushed
contents into the live cell the registered primitives already hold, rather than
swapping in a cell they have no handle to.

### No lock: the signature reaches a primitive as a borrow

The obvious way to back those accessors is to hand the primitive an
`Arc<RwLock<_>>`, and the first cut of this branch did. That lock would never
have been guarding a race — declaring anything needs `&mut EGraph`, `run_rules`
already holds that borrow for the whole execution, and nothing on
`Core`/`Read`/`Write` can declare. It was there only because external functions
are stored as `Box<dyn ExternalFunction + 'static>`, so a wrapper cannot hold a
borrow of the e-graph.

So this changes the route instead of the synchronization. `ExecutionState`
carries an `ExternalContext` — a borrowed value the caller of an operation
makes visible to the external functions it reaches — and egglog passes
`&TypeInfo` into `run_rules` and `with_execution_state*`. The state wrappers
read it back, so `constructor_schema` hands out a `&FuncType` borrowed from the
e-graph itself.

The invariant is then enforced by the borrow checker rather than stood in for
by a lock: that `&TypeInfo` is live for exactly the operation that supplied it,
so the e-graph cannot be mutably borrowed to declare a function while a rule is
running. `TypeInfo::func_types` stays a plain owned map, `get_func_type` keeps
its `Option<&FuncType>` signature and costs a map lookup, `TypeInfo::clone`
stays derived, and `push`/`pop` restore it like every other field.

Rebuild paths pass `None`: they run generated rebuild rules whose actions
evaluate in the `Write` context, which has no schema accessors.

This does not touch the pre-existing `ActionRegistry` read lock in
`RegistryPrimWrapper::invoke`, which every registry primitive still pays per
invocation — the same seam would remove it, as a follow-up.

`table_subtype` also retires an existing workaround in `egglog-experimental`,
which determines a table's subtype by starting a constructor scan and reading
the error off the subtype check — removed in the companion PR.

### `Core::rebuild_container(type_id, value, remap)`

Remaps a container value's contents and interns the result, over the existing
`ContainerValues::rebuild_val_with`. Out-of-tree code cannot reach this through
`Core::register_container`, which requires naming the container's Rust type —
impossible for an arbitrary container sort.

### Notes

- No behaviour change to existing programs; this is additive.
- `FunctionSchemas` costs one `HashMap` entry per declared table.
- Adds one rule to the repo's `tidy-diff-docs` skill: prefer an import over an
  inline full path for a type.
- Verified: full workspace test suite passes (including the 799-program `files`
  harness), `cargo clippy --tests --workspace` clean, `cargo fmt --check` clean,
  and `cargo doc` adds no new warnings.


src/typechecking.rs
```diff
@@ -152,6 +179,15 @@ impl Hash for FuncType {
...
     }
 }
+/// Every declared function's signature, by name.
+///
+/// Shared (via `Arc<RwLock<_>>`) with the state wrappers, which cannot reach
+/// [`TypeInfo`] from a primitive body and read it through
+/// [`Read::constructor_schema`] / [`Read::function_schema`]. `EGraph`'s
+/// `push`/`pop` snapshot and restore its contents so a popped declaration
+/// stops resolving.
+pub(crate) type SharedFuncTypes = Arc<RwLock<HashMap<String, Arc<FuncType>>>>;
```
saulshanabrook: This seems a bit awkward, having a global like this

src/typechecking.rs
```diff
@@ -1139,12 +1145,23 @@ impl TypeInfo {
     }
 
     pub fn get_func_type(&self, sym: &str) -> Option<&FuncType> {
-        self.func_types.get(sym)
+        self.func_types.get(sym).map(Arc::as_ref)
+    }
+
+    /// The signature as a shared handle, for callers that keep it around.
```
yihozhang:
```suggestion
    /// Returns a shared FuncType of the given name
```

yihozhang: Actually, do we need this function? Can we just use `get_func_type`?

src/lib.rs
```diff
@@ -798,22 +782,53 @@ impl EGraph {
...
+            subtype: decl.subtype,
+            input: decl
+                .schema
+                .input
+                .iter()
+                .map(get_sort)
+                .collect::<Result<Vec<_>, _>>()?,
+            output: get_sort(&decl.schema.output)?,
+        });
+        self.type_info.declare_func_type(func_type.clone());
```
yihozhang: Nit: this function can just be inlined I think since it's only used here and there are other places where `func_types.insert` is directly called.

core-relations/src/action/mod.rs
```diff
@@ -483,6 +503,63 @@ impl<'a> ExecutionState<'a> {
...
+        }
+
+        let imp = &table_info.table;
+        let cols: SmallVec<[_; 8]> = (0..imp.spec().arity()).map(ColumnId::from_usize).collect();
+        let mut cur = Offset::new(0);
+        let mut buf = TaggedRowBuffer::new_inline(imp.spec().arity());
+
+        macro_rules! drain_buf {
+            ($buf:expr) => {
+                for (_, row) in $buf.non_stale() {
```
yihozhang: It seems scan_project already guarantees rows in `buf` is non-stale, so we can just use `iter` here. Can you check if this is true?

core-relations/src/action/mod.rs
```diff
@@ -483,6 +503,63 @@ impl<'a> ExecutionState<'a> {
...
+    /// with the whole row as a value slice.
+    pub fn for_each_matching_col(
+        &self,
+        table: TableId,
+        col: ColumnId,
+        value: Value,
+        mut f: impl FnMut(&[Value]),
+    ) {
+        let table_info = &self.db.table_info[table];
+        let constraint = Constraint::EqConst { col, val: value };
```
yihozhang, line 516: The logic here is a bit over-complicated. I wonder if the following (or some variant of it) is simpler while being correct:

```
       let subset = 
          if Some(subset) = table_info.table.fast_subset(constraint) {
              // if the table is sorted by this column
              subset
          } else if !*table_info
            .spec
            .uncacheable_columns
            .get(col)
            .unwrap_or(&false)
        {
            // if an index is available 
            let index = get_column_index_from_tableinfo(table_info, col);
            match index.get().unwrap().get_subset(&value) {
                Some(s) => {
                    s
                }
                None => {
                    Subset::empty()
                }
            }
        } else {
            // slow path
            table.refine(table.all(), constraint)
        }

       // drain buf ...
```

src/lib.rs
```diff
@@ -1959,21 +1977,23 @@ impl EGraph {
         let table_action = egglog_bridge::TableAction::new(&self.backend, func.backend_id);
 
         if function_type.subtype != FunctionSubtype::Constructor {
-            self.backend.with_execution_state(|es| {
-                for row in parsed_contents.iter() {
-                    table_action.insert(es, row.iter().copied());
-                }
-                Some(unit_val)
-            });
+            self.backend
```
yihozhang: nit: do we need to smuggle the type info here? If so (e.g., the merge function calls for_each_matching_col), can we add a test that exercises this path?

src/lib.rs
```diff
@@ -1959,21 +1977,23 @@ impl EGraph {
...
-            self.backend.with_execution_state(|es| {
-                for row in parsed_contents.iter() {
-                    // Constructor semantics: mint a fresh eclass id for
-                    // each missing key.
-                    table_action.lookup_or_insert(es, row);
-                }
-                Some(unit_val)
-            });
+            self.backend
+                .with_execution_state(Some(&self.type_info), |es| {
```
yihozhang: nit: same here

src/lib.rs
```diff
@@ -2267,9 +2287,10 @@ impl EGraph {
 
     /// Convert from a Rust container type to an egglog value.
     pub fn container_to_value<T: ContainerValue>(&mut self, x: T) -> Value {
-        self.backend.with_execution_state(|state| {
-            self.backend.container_values().register_val::<T>(x, state)
-        })
+        self.backend
+            .with_execution_state(Some(&self.type_info), |state| {
```
yihozhang: Nit: We probably don't need to smuggle type info here. 

Alternatively, we can make a helper function that always passes type_info to with_execution_state, and every function calls that helper..

src/exec_state.rs
```diff
@@ -541,6 +628,41 @@ pub trait Write<'a, 'db: 'a>: Core<'a, 'db> + RegistrySealed<'a, 'db> {
     }
 }
 
+fn subtype_label(subtype: FunctionSubtype) -> &'static str {
```
yihozhang: nit: make this a member method of FunctionSubtype

src/exec_state.rs
```diff
@@ -183,6 +194,29 @@ pub trait Core<'a, 'db: 'a>: Internal<'a, 'db> {
...
+    /// Rebuild the contents of a container value through `remap`,
+    /// returning the interned value of the result (or `value` unchanged
+    /// if it is not a registered container of `type_id`, or if nothing
+    /// inside it was remapped). `type_id` comes from
+    /// [`Sort::value_type`](crate::sort::Sort::value_type).
+    ///
+    /// `remap` sees every value the container stores, including ones of
+    /// sorts it does not care about, so it should return its argument
+    /// unchanged for those.
+    fn rebuild_container(
```
yihozhang: The name `rebuild` is confusing here. Is this like a `map` primitive over a container, and is it by itself not mutating?

This also has the weird behavior that if TypeId and Value do not match, it won't throw an error and instead just returns the value unchanged.

I wonder if we can directly call `ContainerValue::rebuild_contents` without reusing the rebuilding functions.

src/exec_state.rs
```diff
@@ -395,6 +456,32 @@ pub trait Read<'a, 'db: 'a>: Core<'a, 'db> + RegistrySealed<'a, 'db> {
...
+    /// merged into another are matched under the id they store, not their
+    /// canonical one. Errors with `WrongSubtype` if `name` is a function.
+    fn enodes_for_eclass(
+        &self,
+        name: &str,
+        eclass: Value,
+        mut f: impl FnMut(Enode<'_>),
+    ) -> Result<(), Error> {
+        let action = lookup_action(self.registry(), name)?;
+        check_subtype(name, &action, TableKind::Constructor, "constructor")?;
```
yihozhang: Nit: now that we have `subtype_label`, we shouldn't pass the name strings around..

yihozhang (commented):
Great work! Mostly just low-level nits.

src/typechecking.rs
oflatt-claude: Good catch — collapsed to one accessor. `get_func_type` now returns `Option<&Arc<FuncType>>` and `func_type_arc` is gone.

Callers that only read fields are unchanged (auto-deref); the three that need to keep it past the borrow (`ResolvedCall::from_resolution*`, `declare_function`) call `.clone()`, which is an `Arc` clone rather than a `FuncType` clone — which was the only reason for the second accessor. That also subsumes your doc suggestion below, since the return type now says "shared" itself.

oflatt-claude: Applied in spirit rather than literally — the function is gone. `get_func_type` now returns `Option<&Arc<FuncType>>`, so the signature carries the "shared" part that this doc line was adding.

src/lib.rs
oflatt-claude: Done, inlined into `declare_function`.

It only had the one caller left — an earlier revision of this branch also called it from a "record everything desugaring generated" pass, and that pass turned out not to work (the desugared program is typechecked again afterwards, so recording early trips the already-bound check), which left the helper with nothing to justify it.

I kept `TypeInfo::declare_func_type` rather than inlining that one too: `func_types` is private to the `typechecking` module, so `EGraph` cannot insert into it directly the way `typecheck_function` does.

core-relations/src/action/mod.rs
oflatt-claude: Checked, and you are right — for both `Table` impls. `SortedWritesTable::scan_generic_bounded` skips stale entries in its constrained and unconstrained branches alike (the fast path that does not check is guarded by `stale_rows == 0`), and `DisplacedTable` is an append-only displacement log with no stale rows to begin with.

Worth noting it is a *new* invariant: it only became true for the constrained path in `9369a702 Filter stale rows in constrained bounded scans`, which is why these filters exist at all — they were compensating for that bug.

So rather than drop it only here, I wrote the invariant down where it is established (`Table::scan_generic_bounded`, and the `scan_bounded` / `scan_project` wrappers) and dropped all five vestigial filters over scan output: this one, both `drain_buf!` macros, and the two in `scan_table`. The three in `table/mod.rs` stay — those read pending-insert queues, where rows really can be stale.

`tests/repro-stale-rows.egg` is the regression guard now that nothing filters defensively, and its seven harness variants pass.

core-relations/src/action/mod.rs
oflatt-claude: Taken, close to your shape. It now prefers, in order: a sort the table already has (`fast_subset`), then a cacheable column's index, then the constraint applied during the scan — the same order of preference `Database::process_constraints` uses, so the two read the same way now.

Worth flagging that it was not only simpler but a fix: the old version checked `uncacheable_columns` alone, so it built (and refreshed) a column index even when `split_fast_slow` had already answered the query by binary search. Yours skips the index in that case, which is what `process_constraints` does by only indexing constraints that landed in `slow`.

I checked the three branches against the old code and they agree: for a single constraint `split_fast_slow` starts from `all()` and intersects, so `all() ∩ fast_subset == fast_subset` and `all() ∩ index_subset == index_subset`. The `debug_assert!` on `slow` went with it — with one constraint in and one out it could not fail, so it only misled.

src/lib.rs
oflatt-claude: Took the helper suggestion, which answers this and the two above at once. `EGraph` now has private `with_execution_state` / `with_execution_state_tracked` wrappers that always pass `Some(&self.type_info)`, and every egglog-side execution goes through them.

You are right that none of the three sites needs it — neither `input` site nor `container_to_value` can reach a primitive — but with the helper no call site has to make that judgement about a primitive it cannot see, and passing it is never wrong, only sometimes unnecessary.

On the test: there was not one covering the path that does need it, so I added `a_primitive_can_resolve_a_signature_from_inside_a_rule` in `tests/api_introspection.rs`. It registers a `ReadPrim` that calls `constructor_schema` and invokes it from a `:naive` rule, which reaches its execution state through `run_rules` rather than `with_execution_state` — a different entry point, and the one that matters. I checked it actually guards that: passing `None` at the `run_rules` call site makes it fail.

Not the merge-function path you floated, though — a merge expression is typechecked in `Context::Write`, and `WriteState` does not implement `Read`, so a merge function cannot reach these accessors at all. That is why the rebuild and merge paths pass `None`.

src/lib.rs
oflatt-claude: You are right, this one does not need it — the closure only inserts rows and cannot reach a primitive. Handled via the helper suggestion in your third comment: `EGraph::with_execution_state` now always passes the type info, so this site no longer names it. See the reply there for the test.

src/lib.rs
oflatt-claude: Same as above — this site goes through the `EGraph::with_execution_state` helper now.

src/exec_state.rs
oflatt-claude: Done — `FunctionSubtype::label()`. `Display for FunctionSubtype` was already producing those exact strings, so it now delegates to `label()` instead of repeating the match.

src/exec_state.rs
oflatt-claude: Done. `TableKind` got a `label()` too, which lets `check_subtype` derive both labels itself: its `expected_label` parameter is gone along with the inline match it used for the actual one, and the seven call sites no longer pass a string that had to agree with the `TableKind` sitting next to it.

src/exec_state.rs
oflatt-claude: All three points taken.

**The name.** It is a `map` over a container's contents, and it does not mutate: the original container is untouched and the result is interned separately. "Rebuild" collided with egglog's other meaning (canonicalizing after unions), which is where the confusion came from — the underlying `ContainerValues::rebuild_val_with` is named for that. Renamed to `map_container`, and the doc now says outright that the input is not modified.

**The silent pass-through.** Fixed, and the information turned out to be there already and discarded: `ContainerEnv::rebuild_val_with` returns `None` exactly when the value is not a container of that type, and only the outer `ContainerValues::rebuild_val_with` flattened that into `.unwrap_or(value)`. That wrapper returns `Option<Value>` now and `map_container` passes it through, so a `TypeId`/`Value` mismatch is reported rather than looking like "nothing changed". The one other caller (`proof_container_rebuild`) keeps the old behaviour with an explicit `.unwrap_or(value)` and a comment saying why it cannot miss. `tests/api_introspection.rs` covers all three outcomes.

**Calling `rebuild_contents` directly.** That is the one I could not do. `ContainerValue::rebuild_contents` is on the typed trait, so reaching it means naming the concrete container type (`VecContainer`, `MapContainer`, ...). This API exists for callers that only have an `ArcSort` — an out-of-tree primitive walking arbitrary container columns cannot name the type, which is also why `type_id` is a parameter rather than a generic.

yihozhang (approved):
Great! I see it also fixes some other places that use `non_stale()` while an `iter` suffices, which is nice.

PR #990, opened 2026-08-14
Stop check_facts after its first match

saulshanabrook:
## Summary

- stop the temporary `check_facts` rule after its first successful callback
- add one high-cardinality parallel regression that also checks stop-state lifecycle

## Why

`check_facts` only needs to determine whether at least one matching substitution exists, but its temporary rule previously continued invoking the callback for every match. The callback now records success and calls `trigger_early_stop()`, allowing the backend to stop cooperatively after the current batch.

In a local serial 500,000-match benchmark, executed callback lanes fell from 500,000 to 128: 99.9744% fewer, or 3,906x. Clean counterbalanced wall-time runs improved from a 10.033 ms median to 4.805 ms, a 52.1% reduction (2.09x faster); paired reductions ranged from 51.4% to 55.4%. The benchmark was used to validate the change but is not retained as a per-PR CI workload.

Existing `.egg` fixtures already cover ordinary successful and failing checks, while the core-relations suite directly covers callback bounds for cooperative early stopping in serial and parallel modes. The retained 20,000-row, 32-thread regression covers the distinct `check_facts` integration and verifies that stop state does not leak into the next backend run.

This deliberately does not change `exists`, action execution, or unrelated frontend code.

## Validation

- `cargo test --release --test integration_test check_facts_stops_safely_in_parallel_mode -- --exact`
- `make nits`
- `git diff --check`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -96,7 +96,6 @@ jobs:
           - proof_testing_typecheck
           - eggcc-2mm
           - rust_rule_fib
-          - check_facts_many_matches
           - rust_rule_match_overhead
           - rust_rule_match_with_serialize
           - rust_rule_insert_loop
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -96,6 +96,7 @@ jobs:
           - proof_testing_typecheck
           - eggcc-2mm
           - rust_rule_fib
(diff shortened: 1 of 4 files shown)


saulshanabrook:
@CodeRabbit full review

oflatt (approved):
nice small change!

PR #991, opened 2026-08-14
Avoid formatting unused rule names during desugaring

saulshanabrook:
## Summary

- Generate automatic names only for unnamed `rule`, `rewrite`, and `birewrite` commands.
- Preserve supplied names without formatting or cloning them.
- Preserve generated names byte-for-byte, with exact regression coverage for rule options, rewrite conditions, quote normalization, and both birewrite directions.
- Keep a declaration-heavy local `resolve_program` benchmark over `tests/taylor51.egg`; it is intentionally absent from the CodSpeed CI matrix.

## Why

`desugar_command` previously formatted every command before matching on its variant, even though declarations and most other commands cannot consume an automatically generated rule name.

The final implementation resolves names inside the three relevant match arms. A borrowed `RuleLikeCommand` view ties the actual `Rule` or `Rewrite` name field to its canonical formatter. `resolve_name` moves a supplied name out unchanged and formats only when that field is empty. The original owned payload then continues through its already-refined desugaring arm.

This keeps one naming invariant without an `Option`- or sentinel-based resolver, AST reconstruction, name clone, partial match, or panic extraction. The public `rule_name` formatter remains unchanged. This PR does not touch the typechecker.

After the final simplification review, the scoped diff against fresh main is `87` additions and `21` deletions. The production file is 10 lines shorter, and the regression test is 9 lines shorter, than the preceding owned-wrapper candidate and expanded test reviewed during this iteration.

## Performance

I compared fresh upstream main at `7b1adf249c918226871b9b3d5e8f089585e46e99` with the final implementation using byte-identical benchmark code and the same Taylor51 fixture:

- Divan OS timer, sample size 1
- 5 warmups per endpoint
- 12 balanced blocks with 5 samples per endpoint per block (60 measurements each)
- all six endpoint orders repeated twice
- geometric mean of paired block-median ratios with a 20,000-resample block-bootstrap interval

| Endpoint | Median of block medians |
| --- | ---: |
| Fresh main | 20.180 ms |
| Intermediate owned wrapper | 19.870 ms |
| This patch | 19.745 ms |

The patch/main ratio was `0.9773`: **2.27% less time**, with a 95% interval of **1.80% to 2.77% less time**.

The patch/intermediate ratio was `0.9900`: **1.00% less time**, with a 95% interval of **0.12% to 2.08% less time**.

The final one-line source cleanup was rebuilt separately; its complete Mach-O `__TEXT,__text` section was byte-identical to the measured candidate. This Rust API benchmark uses the system allocator; both measured endpoints used the same allocator and harness.

## Validation

- `cargo test --release --test test_desugar` (2 passed)
- `make nits`
- `make test` (1,167 tests passed, plus doctests; no snapshot issues)
- `git diff --check`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,7 +100,6 @@ jobs:
           - rust_rule_match_with_serialize
           - rust_rule_insert_loop
           - rust_rule_tableaction_hot_path
-          - resolve_declaration_heavy
           - factoring-multisets
           # CodSpeed filters by substring; `math` would also run
           # `math-microbenchmark` and `proof_testing_math`.
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,6 +100,7 @@ jobs:
           - rust_rule_match_with_serialize
           - rust_rule_insert_loop
           - rust_rule_tableaction_hot_path
+          - resolve_declaration_heavy
           - factoring-multisets
           # CodSpeed filters by substring; `math` would also run
           # `math-microbenchmark` and `proof_testing_math`.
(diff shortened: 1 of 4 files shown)


oflatt (approved):
looks good, small change

PR #993, opened 2026-08-14
Add small residual join indexes

ezrosent:
This PR is part of an eight-PR stack that jointly replaces #959.

Stack base: #969

Replacement stack:
1. #993 — small residual indexes
2. #994 — packed root caches
3. #995 — prepared index state
4. #996 — packed probe abstractions
5. #997 — join-tail execution state
6. #998 — executor integration
7. #999 — compact frame updates
8. #1000 — lazy preparation

Validation on the complete stack:
- cargo fmt --check
- cargo nextest run -p egglog-core-relations
- 113 tests passed

Layer 1 of 8.

Adds allocation-free residual index primitives in a dedicated implementation file. The executor wiring is intentionally deferred to #998 so this abstraction can be reviewed separately.

Non-test diff: +283 / -0.

PR #994, opened 2026-08-14
Add packed join root caches

ezrosent:

Stack base: #969

Replacement stack:
2. #994 — packed root caches
8. #1000 — lazy preparation

- cargo fmt --check
- 113 tests passed

Layer 2 of 8.

Adds shared packed-root projection and trie-cache support in a dedicated implementation file. The executor wiring is intentionally deferred to #998.

Non-test diff: +246 / -0.


core-relations/src/free_join/packed_cache.rs
```diff
@@ -0,0 +1,246 @@
+type BaseSig = (TableId, SmallVec<[Constraint; 2]>);
+
+/// Key for a shared trie root: the table plus an interned id for its fast
+/// (header) constraints.
+type RootKey = (TableId, u32);
```
ezrosent: should we have a `NumericId` for this header constraint?

core-relations/src/free_join/packed_cache.rs
```diff
@@ -0,0 +1,246 @@
+type BaseSig = (TableId, SmallVec<[Constraint; 2]>);
```
ezrosent: Explain what this is, with more detailed examples and context.

core-relations/src/free_join/packed_cache.rs
```diff
@@ -0,0 +1,246 @@
...
+/// Only roots that more than one plan actually uses are shared (`shared`), so
+/// single-use roots stay per-plan and keep the pool-recycling behavior of the
+/// unshared path — sharing a root that is never reused is pure overhead.
+///
+/// Concurrency: the parallel executor runs plans on multiple threads, so the maps
+/// are concurrent. Tables are frozen during a run, so a given key always denotes
+/// the same subset.
+#[derive(Default)]
+struct TrieCache {
+    roots: DashMap<RootKey, Arc<TrieNode>>,
```
ezrosent, line 129: DashMap isn't very parallel-scalable. When do we actually use this and will we have a lot of these lookups? Or is it just part of setup for a plan. (This should be documented in the comments.)

core-relations/src/free_join/packed_cache.rs
```diff
@@ -0,0 +1,271 @@
...
+/// constraints.
+///
+/// For example, if `edge` has columns `(src, dst)`, the atom `edge(x, y)` has
+/// signature `(edge, [])`, while `edge(x, 3)` has signature
+/// `(edge, [dst == 3])`. Two plans with the latter atom share the same owning
+/// root subset even if their later join stages differ. Sorting makes
+/// `src > 0 && dst == 3` identical to `dst == 3 && src > 0`; including the
+/// table prevents the same constraints on another relation from sharing a
+/// root.
+type RootSignature = (TableId, SmallVec<[Constraint; 2]>);
```
yihozhang, line 12: Is size 2 enough? There's always a fast constraint on timestamp with seminaive, so it leaves only one constraint for EqConstraint with constants? Maybe this is enough. It's hard to think of adversarial examples.

core-relations/src/free_join/packed_cache.rs
```diff
@@ -0,0 +1,271 @@
...
+        } else {
+            // SAFETY: construction consumes pairs sorted by `(Value, RowId)`,
+            // so every equal-value range is RowId ordered.
+            SubsetRef::Sparse(unsafe { SortedOffsetSlice::new_unchecked(rows) })
+        }
+    }
+}
+
+type RootProjectionSlot = Arc<OnceLock<RootProjection>>;
+type RootProjectionMap = DashMap<RootProjectionKey, RootProjectionSlot>;
```
yihozhang: Should we use RootKey here?

yihozhang, line 107: This seems like overkill since there aren't many ColumnIds. Is there a better struct (e.g., an association list?) here? Or maybe we can build one shared RootProjectionMap in TrieCache?

Or is this only called when initializing the plan and is not a hot spot?

core-relations/src/free_join/packed_cache.rs
```diff
@@ -0,0 +1,271 @@
...
+/// (header) constraints.
+type RootKey = (TableId, HeaderConstraintId);
+
+/// One round-local root projection can be reused by plans that share the same
+/// root subset. Slow constraints are part of the key because they are applied
+/// before projection; sorting them makes conjunction order irrelevant.
+#[derive(Clone, Eq, Hash, PartialEq)]
+struct RootProjectionKey {
+    column: ColumnId,
+    constraints: SmallVec<[Constraint; 2]>,
```
yihozhang, line 30: Is it true that RootProjectionKey::constraints contains slow constraints only while RootKey contains fast constraints only? If so can we document this?

core-relations/src/free_join/packed_cache.rs
ezrosent: Follow-up performance work after this code lands:

- Defer sizing tuning to the follow-up performance sweep.
- Map replacement proposal from ezrosent:

> I think the right move is to replace this with a statically-sized lock-free chaining hashmap that allocates chain nodes on an arena. But we should only do this if we can measure a performance speedup for it.

The map replacement is deferred; the current implementation is unchanged.

Written by Codex, in consultation with ezrosent.

core-relations/src/free_join/packed_cache.rs
ezrosent: We’ll defer tuning the inline capacity to a follow-up performance sweep after this code lands. The capacity remains 2 for now; we can measure constraint counts and heap spills before deciding whether a larger inline capacity is worthwhile.

Written by Codex, in consultation with ezrosent.

core-relations/src/free_join/packed_cache.rs
ezrosent: Yes: `RootKey` represents only the fast/header constraints already applied to the root subset, while `RootProjectionKey::constraints` contains only the scan’s remaining slow constraints, applied before projection. Plans with the same fast constraints can share a root, but different slow constraints require separate projections.

Updated the documentation on `RootKey`, `RootProjectionKey::constraints`, and `projection_slot` in this PR to state that distinction explicitly. The change has been carried through the dependent PRs.

Written by Codex, in consultation with ezrosent.

PR #995, opened 2026-08-14
Add prepared join index state

ezrosent:

Stack base: #969

Replacement stack:
2. #994 — packed root caches
8. #1000 — lazy preparation

- cargo fmt --check
- 113 tests passed

Layer 3 of 8.

Adds execution-scoped prepared-index state, continuation caches, and tail-mask preparation in a dedicated implementation file. The executor wiring is intentionally deferred to #998.

Non-test diff: +473 / -0.


core-relations/src/free_join/prepared_index.rs
```diff
@@ -0,0 +1,423 @@
+/// A key ordinal in a root continuation grid.
```
ezrosent: Explain what a continuation is in the context of a join. The tone here is too cryptic.

core-relations/src/free_join/prepared_index.rs
```diff
@@ -0,0 +1,423 @@
...
+                .expect("an index shard cannot contain more than u32::MAX keys"),
+        }
+    }
+}
+
+/// Lazily-created continuation slots for rows read from a persistent or
+/// round-local root index. Slots are deliberately per logical scan: a
+/// different frozen plan may continue the same root key with a different
+/// column or constraint set.
+type RootContinuationSlots = Box<[Box<[OnceLock<usize>]>]>;
```
ezrosent, line 60: I thought these things would all land on the arena? Why do we have tons of Boxes of arrays of boxes of arrays?

core-relations/src/free_join/prepared_index.rs
```diff
@@ -0,0 +1,423 @@
...
+    one_index_access_stages: u64,
+    multiple_index_access_stages: u64,
+}
+
+/// Order-independent tail metadata for plans small enough to represent their
+/// remaining stages in one word. DVO only permutes stages within the fixed
+/// barrier phases, so successor shape depends on the remaining set, not its
+/// current permutation.
+struct PreparedTailMasks {
+    atom_uses: DenseIdMap<AtomId, PreparedAtomUse>,
```
ezrosent, line 234: Document what each of these fields are used for

core-relations/src/free_join/prepared_index.rs
```diff
@@ -0,0 +1,454 @@
...
+/// Index handles for one immutable [`JoinStages`] value, positionally aligned
+/// with `JoinStages::instrs` and with each stage's scans.
+struct PreparedJoinIndexes {
+    stages: Box<[SmallVec<[PreparedIndexSlot; 4]>]>,
+    access_counts: DenseIdMap<AtomId, usize>,
+    tail_masks: Option<PreparedTailMasks>,
+}
+
+impl PreparedJoinIndexes {
+    fn new(db: &Database, atoms: &Arc<DenseIdMap<AtomId, Atom>>, stages: &JoinStages) -> Self {
```
yihozhang, line 354: I realize I should have read this function first, which explains the high-level structure of this file. Maybe leave a note in the top-level doc that someone new to this file should start with this function

core-relations/src/free_join/prepared_index.rs
```diff
@@ -0,0 +1,454 @@
...
+/// the database resets its indexes during `merge_all`.
+enum PreparedIndexKind {
+    Tuple(OnceLock<HashIndex>),
+    Column(OnceLock<HashColumnIndex>),
+    /// The table specification forbids a global cache for at least one key
+    /// column, so execution must use its existing dynamic-index path.
+    Uncacheable,
+}
+
+struct PreparedIndexSlot {
```
yihozhang, line 181: I find this struct hard to understand, probably because all of its uses are in `execute.rs` (in a later PR) via direct field access, and it lacks any associated methods to encapsulate operations over `PreparedIndexSlot`.

PR #996, opened 2026-08-14
Add packed join probe abstractions

ezrosent:

Stack base: #969

Replacement stack:
2. #994 — packed root caches
8. #1000 — lazy preparation

- cargo fmt --check
- 113 tests passed

Layer 4 of 8.

Adds the uniform row and probe abstractions used by packed, catalog, projected-root, and small residual indexes. The executor wiring is intentionally deferred to #998.

Non-test diff: +719 / -0.


core-relations/src/free_join/probe.rs
```diff
@@ -0,0 +1,719 @@
...
+
+impl<'rows, 'exec> AtomRows<'rows, 'exec>
+where
+    'exec: 'rows,
+{
+    fn kind_name(&self) -> &'static str {
+        match self {
+            Self::Root(_) => "root",
+            Self::Catalog { .. } => "catalog",
+            Self::Packed(_) => "packed",
```
ezrosent, line 79: Explain what this is used for

core-relations/src/free_join/probe.rs
```diff
@@ -0,0 +1,719 @@
...
+    }
+}
+
+impl<'rows, 'exec> From<Arc<TrieNode>> for AtomRows<'rows, 'exec> {
+    fn from(root: Arc<TrieNode>) -> Self {
+        Self::Root(root)
+    }
+}
+
+enum ProbeIndex<'ctx, 'rows, 'exec> {
```
ezrosent, line 129: Explain in doc comments on each enum variant what each ProbeIndex is used for.

core-relations/src/free_join/probe.rs
```diff
@@ -0,0 +1,719 @@
...
+    columns: SmallVec<[ColumnId; 4]>,
+    table: WrappedTableRef<'ctx>,
+    handle: &'ctx Handle<'exec>,
+    scratch: &'ctx RefCell<Vec<(Value, RowId)>>,
+    terminal_child_shape: ChildShape,
+}
+
+/// Direct probe over a final-form root projection shared across plans in the
+/// current ruleset execution. Immutable keys and rows stay in the shared root;
+/// only continuation slots and packed descendants belong to this query.
```
ezrosent: Too cryptic again: under what circumstances do you have a RootProjectionProbe? what does it mean? Define terms before you use them ("final-form root projection" ? )

core-relations/src/free_join/probe.rs
```diff
@@ -0,0 +1,719 @@
...
+    }
+}
+
+struct Prober<'ctx, 'rows, 'exec> {
+    source: AtomRows<'rows, 'exec>,
+    ix: ProbeIndex<'ctx, 'rows, 'exec>,
+    keep_rows: bool,
+}
+
+struct ProbeRequest<'scan, 'rows> {
```
ezrosent, line 424: Explain what this is

core-relations/src/free_join/probe.rs
ezrosent, line 1: Provide a module doc comment about what the helpers in this module are doing.

core-relations/src/free_join/probe.rs
```diff
@@ -0,0 +1,762 @@
...
+            let resr = cmp::min(r.end, range.end);
+            if resl >= resr {
+                None
+            } else {
+                Some(SubsetRef::Dense(OffsetRange::new(resl, resr)))
+            }
+        }
+        SubsetRef::Sparse(s) => {
+            let l = s.binary_search_by_id(range.start);
+            let r = s.binary_search_by_id(range.end);
```
yihozhang, line 21: nit: can start the search for `end` from `l`

core-relations/src/free_join/probe.rs
```diff
@@ -0,0 +1,762 @@
...
+/// again.
+#[derive(Clone, Copy)]
+pub(super) struct CatalogContinuation<'rows> {
+    cache: &'rows RootContinuationCache,
+    position: ContinuationPosition,
+}
+
+#[derive(Clone)]
+pub(super) enum AtomRows<'rows, 'exec> {
+    Root(Arc<TrieNode>),
```
yihozhang, line 46: Should we rename `TrieNode` to `TrieRoot`? Then `PackedTrieNode` is `TrieNode`

core-relations/src/free_join/probe.rs
```diff
@@ -0,0 +1,762 @@
...
+        });
+        // SAFETY: this continuation cache belongs to the prepared query and can
+        // only publish nodes allocated by the same query's SharedArena.
+        let child = unsafe { &*(address as *const PackedTrieNode<'exec>) };
+        assert_eq!(child.child_shape(), child_shape);
+        child
+    }
+
+    fn get(&self, key: &[Value]) -> Option<AtomRows<'rows, 'exec>> {
+        if key.len() != self.columns.len() {
```
yihozhang, line 260: This should be a debug_assert?

core-relations/src/free_join/probe.rs
```diff
@@ -0,0 +1,762 @@
...
+        }
+    }
+}
+
+impl<'ctx, 'rows, 'exec> PackedProbe<'ctx, 'rows, 'exec>
+where
+    'exec: 'rows,
+{
+    fn get(&self, key: &[Value]) -> Option<AtomRows<'rows, 'exec>> {
+        if key.len() != self.columns.len() {
```
yihozhang, line 344: This should be a debug assert?

core-relations/src/free_join/probe.rs
```diff
@@ -0,0 +1,762 @@
...
+            key.pop();
+        }
+    }
+}
+
+impl<'ctx, 'rows, 'exec> PackedProbe<'ctx, 'rows, 'exec>
+where
+    'exec: 'rows,
+{
+    fn get(&self, key: &[Value]) -> Option<AtomRows<'rows, 'exec>> {
```
yihozhang, line 343: For the majority of the workload (i.e., `Intersect`, or `FusedIntersect` where the intersected atoms only share one variable with `cover`), is `key.len() == 1`?

core-relations/src/free_join/probe.rs
```diff
@@ -0,0 +1,762 @@
...
+impl<'scan, 'rows> ProbeRequest<'scan, 'rows> {
+    fn column(
+        scan: &'scan SingleScanSpec,
+        keep_rows: bool,
+        terminal_child_shape: ChildShape,
+        prepared: &'rows PreparedIndexSlot,
+    ) -> Self {
+        Self {
+            atom: scan.atom,
+            columns: SmallVec::from_slice(&[scan.column]),
```
yihozhang, line 442: smallvec![scan.column]

core-relations/src/free_join/probe.rs
```diff
@@ -0,0 +1,762 @@
...
+
+    fn tuple(
+        scan: &'scan ScanSpec,
+        keep_rows: bool,
+        terminal_child_shape: ChildShape,
+        prepared: &'rows PreparedIndexSlot,
+    ) -> Self {
+        Self {
+            atom: scan.to_index.atom,
+            columns: scan.to_index.vars.iter().copied().collect(),
```
yihozhang, line 458: We should be consistent about what `N` we use for `SmallVec<[ColumnId; N]>`

core-relations/src/free_join/probe.rs
```diff
@@ -0,0 +1,762 @@
...
+    fn for_each(&self, f: &mut impl FnMut(&[Value], AtomRows<'rows, 'exec>)) {
+        let mut key = SmallVec::new();
+        self.for_each_recur(self.first, 0, &mut key, f);
+    }
+}
+
+struct Prober<'ctx, 'rows, 'exec> {
+    source: AtomRows<'rows, 'exec>,
+    ix: ProbeIndex<'ctx, 'rows, 'exec>,
+    keep_rows: bool,
```
yihozhang, line 412: Refer to `ProbeMatch` and document in what scenarios keep_rows is true and false.

core-relations/src/free_join/probe.rs
```diff
@@ -0,0 +1,762 @@
...
+/// Normalized input to [`JoinState::get_index`] for one indexed scan.
+///
+/// Both scalar [`SingleScanSpec`]s and tuple [`ScanSpec`]s are converted to
+/// this form so physical-strategy selection has one code path. It identifies
+/// the atom and projected columns, carries constraints that must be applied to
+/// the source rows, records whether later stages need the matching rows or only
+/// existence, and supplies the prepared sidecar slot used for cached indexes
+/// and continuations. `terminal_child_shape` describes how execution may
+/// continue after the final requested column.
+struct ProbeRequest<'scan, 'rows> {
```
yihozhang, line 424: I wonder if we should specialize ProbeRequest (and ProberIndex) into single-column and multi-column cases.

yihozhang (commented):
Still trying to understand the new trie structures

core-relations/src/free_join/probe.rs
ezrosent: I think we should defer this to a follow-up PR.

core-relations/src/free_join/probe.rs
ezrosent: I think so... I'd say this should also be picked up in a follow-up specialization PR.

PR #997, opened 2026-08-14
Add join-tail execution state

ezrosent:

Stack base: #969

Replacement stack:
2. #994 — packed root caches
8. #1000 — lazy preparation

- cargo fmt --check
- 113 tests passed

Layer 5 of 8.

Adds join-tail analysis, binding-state projection, ordering helpers, and worker-local state in a dedicated implementation file. The executor wiring is intentionally deferred to #998.

Non-test diff: +633 / -0.


core-relations/src/free_join/join_tail.rs
ezrosent, line 1: Give a module doc comment explaining what these helpers do

core-relations/src/free_join/join_tail.rs
```diff
@@ -0,0 +1,633 @@
...
+    (resume_pos..instr_order.len()).any(|position| {
+        let mut found = false;
+        for_each_stage_materialization(&stages[instr_order.get(position)], |candidate| {
+            found |= candidate == materialization;
+        });
+        found
+    })
+}
+
+type BindingSet = Vec<(SmallVec<[Variable; 4]>, Arc<TaggedRowBuffer<SmallValueVec>>)>;
```
ezrosent, line 212: Explain what a BindingSet is in a doc comment.

core-relations/src/free_join/join_tail.rs
```diff
@@ -0,0 +1,633 @@
...
+    matches!(
+        stage,
+        JoinStage::FusedIntersectMat {
+            mode: MatScanMode::Lookup(_) | MatScanMode::Value(_) | MatScanMode::Full,
+            ..
+        }
+    )
+}
+
+/// Reorder a suffix of the cached logical plan using the original DVO policy.
```
ezrosent: "original DVO policy?" what's that? I think there's really just one DVO policy in the new proposed commit chain?

core-relations/src/free_join/join_tail.rs
```diff
@@ -0,0 +1,633 @@
...
+        if is_reorder_barrier(&instrs[i]) {
+            sort_plan_by_size_inner(order, last_pos..i, instrs, binding_info);
+            last_pos = i + 1;
+        }
+    }
+    sort_plan_by_size_inner(order, last_pos..instrs.len(), instrs, binding_info);
+    recompute_leaf_scans(order, leaf_scans, instrs, start);
+}
+
+/// Recompute `leaf_scans[i]` for every position `i` in `[start, order.len())` against the
```
ezrosent: Give more context: what is a LeafScan and under what circumstances does it need to be recomputed?

yihozhang (commented):
Two thoughts:

* `materialization_is_live_in_tail` may repeat the same analysis even when the variable ordering is unchanged
* Sibling branches do not share analysis results, even though they may share the same (updated) variable ordering from DVO. 

The first one is a low-hanging fruit IMO. We can follow up on them and benchmark the actual cost before deciding whether optimization is worthwhile.

core-relations/src/free_join/join_tail.rs
```diff
@@ -0,0 +1,660 @@
...
+type LeafScans = SmallVec<[bool; 8]>;
+
+struct BorrowedLocalState<'a, 'rows, 'exec> {
+    instr_order: &'a mut InstrOrder,
+    leaf_scans: &'a mut LeafScans,
+    binding_info: &'a mut BindingInfo<'rows, 'exec>,
+    updates: &'a mut FrameUpdates<'rows, 'exec>,
+}
+
+struct SubsetClonePlan<'a> {
```
yihozhang, line 604: Let's have a brief documentation for each struct.

ezrosent:
+1... I think we should look at sharing some of that in a follow-up PR. Definite potential low-hanging fruit.

PR #998, opened 2026-08-14
Use packed arena tries for generic joins

ezrosent:

Stack base: #969

Replacement stack:
2. #994 — packed root caches
8. #1000 — lazy preparation

- cargo fmt --check
- 113 tests passed

Layer 6 of 8.

Wires the supporting abstractions into generic-join execution and switches the executor to packed arena tries. Unit tests live in a separate execute_tests.rs file; end-to-end coverage remains in the crate test module.

Diff (including tests): +3130 / -2276, net +854.


core-relations/src/free_join/execute.rs
```diff
@@ -939,358 +402,77 @@ impl Database {
...
-
-    fn has_empty_subset(&self, atom: AtomId) -> bool {
-        self.subsets[atom].subset.is_empty()
     }
+}
 
-    fn unwrap_val(&mut self, atom: AtomId) -> Arc<TrieNode> {
-        self.subsets.unwrap_val(atom)
-    }
+struct JoinState<'db, 'state, 'exec> {
```
ezrosent, line 476: Document what JoinState is used for

core-relations/src/free_join/execute.rs
```diff
@@ -1371,90 +553,328 @@ impl<'a> JoinState<'a> {
         Some(subset)
     }
 
-    fn get_index<'plan>(
+    fn build_packed_node(
+        &self,
+        table: WrappedTableRef<'_>,
+        subset: SubsetRef<'_>,
+        can_be_stale: bool,
```
ezrosent: What does `can_be_stale` mean for a node?

core-relations/src/free_join/execute.rs
```diff
@@ -1371,90 +553,328 @@ impl<'a> JoinState<'a> {
...
+            &self.handle,
+            table,
+            filtered.as_ref(),
+            column,
+            child_shape,
+            &mut self.packed_scratch.borrow_mut(),
+        )
+    }
+
+    /// Acquire the final-form scalar root index shared by plans using the same
```
ezrosent: Too cryptic. Explain what all these terms mean "final-form scalar root index", "prepared sidecar" "slot Arc"

core-relations/src/free_join/execute.rs
```diff
@@ -1671,43 +1118,73 @@ impl<'a> JoinState<'a> {
...
+        stages: &'plan JoinStages,
+        prepared: &'rows PreparedJoinIndexes,
+        atoms: &'plan Arc<DenseIdMap<AtomId, Atom>>,
         action: A,
         instr_order: &mut InstrOrder,
         leaf_scans: &mut LeafScans,
         cur: usize,
         index_shard: Option<usize>,
-        binding_info: &mut BindingInfo,
+        remaining_stages: Option<u64>,
```
ezrosent, line 1198: Explain what every parameter is used for in a doc comment. (not just this new one).

core-relations/src/free_join/execute.rs
```diff
@@ -1379,90 +579,347 @@ impl<'a> JoinState<'a> {
...
+            AtomRows::Root(root) => match &root.subset {
+                Subset::Dense(range) => Some(*range),
+                Subset::Sparse(_) => None,
+            },
+            _ => None,
+        };
+        let can_use_catalog = root_range.is_some()
+            && all_cacheable
+            && constraints.is_empty()
+            && !info.table.has_stale_rows()
```
yihozhang, line 801: Do we need this condition? We used to not need `info.table.has_stale_rows()` and instead do a `filter_live()` when scanning the subset.

Similarly for `constraints.is_empty()`

core-relations/src/free_join/execute.rs
```diff
@@ -63,588 +63,16 @@ use crate::{
...
-        match &self.ix {
-            DynamicIndex::Cached { table, .. } => table.len(),
-            DynamicIndex::CachedColumn { table, .. } => table.len(),
-            DynamicIndex::Dynamic(tab) => tab.len(),
-            DynamicIndex::DynamicColumn(tab) => tab.len(),
-            DynamicIndex::SparseColumn(tab) => tab.len(),
-        }
-    }
-}
+include!("residual_index.rs");
```
yihozhang, line 71: Use `mod` instead?

core-relations/src/free_join/execute.rs
ezrosent: Added some tests and confirmed that this is required, as is. I agree it's a bit gross.

PR #1003, opened 2026-08-19
Add Read::eclass_enodes, and scope the old name to one table

oflatt-claude:
Follow-up to #986, from using that API to build `unstable-subst` in
egglog-experimental.

### The name was a promise the method did not keep

`Read::enodes_for_eclass(name, eclass, f)` reads as "the e-nodes of this
e-class", but it returns only the rows of the one table it is handed. A caller
who genuinely wants an e-class's e-nodes has to know which constructors could
produce it and loop over them — which is what `unstable-subst` does today.

So this splits the two ideas:

| method | scope |
| --- | --- |
| `constructor_enodes(name, f)` | every row of one table |
| `constructor_enodes_for_eclass(name, eclass, f)` | rows of one table, one e-class |
| `eclass_enodes(eclass, f)` | rows of **every** constructor, one e-class |

The renamed one reads as "`constructor_enodes`, filtered" and sits with its
neighbours; the new one is what the old name was reaching for.

### `Enode` gained a `name`

A caller that did not name a table cannot otherwise interpret what it gets
back: the children's sorts come from the constructor's signature, and
re-applying it needs the name. Both existing scans fill it in, so there is one
`Enode` shape across all three methods.

### Which one to reach for

`eclass_enodes` probes every constructor whose output is an eq-sort. It has to:
a `Value` does not say what sort it belongs to — there is no value-to-sort map,
only one global id counter — so there is nothing to narrow by. That same fact
is what makes it sound: probing a constructor of the wrong sort simply finds
nothing.

A caller that *does* know the sort should keep narrowing to that sort's
constructors and use `constructor_enodes_for_eclass`, which is one probe rather
than one per constructor. The doc on each says so. In `unstable-subst` the
distinction is load-bearing: it knows each child's sort from its parent's
signature and narrows there, and only the root — the one e-class whose sort it
cannot know — needs the spanning version.

### Notes

- Additive apart from the rename and the new `Enode` field. `Enode` is only
  constructed inside egglog, so downstream code that reads its fields is
  unaffected.
- `tests/api_introspection.rs` covers the new method: an e-class with e-nodes
  from two different constructors, checking both come back with the right
  `name`, and that neither a constructor of another sort that merely references
  the e-class nor a `function` table is included.
- Full workspace suite, clippy, and `fmt --check` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - **Add `Read::eclass_enodes`, and rename `Read::enodes_for_eclass` to `Read::constructor_enodes_for_eclass`.** The old name read as "the e-nodes of this e-class" but returned only the rows of the one table it was given. `eclass_enodes(eclass, f)` is the method that name promised: it spans every constructor, and `Enode` gained a `name` field saying which one each row came from. A caller that already knows the e-class's sort should still narrow to that sort's constructors and call `constructor_enodes_for_eclass`, which is one probe rather than one per constructor — a `Value` does not say what sort it belongs to, so the spanning version has to ask each of them.
 
+- Fix a panic when a `Read` or `Write` API names a table that a `pop` has since dropped. Lookups now report it missing and `table_sizes` leaves it out, instead of handing the backend an id it no longer has.
+
 ## [3.0.0] - 2026-08-18
 
 ### Breaking changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## [Unreleased] - ReleaseDate
 
-- **Add `Read::eclass_enodes`, and rename `Read::enodes_for_eclass` to `Read::constructor_enodes_for_eclass`.** The old name read as "the e-nodes of this e-class" but returned only the rows of the one table it was given. `eclass_enodes(eclass, f)` is the method that name promised: it spans every constructor, and `Enode` gained a `name` field saying which one each row came from. A caller that already knows the e-class's sort should still narrow to that sort's constructors and call `constructor_enodes_for_eclass`, which is one probe rather than one per constructor — a `Value` does not say what sort it belongs to, so the spanning version has to ask each of them.
+- Add `Read::eclass_enodes` to scan an e-class across every constructor, rename the single-constructor method to `constructor_enodes_for_eclass`, and add `Enode::name`.
 
-- Fix a panic when a `Read` or `Write` API names a table that a `pop` has since dropped. Lookups now report it missing and `table_sizes` leaves it out, instead of handing the backend an id it no longer has.
+- Treat tables dropped by `pop` as missing in the name-indexed `Read` and `Write` APIs.
 
 ## [3.0.0] - 2026-08-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased] - ReleaseDate
 
+- **Add `Read::eclass_enodes`, and rename `Read::enodes_for_eclass` to `Read::constructor_enodes_for_eclass`.** The old name read as "the e-nodes of this e-class" but returned only the rows of the one table it was given. `eclass_enodes(eclass, f)` is the method that name promised: it spans every constructor, and `Enode` gained a `name` field saying which one each row came from. A caller that already knows the e-class's sort should still narrow to that sort's constructors and call `constructor_enodes_for_eclass`, which is one probe rather than one per constructor — a `Value` does not say what sort it belongs to, so the spanning version has to ask each of them.
+
 ## [3.0.0] - 2026-08-18
 
 ### Breaking changes
--- a/core-relations/src/action/mod.rs
+++ b/core-relations/src/action/mod.rs
@@ -497,6 +497,13 @@ impl<'a> ExecutionState<'a> {
         self.db.table_info.iter().map(|(id, _)| id)
     }
 
+    /// Whether `table` is visible to this execution state. Ids obtained
+    /// elsewhere can name a table this state does not have, and every accessor
+    /// here panics on those.
+    pub fn table_exists(&self, table: TableId) -> bool {
+        self.db.table_info.get(table).is_some()
+    }
+
     /// Get an immutable reference to the table with id `table`.
     /// Dangerous: Reading from a table during action execution may break the semi-naive evaluation
     pub fn get_table(&self, table: TableId) -> &'a WrappedTable {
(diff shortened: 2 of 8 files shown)


oflatt-claude:
Pushed one more commit, a bug fix I hit while exercising this API from `egglog-experimental`.

`ActionRegistry` lives behind an `Arc<RwLock<_>>` that `push` shares rather than deep-copies, so it keeps naming tables the popped e-graph no longer has. Every accessor that resolves a name through the registry then indexed the backend with a dangling `TableId` and panicked in `ExecutionState::get_table`. `table_sizes` did it for the whole registry at once, so *any* primitive that enumerates tables crashed after a `pop` — which is how I found it.

```
(datatype Math (Num i64) (Add Math Math))
(function count-of (String) i64 :no-merge)
(push)
(constructor Neg (Math) Math)
(pop)
(rule () ((set (count-of "n") (table-count "x"))) :naive)   ;; table-count calls table_sizes()
(run 1)
;; thread 'main' panicked: called `Option::unwrap()` on a `None` value
```

The fix: a table dropped by a `pop` reads as missing rather than stale. `lookup_action` filters on a new `TableAction::is_live`, so `Read`/`Write` lookups by name return `ApiError::MissingTable`, and `table_sizes` leaves those entries out. Both paths have regression tests in `tests/api_introspection.rs`, and I confirmed each fails without the fix.

Note the id-reuse wrinkle, which is why the second test declares `count-of` before the `push`: if a later declaration reclaims the freed `TableId`, the stale entry silently points at an unrelated table instead of panicking. That is the quieter half of the same bug.

Happy to split this into its own PR if you'd rather keep this one to the rename — it is independent of `eclass_enodes`.

saulshanabrook:
@CodeRabbit full review

egglog-bridge/src/lib.rs
```diff
@@ -71,15 +71,19 @@ impl ActionRegistry {
     }
 
     /// Look up the [`TableAction`] for a table by name, or `None` if
-    /// no table with that name has been registered.
+    /// no table with that name has been registered. The registry outlives
+    /// `push`/`pop`, so a hit may name a table a given execution state no
```
```suggestion
    /// no table with that name has been registered. The registry may become obsolete
    /// because of `push`/`pop`, so a hit may name a table a given execution state no
```

core-relations/src/action/mod.rs
```diff
@@ -497,6 +497,13 @@ impl<'a> ExecutionState<'a> {
         self.db.table_info.iter().map(|(id, _)| id)
     }
 
+    /// Whether `table` is visible to this execution state. Ids obtained
+    /// elsewhere can name a table this state does not have, and every accessor
+    /// here panics on those.
+    pub fn table_exists(&self, table: TableId) -> bool {
```
yihozhang: Should we change places in ExecutionState where we use `table_info[table]` to do `table_info.get(table)?` and return errors? Or do you think panic is fine

src/exec_state.rs
```diff
@@ -483,6 +489,34 @@ pub trait Read<'a, 'db: 'a>: Core<'a, 'db> + RegistrySealed<'a, 'db> {
...
+    /// [`Enode::name`] says which one each row came from.
+    ///
+    /// A [`Value`] does not say what sort it belongs to, so this probes each
+    /// constructor whose output is an eq-sort. A caller that already knows the
+    /// sort should narrow to those constructors itself and use
+    /// [`Read::constructor_enodes_for_eclass`], which is one probe rather than
+    /// one per constructor.
+    fn eclass_enodes(&self, eclass: Value, mut f: impl FnMut(Enode<'_>)) -> Result<(), Error> {
+        let names: Vec<String> = self
+            .table_sizes()
```
yihozhang: nit: we can make a `tables()` function and call it here. `tables()` can return an iterator created from `table_sizes()`

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased] - ReleaseDate
 
+- **Add `Read::eclass_enodes`, and rename `Read::enodes_for_eclass` to `Read::constructor_enodes_for_eclass`.** The old name read as "the e-nodes of this e-class" but returned only the rows of the one table it was given. `eclass_enodes(eclass, f)` is the method that name promised: it spans every constructor, and `Enode` gained a `name` field saying which one each row came from. A caller that already knows the e-class's sort should still narrow to that sort's constructors and call `constructor_enodes_for_eclass`, which is one probe rather than one per constructor — a `Value` does not say what sort it belongs to, so the spanning version has to ask each of them.
```
yihozhang: nit: can be made more concise

yihozhang (approved):
Looks good!

egglog-bridge/src/lib.rs
oflatt-claude: Applied this wording in 6832a667.

egglog-bridge/src/lib.rs
```diff
@@ -1423,6 +1427,14 @@ impl TableAction {
         }
     }
 
+    /// Whether the table this acts on still exists in `state`. Every other
+    /// method taking a state panics when it does not: a table declared inside
+    /// a `push` is dropped by the `pop`, while `TableAction`s handed out
+    /// beforehand keep naming it.
+    pub fn is_live(&self, state: &ExecutionState) -> bool {
+        state.table_exists(self.table)
+    }
```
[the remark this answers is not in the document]
oflatt-claude: Fixed in 6832a667. Each table allocation now has an opaque identity that snapshots preserve but later allocations do not reuse. `TableAction` stores that identity and its registered name, and `is_live` compares both with the current state. I also added a regression that pops `Neg`, creates `Other` with the reused numeric ID, and verifies reads and writes through the old name return `MissingTable`.

core-relations/src/action/mod.rs
oflatt-claude: I think panic is appropriate for the low-level `ExecutionState` operations that take a `TableId`: those IDs are expected to come from the current `DbView`, and changing all of those internal/hot-path APIs to return `Result` would be a much broader change. At the boundary that can legitimately retain a stale ID (`TableAction`), I changed `table_name` and the new `table_identity` lookup to use `.get()`/`Option`; the public name-indexed APIs then turn a failed liveness check into `MissingTable`.

CHANGELOG.md
oflatt-claude: Done in 6832a667. I condensed both Unreleased entries to one sentence each while retaining the user-facing API changes and `pop` behavior.

src/exec_state.rs
oflatt-claude: Done in 6832a667: added `Read::tables()`, derived from `table_sizes()`, and changed `eclass_enodes` to use it.

PR #1005, opened 2026-08-24
Give schedulers e-graph access and per-rule match application

yihozhang:
## What

Two extensions to the custom-scheduler API, motivated by size-limited scheduling (keeping an e-graph within a node budget, as Herbie does):

1. **`SchedulerContext`.** `Scheduler::filter_matches` and `Scheduler::can_stop` now take a `&SchedulerContext` exposing the e-graph (read-only) and a 0-based iteration counter, so schedulers can consult e-graph sizes when choosing matches. Borrow-wise this needed no runner restructuring: during the filter step the runner holds only shared borrows of the `EGraph`, and the scheduler itself is taken out of it. `can_stop` gets the context too so a size-aware scheduler can declare saturation at its limit instead of spinning while all rules are delayed.

2. **Per-rule match application.** The runner now applies each rule's chosen matches (flush + that rule's action rule) right after its `filter_matches`, before consulting the next rule, instead of batching all rules' matches into one apply at the end of the iteration. Sizes observed through the context are therefore live within an iteration, so a scheduler can enforce a node limit by measurement rather than by estimating growth. Rules that choose no matches are skipped entirely (and no longer appear in the iteration's report). The e-graph is **rebuilt once at the end of the iteration**, as in egg, not after every rule: egglog-bridge gains `run_rules_no_rebuild`, `flush_updates_no_rebuild` and `rebuild_now` for this (between rules, size queries are upper bounds since tables may hold rows a rebuild would merge). This was first an opt-in (`Scheduler::apply_immediately`); measuring it showed no reason to keep the batched path (below), so it is now the only path.

Also adds `EGraph::total_size()` and `EGraph::num_nodes()`, and `Read::is_sort_unionable` so read primitives can draw the same e-node/fact distinction; `num_nodes` counts only rows of tables whose output is a *unionable* eq-sort — constructors, but not `relation`s (which desugar to constructors over a non-unionable sort) or functions to base sorts — matching the e-node count of a traditional e-graph, unlike `(get-size!)` which counts every table.

Smaller: the scheduler's per-rule query rule follows the EGraph's `seminaive` setting (`--naive`) like ordinary rules.

## Why this shape

On Herbie's workload (155 real rewriting runs, node limit 4000), the stock back-off scheduler overshoots the limit by +93% median / +472% max because all decisions are made against iteration-start sizes. A growth-*estimating* scheduler using only `SchedulerContext` reduces that to +5%/+93%, and a live-checking scheduler using `apply_immediately` reaches +4% median / +53% max — statistically indistinguishable from egg on the same inputs — while measuring *faster* end-to-end than batched application (tighter graphs repay the extra flushes). Rebuilding per rule cost 12% of engine CPU on that workload; rebuilding once per iteration recovers it (Herbie: 78.3 s → 74.3 s wall on bench/numerics) with the same size control.

**Cost of per-rule application on its own** (156 rewrite programs, identical scheduling decisions in both modes, rewrite phase only, 5 reps): single-threaded, per-rule application is 0.96× / 0.99× / 1.00× the batched time at node limits 4k / 16k / 50k (per-file p90 ≤ 1.04×); with `-j 8` it is 1.02× / 1.07× / 1.07× wall while using 0.78–1.01× the CPU (the apply phase runs one rule set per rule, so it forgoes `run_rule_set`'s cross-rule parallelism). On egglog's `gemma4_moe` test forced under `back-off` it is 1.00× (`-j 1`) and 0.98× (`-j 8`). That is small enough that the batched path was removed rather than kept behind an option.

## Breaking change

`Scheduler::filter_matches` and `can_stop` gain the context parameter; migration is mechanical (accept and ignore `ctx`). CHANGELOG updated.

## Tests

- `test_scheduler_sees_egraph_size`: a scheduler halts growth exactly at its node limit via `ctx`, and `num_nodes`/`total_size` split e-nodes from analysis rows (base-sort functions and a `relation`).
- `test_filter_matches_sees_fresh_sizes`: a later rule observes an earlier rule's insertions within the same iteration.
- `test_first_n_scheduler` updated: a rule whose matches were all consumed is not run, so it has no entry in the final iteration's report.
- Existing scheduler tests migrated; full `cargo test --release` passes.

The consumer side (a size-aware `back-off`) is in the companion egglog-experimental PR.


saulshanabrook:
@CodeRabbit full review

src/scheduler.rs
```diff
@@ -20,16 +35,35 @@ pub trait Scheduler: dyn_clone::DynClone + Send + Sync {
...
+    /// instead of batching all rules' chosen matches into a single apply at
+    /// the end of the iteration.
+    ///
+    /// With immediate application, the e-graph sizes observed through the
+    /// [`SchedulerContext`] are up to date within an iteration, at the cost
+    /// of one flush and one action-rule run per rule that chose matches.
+    /// Default is `false` (batched).
+    fn apply_immediately(&self) -> bool {
+        false
+    }
```
saulshanabrook: Would be interesting to see any benchmarks that show that this makes a difference... Like could we just keep this on as true always to remove the extra flag, or does that really harm behavior of using backoff without node limits?

saulshanabrook: like totally fine if we need it, just preference would be to change our behavior now and keep a simpler code path than preserve an option, unless its really needed

yihozhang: The downside of always keeping it true is that issuing many small rule firings can hurt performance. I will measure the exact numbers. But it actually makes Herbie faster because it allows Herbie to stop at a much smaller node limit.

yihozhang: Verified that the slowdown is not more than 1% for single-thread performance. So I simplified the code path. If this turned out to be a problem for multi-thread settings (because we lose inter-query parallelism when applying actions), we can add it back.

saulshanabrook (approved):

PR #1006, opened 2026-08-28
Fix(Bug): Fix collisions

wjrforcyber:
Fresh-name collision. This will cause term bound to even wrong term.
This issue is identified originally in egglog-python and traced back here. If possible, update egglog-python too.

## Root cause

https://github.com/egraphs-good/egglog/blob/e264c37a3332453eb0b6486c82c8766dd1af17df/src/util.rs#L61-L75

The counter is **per hint**, so two different hints can produce the same string:

| hint    | count | minted name |
|---------|-------|-------------|
| `n414`  | 11    | `@n41411`   |
| `n4141` | 1     | `@n41411`   |

Downstream, variable identity is name-based, so the two variables were conflated while compiling the set action and node was silently bound to the wrong term.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3495,4 +3495,56 @@ mod tests {
             .unwrap_err();
         assert!(matches!(err, Error::IoError(..)));
     }
+
+    /// Regression test: `SymbolGen::fresh` minted colliding names for
+    /// different hints (e.g. hint "x" at count 11 and hint "x1" at count 1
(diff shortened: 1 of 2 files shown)


saulshanabrook (approved):

saulshanabrook:
@wjrforcyber Thank you for finding this and fixing it!

Makes me wonder about all our fresh generation if there are more issues like this around, but this seems like a good fix for now.

oflatt:
Nice catch!

PR #1007, opened 2026-08-28
Fix scheduler semantics and node counting

saulshanabrook:
This is a focused follow-up to #1005. It fixes scheduler correctness and recovery edge cases, provides the small read-only context needed by size-aware schedulers, and makes e-node counting consistent across normal, term, and proof encodings. The corresponding experimental consumer is egraphs-good/egglog-experimental#66, for the size-aware back-off scheduler used by herbie-fp/herbie#1646.

## Scheduler execution model

Every rule's selected matches are applied before the next scheduler callback. Queries are collected together first, but actions selected by one callback are flushed before the next callback, so the next scheduler observes their effects. This is unconditional for normal seminaive, rule-local <code>:naive</code>, and <code>:unsafe-seminaive</code> rules; there is no <code>apply_immediately</code> option or batched fallback.

Rebuilding remains deferred until the rule loop finishes, so a successful scheduler step rebuilds at most once. The bridge exposes the three operations needed for that runner-owned boundary:

- <code>run_rules_no_rebuild</code> applies a rule without rebuilding;
- <code>flush_updates_no_rebuild</code> makes its writes visible to the next callback;
- <code>rebuild_now</code> restores canonical tables at the end of the step.

### Scheduled rules retain their compiled execution options

The scheduler splits each source rule into a query rule and an action rule. Reconstructing the query from global e-graph settings loses rule-local choices such as <code>:naive</code>, <code>:unsafe-seminaive</code>, <code>:no-decomp</code>, and <code>:internal-include-subsumed</code>.

For example, a rule-local naive rule in an otherwise seminaive e-graph must search the whole database again. If its scheduler query instead uses the global seminaive mode, a second scheduler step can fail to re-match the same trigger.

The scheduler now stores the already-compiled rule metadata and uses its exact evaluation mode, planner setting, read-context requirement, and subsumption setting when constructing the query/action pair. The regression uses two ordered rules:

1. <code>a-write</code> changes <code>(f)</code> from 0 to 1;
2. <code>b-read</code> copies the live <code>(f)</code> into <code>(g)</code>.

After the first step, both values are 1, proving that the second callback observes the first rule's action in every supported mode. A second local/global-naive step reaches 2, proving that naive rules re-query the full database. A separate case proves that only a rule declared with <code>:internal-include-subsumed</code> sees a subsumed row.

### Deferred matches are not accidentally discarded

Residual handling must follow the scheduled rule's mode and the scheduler's re-query decision. A concrete regression starts with 20 naive matches, selects three per callback, and returns <code>false</code> to continue using the current match set. Clearing the other 17 matches stalls processing after 3. The fixed runner retains and drains them as 3, 6, 9, 12, 15, 18, 20. A naive rule discards residuals only when its scheduler explicitly requests a fresh full query.

### Rule-action errors leave a storage-canonical partial state

<code>(panic ...)</code> inside an egglog action is a catchable egglog error, not a Rust panic. It requests cooperative early stopping. Rule execution is batched and may be parallel, so an unspecified subset of action effects may already have executed or been staged when the error is observed; there is no transaction, exact-once guarantee, or strict textual/global action prefix.

The ordinary runner commits those partial effects and advances every participating rule's seminaive <code>last_run_at</code> before execution. This branch rebuilds before returning the egglog error but does not rewind row timestamps or rule cursors. If recovery rebuilding succeeds, later commands may safely inspect and call the storage-canonical e-graph, but matches that were not completed before the stop are not guaranteed to run on the next seminaive step.

<code>(fail (run problematic 1))</code> therefore catches the returned egglog error without restoring the pre-run state. Following commands observe whatever partial database effects committed. On unchanged input, a later seminaive run commonly does not retry the same match; a naive or newly enabled query may rediscover it.

The custom scheduler follows the same no-rollback, no-retry policy. Before the fix, an action error could additionally leave a hidden <code>decided</code> row stranded and skip rebuilding. This branch clears the failed <code>decided</code> batch and rebuilds before returning. Residual or already collected matches for other callbacks can remain, but that is not a general guarantee that all unattempted work survives.

Rewinding only seminaive timestamps would be a different, at-least-once policy rather than rollback: partial database effects would remain while old matches became eligible again, potentially repeating non-idempotent actions, external effects, or a deterministic panic. A real retry transaction would need to restore the database and union-find, row timestamps and every rule cursor, scheduler state and match buffers, reports, and any external side effects. That broader policy is intentionally outside this PR.

<code>push</code>/<code>pop</code> is a separate explicit snapshot mechanism; neither <code>panic</code> nor <code>fail</code> invokes it automatically. <code>pop</code> intentionally preserves reporting and fresh-symbol progress and cannot undo arbitrary external effects, so it is not a universal transaction. The pre-existing custom-scheduler match-buffer cloning limitation is also separate from #1005 and this follow-up; this PR does not rely on <code>push</code>/<code>pop</code> for error recovery.

An actual Rust panic in extension code still unwinds past <code>fail</code> and has no reuse guarantee. Likewise, if recovery rebuilding fails, the returned error reports both failures and the e-graph has no reuse guarantee. <code>run_rules_no_rebuild</code> remains deliberately lower-level and leaves rebuilding to its caller.

The focused recovery example starts with <code>P(A)</code> and <code>P(B)</code>, then runs an RHS that unions <code>A</code>/<code>B</code> and raises an egglog panic. The returned error leaves one canonical <code>P</code> row. The scheduler regression also shows one already collected unrelated match continuing on a later step; it does not claim preservation of every match. These boundaries are documented on <code>EGraph::run_program</code>, the <code>fail</code> command, both bridge runner variants, the custom scheduler entry point, and the changelog.

### Internal tables and reports match the public abstraction

The scheduler worklist previously used a fixed table name, so a user constructor with that name could be shadowed. Worklist names now come from the e-graph's fresh-symbol generator.

The action side also produced one report iteration per selected rule plus a rebuild-only iteration. The fixed runner aggregates those calls into one logical action iteration while preserving per-rule timing and match data.

Variable-free rules retain an explicit unit marker so their match count and action row are not lost.

## Read-only scheduler context and size measures

<code>SchedulerContext</code> exposes:

- a read-only <code>&EGraph</code>, so custom schedulers can use existing read APIs without adding one wrapper per query;
- a <code>num_nodes()</code> convenience for the common size-aware case.

The node count is lazy. Its cache is owned by the runner and shared across callbacks while the database is unchanged, then invalidated after an action or rebuild. This avoids repeated full scans without letting scheduler implementations own invalidation.

The related public surface is kept small:

- <code>EGraph::total_size()</code> is the exact general table-row measure: all non-hidden functions, including global <code>let</code> bindings;
- <code>EGraph::num_nodes()</code> is the semantic e-node measure: visible constructor rows whose output is a unionable eq-sort;
- <code>Read::is_sort_unionable</code> and <code>Read::is_table_hidden</code> let read primitives reproduce that classification;
- <code>Matches::any_chosen()</code> reports whether a scheduler selected real work.

The semantic node count excludes relations, custom analysis functions, global aliases, and <code>:internal-hidden</code> declarations. It also recognizes the constructor views used by term and proof encoding. A focused cross-mode example creates one real constructor <code>C</code>, an eq-sort-valued analysis function, a relation, a global alias, and a hidden constructor; <code>num_nodes()</code> returns 1 in normal, term, and proof modes.

## Performance

A seven-run alternating release benchmark compared the scheduler fast path with the #1005 head. The workload uses global-naive scheduling, one match per rule, idempotent relation writes, and no unions; steps were 30,000 / 10,000 / 2,000 for 1 / 10 / 100 rules.

| Threads / rules | This branch, median [range] us/step | #1005 head, median [range] us/step | Time reduction |
|---|---:|---:|---:|
| 1 / 1 | 3.776 [3.734-3.887] | 4.129 [4.075-4.343] | 8.6% |
| 1 / 10 | 28.917 [28.398-29.236] | 31.096 [30.746-31.370] | 7.0% |
| 1 / 100 | 268.143 [264.852-270.074] | 288.415 [287.066-292.765] | 7.0% |
| 8 / 1 | 4.045 [3.999-4.108] | 4.376 [4.298-4.423] | 7.6% |
| 8 / 10 | 32.026 [31.703-32.631] | 34.547 [34.265-35.569] | 7.3% |
| 8 / 100 | 303.879 [302.129-308.767] | 323.076 [319.588-328.388] | 5.9% |

A separate 5,000-table probe measured 290.130 us/step when the scheduler ignored its context and 324.401 us/step when it requested <code>num_nodes()</code>: an explicit scan added 34.271 us at the median. Schedulers that do not request the count do not pay for the scan. Rebuild work dominates this synthetic many-table case, so this is a laziness check rather than a claimed Herbie end-to-end speedup.

## Validation

- <code>CARGO_PROFILE_RELEASE_DEBUG_ASSERTIONS=true cargo test -p egglog -p egglog-bridge --release</code>: 68 egglog unit tests, 799 file tests, 51 integration tests, 19 bridge tests, and doctests passed.
- <code>make nits</code>: Clippy, rustfmt, and documentation checks passed.
- Focused recovery tests passed for the bridge runner, high-level <code>fail</code>, and custom scheduler continuation.
- <code>git diff --check</code> passed.

The scheduler regressions additionally cover global and rule-local naive modes, unsafe seminaive mode, live reads of earlier actions, per-rule subsumption, residual draining, report shape, zero-variable rules, constructor-name collisions, semantic node counts, and runner-cache invalidation.


src/lib.rs
```diff
@@ -2313,27 +2325,29 @@ impl EGraph {
         self.backend.table_size(function_id)
     }
 
-    /// The total number of rows across all user-visible functions.
-    pub fn total_size(&self) -> usize {
```
yihozhang, line 2317: this function seem useful to have?

src/scheduler.rs
```diff
@@ -450,25 +480,28 @@ impl SchedulerRuleInfo {
...
+        // options as the ordinary compiled rule.
+        let mut qrule = egraph.backend.new_rule(name, rule.seminaive);
+        qrule.set_no_decomp(rule.no_decomp);
         let mut qrule_builder = BackendRule::new(
-            egraph.backend.new_rule(name, seminaive),
+            qrule,
             &egraph.functions,
             &egraph.type_info,
-            !seminaive,
+            rule.requires_read_context,
```
yihozhang: do we need to check `egraph.seminaive`?

src/scheduler.rs
```diff
@@ -8,23 +9,29 @@ use egglog_bridge::{
...
-    /// Each rule's chosen matches are applied before the next rule's
-    /// `filter_matches`, so sizes are live within an iteration. Between rules
-    /// the tables may still hold rows that the end-of-iteration rebuild will
-    /// merge, so a size read mid-iteration is an upper bound.
-    pub egraph: &'a EGraph,
-    /// 0-based count of iterations this scheduler has completed on this
-    /// ruleset. Schedulers can use a change in this value to detect an
-    /// iteration boundary (e.g., to re-read sizes once per iteration).
-    pub iteration: usize,
+    egraph: &'a EGraph,
```
yihozhang: I think we should make the egraph `pub` because schedulers may query it for different statistics. In fact I'm not sure if the approach here (that `num_nodes` is the only statistics we need) is future-proof.

src/scheduler.rs
```diff
@@ -293,66 +293,91 @@ impl EGraph {
...
-
                         let matches: Vec<Value> =
                             std::mem::take(rule_info.matches.lock().unwrap().as_mut());
                         let mut matches = Matches::new(matches, rule_info.free_vars.clone());
                         rule_info.should_seek =
                             record
                                 .scheduler
                                 .filter_matches(&ctx, rule_id, ruleset, &mut matches);
-                        inserted = matches.any_chosen();
+                        inserted = (matches.all_chosen && matches.match_size() > 0)
```
yihozhang: I think the old was good and we don't need to inline this function

src/scheduler.rs
```diff
@@ -293,66 +293,91 @@ impl EGraph {
...
-            // upper bounds.
-            let seminaive = self.seminaive;
-            let mut action_report = RunReport::default();
+            // Steps 3 and 4: choose and immediately apply each rule's matches.
+            // Rebuilding remains deferred until every rule has run.
+            let mut action_iteration_report = IterationReport::default();
             let mut any_applied = false;
             for (rule_id, _rule) in rules.iter() {
-                let mut inserted = false;
+                let (pending, action_rule) = {
```
yihozhang: Do we need to guarantee anything in the case of "a previous command panicked"? I wonder if we really need this `pending` flag which I assume is for handling this case.

src/scheduler.rs
```diff
@@ -8,23 +9,29 @@ use egglog_bridge::{
...
-    /// `filter_matches`, so sizes are live within an iteration. Between rules
-    /// the tables may still hold rows that the end-of-iteration rebuild will
-    /// merge, so a size read mid-iteration is an upper bound.
-    pub egraph: &'a EGraph,
-    /// 0-based count of iterations this scheduler has completed on this
-    /// ruleset. Schedulers can use a change in this value to detect an
-    /// iteration boundary (e.g., to re-read sizes once per iteration).
-    pub iteration: usize,
+    egraph: &'a EGraph,
+    num_nodes: OnceCell<usize>,
```
yihozhang: When is this OnceCell get reused? Is seems the SchedulerContext is created for every filter_matches and can_stop call. My agent suggested to me the old reuse mechanism is more effective.

src/lib.rs
saulshanabrook: Added back

src/scheduler.rs
saulshanabrook: fixed

src/scheduler.rs
saulshanabrook: fixed

src/scheduler.rs
saulshanabrook: removed

saulshanabrook:
@yihozhang ok the goal here is to preserve at least previous semantics around panic/rollback. So that means we do skip things on semi-naive if we execute once, it panics, then again (wont panic again). But is kept rebuilt at least so you can safely continue.

yihozhang (approved):

Issue #67, opened 2022-11-23
Define command vs define action


saulshanabrook:
There is currently both a top-level define command as well as a define action, with the same names and similar signatures, but different behavior. This was initially a bit confusing for me until I dug into the source and tried out some examples. I am opening this issue to document my understanding of the current behavior (please correct me where I am wrong) and open a discussion to see if there might be a way to make it more straightforward. 

### Top level define command

The top level command is effectively an alias for defining a function that takes no args (a thunk) and setting the value passed in as the result. So for example, I can do:

```lisp
(define x 1)
(check (= (x) 1))
```

Which is, AFAIK, equivalent to:

```lisp
(function x () i64)
(set (x) 1)
(check (= (x) 1))
```

The `define` top-level command is used in many examples to make variables which are then referred to later.

### Define action

The define action however cannot currently be run at the top level of the script but can be used in a rule, to define a temporary local variable that is then accessed later in the rule actions.

For example, this is the initial `fibonacci` example:

```lisp
(function fib (i64) i64)
(set (fib 0) 0)
(set (fib 1) 1)

(rule ((= f0 (fib x))
       (= f1 (fib (+ x 1))))
      ((set (fib (+ x 2)) (+ f0 f1))))

(run 7)

(check (= (fib 7) 13))
```

We can refactor it to define a temporary variable in the rule and use that to set the value:

```lisp
(function fib (i64) i64)
(set (fib 0) 0)
(set (fib 1) 1)

(rule ((= f0 (fib x))
       (= f1 (fib (+ x 1))))
      ((define sum (+ f0 f1))
       (set (fib (+ x 2)) sum)))

(run 7)

(check (= (fib 7) 13))
```

This is different from the above define, in that it sets it as a variable, not as a function. So you cannot call the `sum` in the `set` action below it.

As far as I can tell, this isn't used much at all in the examples. The only place I can see it used is in the `matrix` example. In this case, the local variables defined are not ever actually accessed later. I assume something about simply defining them has an affect on the script, but I am not sure what this is:

```
; demand
(rule ((= e (MMul A B)))
((define demand1 (ncols A))
(define demand2 (nrows A))
(define demand3 (ncols B))
(define demand4 (nrows B)))
)
```

## Possible changes

As I said above, I am not opening this because there are any bugs in the current implementation necessary, but simply because the ergonomic are a bit confusing. Depending on where the define is located it can either be a global function or a local variable. Also the top level one takes a cost parameter, which the local variable does not. 

[...]

PR #1008, opened 2026-08-31
Make displayed commands replayable and enable output downcasting

saulshanabrook:
## Summary

- Render rule names and panic messages through Egglog's string-literal formatter, preserving quotes and backslashes as valid Egglog syntax.
- Render file paths as string literals for `print-stats`, `print-function`, `input`, `output`, and `include`, so displayed commands can be parsed and replayed.
- Add `UserDefinedCommandOutput::as_any()`, allowing a consumer that knows an extension's concrete output type to downcast safely.
- Keep borrowed values eligible for `UserDefinedCommandOutput` by placing the required `'static` bound on `as_any`, rather than on the trait itself.

The replayable display changes are needed by clients that record successful commands and later reconstruct an e-graph. The downcast supports typed extension outputs without adding extension-specific variants to core; egglog-experimental #67 uses it for `MultiExtractOutput`, and egglog-python #414 exposes that result to Python.

## Validation

- `cargo test -p egglog-ast display_ --locked`
- `cargo test -p egglog test_file_command_display_roundtrip --locked`
- `cargo test --test user_defined_command_output --locked`
- `cargo test --test user_defined_command_output --no-default-features --locked`
- `cargo check -p egglog --no-default-features --lib --locked`
- `cargo test --doc -p egglog --locked`
- `cargo fmt --all --check`
- `cargo clippy --tests --locked -- -D warnings`

The current head is `1eea60a14f214505741d22bdd6c9b501b21c04ee`. All GitHub checks are green; the PR is mergeable and awaits non-author approval.

## Stack

This is the first PR in:

1. egraphs-good/egglog#1008
2. egraphs-good/egglog-experimental#67
3. egraphs-good/egglog-python#414

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - Add `Read::eclass_enodes` to scan an e-class across every constructor, rename the single-constructor method to `constructor_enodes_for_eclass`, and add `Enode::name`.
 
+- Allow consumers to safely downcast `UserDefinedCommandOutput` trait objects to known concrete output types.
+
 - Treat tables dropped by `pop` as missing in the name-indexed `Read` and `Write` APIs.
 
 - Escape quotes and backslashes in printed rule names so the generated syntax round-trips through the parser.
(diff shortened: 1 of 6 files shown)


oflatt (approved):
nice fix

PR #1010, opened 2026-09-04
Fix subsuming rows created in the same action

saulshanabrook:
## Summary

Make constructor subsumption work for rows created earlier in the same action, both in compiled egglog rule RHSs and through the `Write`/`FullState` API. Subsumption is dominant regardless of whether an add appears before or after it, and actions following a missing-row `subsume` continue to execute.

This uses the existing table merge and prediction machinery. It adds no public API and makes no `core-relations` changes.

## Bugs fixed

- **Rule RHS ordering could leave a row live and skip later actions.** The first half of rule-level subsumption uses `lookup_or_insert` to supply a constructor output, predicting a subsumed row on a miss. Its second callback nevertheless performed a committed-only status lookup before writing. For a row predicted by a preceding constructor add, that lookup removed the action lane: the row flushed live and every later RHS action was skipped. A direct missing-row `subsume` predicted a hidden row but still skipped later actions. The second callback now writes the same row with `SUBSUMED` directly; the existing table merge makes subsumption dominant for committed and predicted rows without another lookup.
- **The runtime API could not subsume a same-action row.** `FullState::add` followed by `FullState::subsume` panicked because `TableAction::subsume` searched only committed rows. Directly subsuming a missing constructor failed the same way. It now uses `lookup_or_insert`, reusing any predicted output before staging the subsumed row.

Committed live and already-subsumed rows retain their existing explicit-subsume behavior.

## Validation

- `make test` — 1,191 tests plus doctests
- `make nits`
- `cargo test -p egglog-bridge` — 19 tests
- `cargo test --test integration_test` — 59 tests
- `cargo test --test api_fact_ops test_subsume_predicted_constructor_rows`
- focused rule and `FullState` regressions for direct subsume and both action orders
- `cargo check -p egglog --no-default-features --lib`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `git diff --check`

An earlier full-suite run hit the known-flaky `core_relations::tests::early_stop`; it passed in isolation and the complete suite passed on rerun.

### DIFF
--- a/egglog-bridge/src/lib.rs
+++ b/egglog-bridge/src/lib.rs
@@ -1674,12 +1674,16 @@ impl TableAction {
         state.stage_remove(self.table, key);
     }
 
-    /// Subsume a row in this table.
+    /// Subsume a row in a table with subsumption enabled. For a constructor,
+    /// the configured default is inserted if the key is absent. Rows predicted
+    /// earlier in the same action are reused.
     pub fn subsume(&self, state: &mut ExecutionState, key: impl Iterator<Item = Value>) {
         let ts = Value::from_usize(state.read_counter(self.timestamp));
         let mut scratch = key.collect::<SmallVec<[_; 8]>>();
 
-        let ret_val = self.lookup(state, &scratch).expect("subsume lookup failed");
+        let ret_val = self
+            .lookup_or_insert(state, &scratch)
+            .expect("subsume lookup failed");
 
         self.table_math.write_table_row(
             &mut scratch,
--- a/egglog-bridge/src/rule.rs
+++ b/egglog-bridge/src/rule.rs
@@ -496,13 +496,7 @@ impl RuleBuilder<'_> {
         let ret: QueryEntry = ret.into();
         self.add_callback(move |inner, rb| {
             // Then, add a tuple subsuming the entry, but only if the entry isn't already subsumed.
-            // Look up the current subsume value.
             let mut dst_entries = inner.convert_all(&entries);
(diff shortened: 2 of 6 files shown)


egglog-bridge/src/rule.rs
```diff
@@ -511,6 +505,19 @@ impl RuleBuilder<'_> {
                     ret_val: Some(inner.convert(&ret)),
                 },
             );
+            // Look up the current subsume value, including rows predicted by
```
oflatt: Is this change necessary really? I'm confused. Rebuilding handles the case where there are two equivalent e-nodes that are subsumed, so we should be able to just write the subsumed flag.

egglog-bridge/src/lib.rs
```diff
@@ -1674,12 +1674,16 @@ impl TableAction {
...
-    /// Subsume a row in this table.
+    /// Subsume a row in a table with subsumption enabled. For a constructor,
+    /// the configured default is inserted if the key is absent. Rows predicted
+    /// earlier in the same action are reused.
     pub fn subsume(&self, state: &mut ExecutionState, key: impl Iterator<Item = Value>) {
         let ts = Value::from_usize(state.read_counter(self.timestamp));
         let mut scratch = key.collect::<SmallVec<[_; 8]>>();
 
-        let ret_val = self.lookup(state, &scratch).expect("subsume lookup failed");
+        let ret_val = self
```
oflatt, line 1684: good fix here

oflatt (approved):
nice

PR #1011, opened 2026-09-05
Add standalone e-graph comparison by partition refinement

ezrosent:
*From Codex:*

Add a standalone `egraph-comparison` library and binary that compare rebuilt JSON databases by conservative partition refinement. Report constructor-term equality separately from full database equality, including ordinary-function structure and subsumption from this first layer.

Use documented numeric-id types for wire versions, classes, partitions, and rows, with structured validation errors. The README explains the algorithm independently, links DFA minimization and generic coalgebraic minimization, and explicitly notes that this implementation scans the whole graph rather than using Hopcroft’s smaller-half worklist.

Validation: comparison/oracle, cyclic/renaming, CLI, typed-error, and function-only counterexample tests pass independently at this layer; Clippy passes.

Part of the actual `gh stack` #1023. Non-test diff: 589 added/deleted lines; tests are in separate files.

### DIFF
--- a/egraph-comparison/README.md
+++ b/egraph-comparison/README.md
@@ -1,0 +1,87 @@
+# egraph-comparison
+
+```sh
+cargo run -p egraph-comparison -- left.json right.json
+cargo run -p egraph-comparison -- --terms-only left.json right.json
+```
+
+The binary reads two complete, rebuilt databases and emits JSON with
+`terms_equal`, `database_equal`, `refinement_rounds`, and
+`database_refinement_rounds`. Exit status is 0 for
+equality, 1 for disequality, and 2 for invalid input or I/O failure. `--terms-only`
+selects term equality for the exit status; both results are always reported.
+
+## Semantics
+
+Partition refinement starts with a coarse grouping of e-classes and repeatedly
+splits groups whose members have different observations. Initially, classes are
+grouped by sort. A class's observation is the set of its e-node labels together
+with the current groups of each node's ordered children. Refinement stops when
+no group can split. Comparing two inputs refines their disjoint union and checks
+that the same groups contain observable roots on both sides.
+
+This is the idea behind [DFA minimization](https://en.wikipedia.org/wiki/DFA_minimization).
+For its generalization to other transition structures, see Jacobs and Wißmann,
+[Fast Coalgebraic Bisimilarity Minimization](https://arxiv.org/abs/2204.12368).
+The implementation here uses whole-graph rounds; it does **not** implement
+Hopcroft's smaller-half splitter worklist or the fast algorithm in that paper.
+
+This is conservative e-graph partition refinement: cycles merge only when the
+complete sets of e-nodes in their e-classes are identical modulo the refined
+child groups. Merely overlapping sets do not suffice. A self-loop and two
+mutually recursive classes compare equal when they have the same complete
+observations, as do duplicated copies of a cycle. Adding an unmatched member to
+one class can distinguish it even when both classes still contain a common term.
+
+`terms_equal` compares constructor-output classes by bisimulation. Node labels
+include function name and signature; children are ordered, e-class members are
+sets, and sorts and primitive literal values are exact. Literals are allowed as
+constructor arguments, but primitive values occurring only in function tables
+do not add constructor terms. Empty classes are observable through constructor
+arguments. Bisimulation includes ungrounded cycles and is deliberately stronger
+than equality of finite ground-term languages. This is not graph isomorphism.
+
+`database_equal` additionally compares function declarations and all rows,
+including constructor rows and subsumption flags. Its full-database refinement
+includes ordinary function calls in class observations. This preserves
+structure carried only by function tables, even when values have no constructor
+terms. Values in the rows are compared using this full-database partition.
+Non-constructor functions never become term operators or enter finite term
+certificates. All comparisons are of sets, so multiplicities of duplicate rows
+or bisimilar values are ignored. Unused classes are ignored. This still compares
+bisimulation quotients, not bijective identities or graph isomorphism: empty
+classes of the same sort remain indistinguishable if no incoming rows give them
+different observations.
+Costs, roots, extraction preferences, and runtime implementation details are
+outside this database format.
+
+The implementation recomputes exact signatures each round. It has at most a
+linear number of splitting rounds and can take quadratic time (plus signature
+sorting). There is no probabilistic equality or depth limit. Performance
+improvements preserve exact signature equality.
+
+## Version 1 JSON
+
+```json
+{
+  "version": 1,
+  "classes": {"a": {"sort": "Expr"}, "n": {"sort": "i64", "literal": "1"}},
+  "functions": {
+    "Num": {"kind": "constructor", "inputs": ["i64"], "output": "Expr"},
+    "cost": {"kind": "function", "inputs": ["Expr"], "output": "i64"}
+  },
+  "rows": [
+    {"function": "Num", "inputs": ["n"], "output": "a"},
+    {"function": "cost", "inputs": ["a"], "output": "n"}
+  ]
+}
+```
+
+Class IDs are arbitrary strings scoped to one file. The numeric `version` field
+is a `FormatVersion` identifying the wire format, not an e-class identifier.
+Setup resolves class names to distinct typed class and partition IDs.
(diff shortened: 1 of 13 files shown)


egraph-comparison/src/main.rs
```diff
@@ -0,0 +1,38 @@
+use clap::Parser;
+use egraph_comparison::{Database, compare};
+use std::{fs::File, io::BufReader, path::PathBuf, process::ExitCode};
+
+#[derive(Parser)]
+#[command(about = "Compare two serialized e-graphs modulo constructor bisimulation")]
+struct Args {
+    left: PathBuf,
+    right: PathBuf,
+    /// Use constructor-term equality for the exit status (still report both).
```
ezrosent: Explain what constructor-term equality is (or link to relevant docs)

egraph-comparison/src/model.rs
```diff
@@ -0,0 +1,128 @@
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
+
+/// Version 1 is a complete, canonicalized database, without visualization limits.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct Database {
+    pub version: u32,
```
ezrosent: use numeric-id here and document what this identifier means.

egraph-comparison/src/model.rs
```diff
@@ -0,0 +1,128 @@
...
+    pub inputs: Vec<String>,
+    pub output: String,
+    /// Subsumed constructors still denote terms; activity affects database equality.
+    #[serde(default)]
+    pub subsumed: bool,
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("{0}")]
+pub struct Error(pub String);
```
ezrosent: Let's have a better thiserror-style Error enum. That'll make the code a bit more readable.

egraph-comparison/src/refine.rs
```diff
@@ -0,0 +1,171 @@
...
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) enum Node {
+    Literal(String),
+    Call(String, Function, Vec<usize>),
+}
+
+pub(crate) struct Graph {
+    pub sorts: Vec<String>,
+    pub nodes: Vec<Vec<Node>>,
+    pub roots: BTreeSet<usize>,
```
ezrosent: user numeric-id to specify what these `usize`s mean

egraph-comparison/src/refine.rs
```diff
@@ -0,0 +1,171 @@
...
+        Self {
+            left,
+            right,
+            blocks,
+            rounds: 0,
+        }
+    }
+
+    pub fn step(&mut self) -> bool {
+        let mut signatures = BTreeMap::new();
```
ezrosent, line 108: Feel free to defer this to the optimization PR, but there's probably a better layout we could use here using a flat HashTable from hashbrown, where each entry contains a hash and a pointer/len into a flat array that contains all keys and all values. (where the main "signature" is just a sorted slice rather than a BTreeSet).

This'd avoid tons of excessive allocations and comparisons of large signatures.

egraph-comparison/src/refine.rs
```diff
@@ -0,0 +1,171 @@
...
+        Self {
+            sorts,
+            nodes,
+            roots,
+            index,
+        }
+    }
+}
+
+pub(crate) struct Partition {
```
ezrosent, line 69: Where are we doing hopcroft's trick?

egraph-comparison/README.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+The binary reads two complete, rebuilt databases and emits JSON with
+`terms_equal`, `database_equal`, and `refinement_rounds`. Exit status is 0 for
+equality, 1 for disequality, and 2 for invalid input or I/O failure. `--terms-only`
+selects term equality for the exit status; both results are always reported.
+
+## Semantics
+
+Comparison follows the partition refinement algorithm in Eli Rosenthal's
```
ezrosent: Make this self-contained. Don't reference this unpublished note (which is written by me). Introduce the concept of partition refinement and reference links on DFA minimization (wikipedia) and generic coalgebra minimization [this paper specifically](https://arxiv.org/abs/2204.12368)

Also mention that this is a conservative partition-refinement algorithm on an e-graph, where we merge cycles only if their e-classes and e-nodes are identical (not just overlap).

egraph-comparison/src/main.rs
ezrosent: *From Codex:*

Expanded the CLI help to say that constructor-term equality compares constructor-built terms, their e-class equalities, and cyclic structure, while ordinary function tables and subsumption do not affect that exit status. It also points to the self-contained Semantics section in the README.

egraph-comparison/src/model.rs
ezrosent: *From Codex:*

Replaced the raw integer with `FormatVersion`, defined using `egglog-numeric-id`, and documented that it selects the wire format rather than identifying an e-class. JSON still uses a scalar integer. Serialized class names remain arbitrary input-local strings; setup resolves them to a separate typed `ClassId` namespace.

egraph-comparison/src/model.rs
ezrosent: *From Codex:*

Replaced the string wrapper with a `thiserror` enum covering unsupported versions, duplicate literals, undeclared functions, arity, missing classes, sort mismatches, constructor literals, and conflicting rows. The variants carry typed row IDs and relevant context, and tests match the variants directly. The exporter in #1013 now has a separate structured `ComparisonExportError` that preserves rebuild/validation error sources.

egraph-comparison/src/refine.rs
ezrosent: *From Codex:*

Added documented `ClassId` and `BlockId` types using `egglog-numeric-id`. Class IDs index the disjoint union, with right-side IDs offset past the left input; block IDs identify groups in a particular partition and round. The optimization layer adds a separate `SymbolId`, and certificates use `TermId` and `RowId` rather than sharing integer namespaces.

egraph-comparison/src/refine.rs
ezrosent: *From Codex:*

Implemented this in #1017. The `hashbrown::HashTable` entries hold a cached hash, an offset/length range into one flat key arena, and a typed block ID. Scratch node signatures are sorted slices; distinct class signatures no longer each own a separate Vec allocation. Length prefixes preserve node/arity boundaries, and hash collisions still compare complete keys. Tests force collisions through table growth. In local three-sample checks, run-11 comparison dropped from 1.665 s to 1.205 s (about 28%); the measurements and harness remain outside the committed tree.

egraph-comparison/src/refine.rs
ezrosent: *From Codex:*

Nowhere in the current implementation: it still recomputes whole-graph signatures each round. I made this explicit in the core README and `compare` documentation, including that the fast algorithm and complexity bounds in the linked paper are not implemented here. The flat-key change in #1017 reduces allocation/hash costs; a predecessor/smaller-half worklist remains a separate algorithmic change.

egraph-comparison/README.md
ezrosent: *From Codex:*

Rewrote this as a self-contained introduction to coarse partitions, signature-based splitting, fixed points, and comparing root blocks. Removed the unpublished-note reference and added the requested DFA-minimization and Jacobs/Wißmann paper links. It now explicitly says that complete e-node sets must match modulo child blocks: overlapping members alone do not justify merging cycles.

egraph-comparison/src/refine.rs
ezrosent: *From Codex:*

Implemented the smaller-half worklist in #1021, directly above the initial optimization layer (now #1022, replacing #1017 after the stack reorder) and before certificates/export. The largest split piece retains its ID; a reverse child index dirties only predecessors of smaller pieces. Processing evaluates dirty signatures plus one clean representative without scanning all clean members. FIFO scheduling and bounded signature-table retention reduce overhead.

Against the reordered parent, the run-12 comparison median dropped from 29.63 s to 20.43 s across three separate measurements each (31% less time, 1.45× speedup). The PR includes the full run-1-through-12 results. Scripts and raw data remain local. Synchronous refinement is retained for depth-based certificates and as a correctness oracle; canonical numbering remains separate.

PR #1012, opened 2026-09-05
Extract and verify e-graph disequality certificates

ezrosent:
*From Codex:*

Extract and verify missing-constructor-term, unequal-term, structural, declaration, and full-database row certificates. Term references use typed IDs in a topologically ordered DAG; `FixedBitSet` compaction keeps only the witness sub-DAG. Full-function certificate handling is included here rather than deferred to #1014.

Document the extraction pipeline, finite-ground productivity, independent evaluator, structural replay, compaction, and DAG validation. An included doctest contrasts an ungrounded cycle with one containing a finite constructor representative.

Validation: certificate extraction/verification, malformed DAGs, shared subterms, cycles, CLI verification, and the doctest pass at this layer; Clippy passes.

This layer follows the comparison optimizations. Structural certificates retain synchronous refinement and depth-based replay; comparison worklist step counts are not certificate depths.

Part of the actual `gh stack` #1023. Non-test diff: 573 added/deleted lines; tests are in separate files.

### DIFF
--- a/egraph-comparison/README.md
+++ b/egraph-comparison/README.md
@@ -93,3 +93,38 @@ are explicit even for empty tables. Unknown fields, unsupported versions,
 dangling IDs, wrong sorts/arity, duplicate literal identities, and conflicting
 function rows are errors. Producers must canonicalize IDs and rebuild first.
 This format is separate from visualization-oriented `egraph-serialize` JSON.
+
+## Disequality certificates
+
+Pass `--certificate` to include a `certificate` field (`null` for equal
+inputs). Save that field alone to `witness.json` and verify it with:
+
+```sh
+cargo run -p egraph-comparison -- left.json right.json --verify-certificate witness.json
+```
+
+Verification emits `{"valid": true}` or `{"valid": false}` with status 0 or 1.
+Malformed JSON/input still exits 2. The Rust API exposes `certificate` and
+`verify` for the same operations.
+
+* `missing_term`: a constructor term exists on only the indicated side.
+* `unequal_terms`: both terms exist in both inputs, and are equal only on the
+  indicated side.
+* `structure`: a constructor class has a bounded bisimulation observation absent
+  from all constructor classes in the other input. The class ID and refinement
+  round count let the verifier replay that observation using synchronous refinement;
+  comparison worklist steps are not certificate depths. This is needed for
+  ungrounded cycles/holes, which may have no finite ground-term witness.
+* `declaration` / `row`: a schema or table row differs. Row IDs are interpreted
+  in the indicated input and compared modulo the joint full-database partition.
+
+Finite terms use a topologically ordered DAG: each application references earlier
+entries, and the certificate identifies its root(s). This avoids exponential
+expansion and recursive evaluation. Functions never appear in these terms.
+The extractor finds one ground representative per productive class, then checks
+constructor rows against their output representatives in both directions.
+Ground-term verification uses direct constructor lookup, independently of the
+refinement algorithm. Certificates are valid witnesses, not guaranteed minimal.
+Structural and row verification use exact refinement. Certificate generation
+currently repeats comparison/refinement and row-witness search can be quadratic;
+it is opt-in so equality checks do not pay this diagnostic cost.
--- a/egraph-comparison/src/certificate.rs
+++ b/egraph-comparison/src/certificate.rs
@@ -1,0 +1,480 @@
+use crate::ids::{RowId, TermId};
+use crate::{Database, Error, FunctionKind, Row, compare, refine::Partition};
+use egglog_numeric_id::NumericId;
+use fixedbitset::FixedBitSet;
+use serde::{Deserialize, Serialize};
+use std::collections::{BTreeMap, VecDeque};
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum Side {
+    Left,
+    Right,
+}
+
+/// A finite term DAG. Children refer to earlier entries in the same vector,
+/// preventing exponential expansion (and recursion on deeply nested terms).
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "kind", rename_all = "snake_case", deny_unknown_fields)]
+pub enum Term {
+    Literal {
+        sort: String,
+        value: String,
+    },
+    Apply {
+        function: String,
+        inputs: Vec<TermId>,
+    },
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "kind", rename_all = "snake_case", deny_unknown_fields)]
+pub enum Certificate {
+    /// `term` exists in `side`, but has no interpretation in the other input.
+    MissingTerm {
+        side: Side,
+        terms: Vec<Term>,
+        term: TermId,
+    },
+    /// Both terms exist on both sides, but are equal only in `side`.
+    UnequalTerms {
+        side: Side,
+        terms: Vec<Term>,
+        first: TermId,
+        second: TermId,
+    },
+    /// A constructor class's depth-bounded bisimulation observation occurs on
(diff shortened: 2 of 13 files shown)


egraph-comparison/src/certificate.rs
```diff
@@ -0,0 +1,431 @@
...
+    UnequalTerms {
+        side: Side,
+        terms: Vec<Term>,
+        first: usize,
+        second: usize,
+    },
+    /// A constructor class's depth-bounded bisimulation observation occurs on
+    /// only one side. This also covers cycles that have no finite ground terms.
+    /// Replaying `rounds` refinement steps verifies the witness without trusting
+    /// input-dependent block numbers. `class` is an ID in the indicated input.
```
ezrosent: Too terse. What does it mean to "replay refinement"?

egraph-comparison/src/certificate.rs
```diff
@@ -0,0 +1,431 @@
...
+
+/// Return a certificate for any difference, preferring finite term witnesses.
+/// The fallback is a finite structural observation, never a fabricated ground
+/// term for an ungrounded cycle. Equal databases return `None`.
+pub fn certificate(left: &Database, right: &Database) -> Result<Option<Certificate>, Error> {
+    let result = compare(left, right)?;
+    if result.database_equal {
+        return Ok(None);
+    }
+    if !result.terms_equal {
```
ezrosent, line 83: more comments for this function

egraph-comparison/src/certificate.rs
```diff
@@ -0,0 +1,431 @@
...
+                return Ok(Some(witness));
+            }
+        }
+    }
+    Err(Error(
+        "disequality without a witness (internal error)".into(),
+    ))
+}
+
+// Extract one representative per productive class with a bottom-up worklist.
```
ezrosent: By 'productive' do you mean in the coinduction sense? If so, say so.

egraph-comparison/src/certificate.rs
```diff
@@ -0,0 +1,431 @@
...
+// Each row becomes ready at most once; repeated arguments are counted separately.
+fn representatives(db: &Database) -> (Vec<Term>, BTreeMap<String, usize>) {
```
ezrosent: Add a doctest to give an example.

egraph-comparison/src/certificate.rs
```diff
@@ -0,0 +1,431 @@
...
+                if remaining[i] == 0 {
+                    ready.push_back(i);
+                }
+            }
+        }
+    }
+    (terms, reps)
+}
+
+struct Evaluator<'a> {
```
ezrosent, line 206: What is this used for?

egraph-comparison/src/certificate.rs
```diff
@@ -0,0 +1,431 @@
...
+                terms,
+                first: roots[0],
+                second: roots[1],
+            });
+        }
+    }
+    None
+}
+
+fn compact(terms: Vec<Term>, roots: &[usize]) -> (Vec<Term>, Vec<usize>) {
```
ezrosent: Explain what this does.

egraph-comparison/src/certificate.rs
```diff
@@ -0,0 +1,431 @@
...
+    let mut used = vec![false; terms.len()];
```
ezrosent: use fixedbitset

egraph-comparison/src/certificate.rs
```diff
@@ -0,0 +1,431 @@
...
+}
+
+fn valid_dag(terms: &[Term]) -> bool {
+    terms.iter().enumerate().all(|(i, term)| match term {
+        Term::Literal { .. } => true,
+        Term::Apply { inputs, .. } => inputs.iter().all(|&child| child < i),
+    })
+}
+
+fn verify_with_partition(
```
ezrosent, line 436: explain what this does

egraph-comparison/README.md
```diff
@@ -62,3 +62,37 @@ are explicit even for empty tables. Unknown fields, unsupported versions,
...
+Finite terms use a topologically ordered DAG: each application references earlier
+entries, and the certificate identifies its root(s). This avoids exponential
+expansion and recursive evaluation. Functions never appear in these terms.
+The extractor finds one ground representative per productive class, then checks
+constructor rows against their output representatives in both directions.
+Ground-term verification uses direct constructor lookup, independently of the
+refinement algorithm. Certificates are valid witnesses, not guaranteed minimal.
+Structural and row verification use exact refinement. Certificate generation
+currently repeats comparison/refinement and row-witness search can be quadratic;
+it is opt-in so equality checks do not pay this diagnostic cost.
```
ezrosent, line 130: Give an example here of two unequal e-graphs and their resulting certificate. Bonus points if it's two egglog programs (in which case, you can defer this until the egglog plumbing is wired in)

egraph-comparison/src/certificate.rs
ezrosent: *From Codex:*

Expanded `Structure` documentation to spell out verification: reconstruct the disjoint constructor graphs, group by sort, compute the stated number of successive partitions, then check that the named constructor root has no root in the same block on the other side. The certificate contains an input-local class name and round count; no supplied partition/block numbering is trusted.

egraph-comparison/src/certificate.rs
ezrosent: *From Codex:*

Added comments explaining the extraction order: try finite constructor witnesses in both directions, use a finite-depth structural witness for differences involving ungrounded cycles/holes, then check declarations and full-database rows. The comments also explain why term witnesses exclude ordinary functions.

egraph-comparison/src/certificate.rs
ezrosent: *From Codex:*

Here it means productivity in the grammar/inductive sense: the class contains a finite ground constructor term. It is not coinductive productivity. The comment now says this explicitly: `x = f(x)` alone never enters the ready worklist, while adding `A()` gives the class a finite representative.

egraph-comparison/src/certificate.rs
ezrosent: *From Codex:*

Added a passing doctest demonstrating the pure `x = f(x)` cycle and the same cycle with `A()` added, then extracting and verifying the finite missing-term certificate. The example is attached to the public `certificate` API and referenced from `representatives`; its source lives in `tests/support/representatives.md` and is included into the documentation so tests stay in their own file.

egraph-comparison/src/certificate.rs
ezrosent: *From Codex:*

Documented `Evaluator` at its definition. Witness extraction uses it to interpret source representatives in the target database; verification uses it independently on both inputs. It looks up only literals and constructor applications, without relying on partition blocks or treating ordinary function rows as terms.

egraph-comparison/src/certificate.rs
ezrosent: *From Codex:*

Documented how `compact` keeps only the sub-DAG reachable from the selected witness roots. It marks dependencies backward using the existing topological order, then removes unrelated representatives and remaps children and roots forward to dense term IDs. This is what keeps a certificate from containing every representative extracted from the source database.

egraph-comparison/src/certificate.rs
ezrosent: *From Codex:*

Changed the reachability marks in `compact` to `FixedBitSet`. The existing shared-subterm and invalid-DAG tests still pass.

egraph-comparison/src/certificate.rs
ezrosent: *From Codex:*

Documented the invariant checked by `valid_dag`: every child must precede its parent. That rejects self-cycles, longer cycles, forward references, and out-of-range child indices without recursive traversal. The verifier checks certificate root bounds separately.

egraph-comparison/README.md
ezrosent: *From Codex:*

Added the concrete example in #1013, where the exporter is available: two egglog programs containing `Num(1)` and `Num(2)`, their export/comparison instructions, and the resulting `missing_term` certificate JSON. I ran the programs and checked that the actual CLI certificate matches the documented literal/application DAG and root.

PR #1013, opened 2026-09-05
Export complete egglog databases for semantic comparison

ezrosent:
*From Codex:*

Add `EGraph::serialize_for_comparison` and `--to-comparison-json`, exporting complete visible constructor/function/relation tables after rebuild. Structured `ComparisonExportError` variants distinguish unsupported values/projections, rebuild failures, and invalid exported databases.

Support equality sorts and built-in scalars; unsupported containers/custom base values and term/proof projections fail explicitly. Document the variable-arity observation design needed for containers, and show two egglog programs with their actual missing-term certificate.

Validation: all 10 exporter tests pass. The README certificate was checked against real CLI output.

Part of the actual `gh stack` #1023. Non-test diff: 271 added/deleted lines; tests are in separate files.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -472,6 +472,7 @@ dependencies = [
  "egglog-core-relations",
  "egglog-numeric-id",
  "egglog-reports",
+ "egraph-comparison",
  "egraph-serialize",
  "enum-map",
  "env_logger",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -122,6 +122,7 @@ required-features = ["bin"]
 default = ["bin"]
 
 bin = [
+  "comparison",
   "serde",
   "graphviz",
   "dep:clap",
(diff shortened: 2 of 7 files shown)


ezrosent (commented):
Containers should be encodeable as "variable-arity tables" I think

ezrosent:
*From Codex:*

Agreed on [the container review](https://github.com/egraphs-good/egglog/pull/1013#pullrequestreview-5128273569): variable-arity table observations are a useful representation. I added that design to the exporter and migration documentation. The label needs the container kind and element sorts, and the child interpretation must preserve sequence order, set semantics, multiset multiplicity, or map key/value pairing as appropriate.

The current input schema has fixed-arity ordered function arguments, so it cannot express all of those semantics directly. This PR still rejects unsupported containers explicitly; adding the variable-arity observation forms, validation, and certificate support remains a separate format/implementation extension.

PR #1014, opened 2026-09-05
Test function-induced database structure and certificates

ezrosent:
*From Codex:*

Add regression coverage for function-induced structure without constructor terms, bisimilar ordinary-function cycles, function-only changes, and row-certificate verification. Extend the independent relation oracle to function-only graphs.

The semantic implementation is already in #1011 and the certificate handling is in #1012. This layer contains tests only. All comparison tests pass at this layer.

Part of the actual `gh stack` #1023. Non-test diff: 0 added/deleted lines; tests are in separate files.

### DIFF
--- a/egraph-comparison/tests/functions.rs
+++ b/egraph-comparison/tests/functions.rs
@@ -1,0 +1,71 @@
+use egraph_comparison::{Certificate, Database, certificate, compare, verify};
+
+fn db(rows: serde_json::Value) -> Database {
+    serde_json::from_value(serde_json::json!({
+        "version":1,
+        "classes":{"x":{"sort":"E"},"y":{"sort":"E"}},
+        "functions":{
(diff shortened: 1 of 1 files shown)


ezrosent (commented):
Should this bug fix be merged in to the bottom of the stack?

ezrosent:
*From Codex:*

Yes, in response to [the stack-placement review](https://github.com/egraphs-good/egglog/pull/1014#pullrequestreview-5128284899), I moved the full-function refinement implementation into #1011 and its certificate handling into #1012. The core PR now has its own function-only counterexample regression.

This PR now contains only the additional function-structure/certificate regression tests and function-mode oracle coverage, with no production-code diff. I renamed the PR accordingly and restacked all eight layers using `gh stack`.

